PR #46, opened 2022-05-06
add some required plugin details

keevan:
- docs: add branches table and installation instructions
- ci: add reusable workflows and set a fixed supported range


jaypha:
I think it may be too soon for a CI.

Issue #20, opened 2022-04-05
Decide on architecture of how steps are declared


brendanheywood:
Usual ideas:

1. sub plugins of this plugin
2. classes in this plugins
3. callback classes

Leaning towards 3) class which is some sort of callback. And then other plugins can implement their own steps. eg perhaps the steps for pushing to SNS or S3 could live inside local_aws and not in this plugin as an example.

keevan:
Mostly in place now, just needs another plugin with the function defined in `lib.php` to confirm whether or not it's linking up as expected.

Making a note here to probably try it using via local_aws as an initial example.

PR #47, opened 2022-05-06
Add callback classes handling

keevan:
This only populates the list of available steps supplied by other plugins.

Currently I have added an example one which only ingests some data and outputs the same data, whilst doing a side effect of passing it to debugging in json format.

Related to #20

### DIFF
--- a/classes/manager.php
+++ b/classes/manager.php
@@ -1,0 +1,52 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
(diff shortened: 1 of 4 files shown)


keevan:
Probably will require some docs in the README, but this can be deferred to an issue later down the line if the intended structure is sound. 

The reason I did not opt for different step "types" too early is because this could probably be handled internally by each plugin's step declaration. And based on where it ends up living it could probably handle a lot more than one action per class/object theoretically.

PR #49, opened 2022-05-09
feat: add base handling to render a workflow graph in dot

keevan:
WIP:

Will need to store and load the dataflow configurations to render an actual workflow, but this will render the example below for now.

image

### DIFF
--- a/visual.php
+++ b/visual.php
@@ -34,6 +34,7 @@ $type = optional_param('type', 'svg', PARAM_TEXT);
 // Generate DOT script based on the configured workflow.
 $dotscript = <<<EXAMPLE
     digraph G {
+      rankdir=LR;
       node [shape = record,height=.1];
       "read users" -> "write local csv";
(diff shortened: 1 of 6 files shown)


visual.php
```diff
@@ -0,0 +1,53 @@
...
+
+require_once(__DIR__ . '/../../../config.php');
+require_once($CFG->libdir.'/adminlib.php');
+
+// Ensure permissions/roles are correct.
+require_login(null, false);
+require_capability('moodle/site:config', context_system::instance());
+
+// TODO: require an id to the dataflow so it can load and render the resulting graph/visual.
+$type = optional_param('type', 'svg', PARAM_TEXT);
```
brendanheywood, line 32: shouldnt this need an input of the dataflow id or similar?

keevan: Yes, correct. I wanted to also see if the output type could be changed as required.

PR #51, opened 2022-05-11
24 admin ui

keevan:
- fix: phpdoc for visual.php
- add settings page and lang strings
- style: clean up names to be dataflow instead of workflow for consistency
- feat: add persistent dataflow model and form confirmed via a small test
- fix: phpcs issues
- [wip] add table to display dataflows with a small preview
- style: fix comment typo
- [wip] add steps table and step form


keevan:
Dataflow list > My Flow (dataflow) > Steps page preview:

image

Will need to add the dependency saving functionality and use that to generate the image.

Issue #25, opened 2022-04-06
UI for CRUD for a step


brendanheywood:
- Select the step type
- each step declares if it accepts 0 1 or 2 input streams
- each step declares if it emits 0 or 1 output streams
- basic wiring for a step it's its inputs and outputs
- it is ok for a step to be in an invalid state while we build a flow
- have a validation overall for the flow
  - do all the inputs and outputs have the right number of connections?
  - it is ok (but weird) if there are disconnected graphs inside a flow

keevan:
Seeing as it's okay for the flow to be in an invalid state (e.g. until it is fixed), it would probably be good to display the list of validation issues on the steps list page (table of steps). 

For now it would only validate:
- That it is in an appropriate dag format (no circular deps, no self referencing, etc)
- The number of expected connections going in and out.

Feel free to add more to this list if needed at this stage.

PR #55, opened 2022-05-12
24 add ability to edit step dependency

### LINKED ISSUE #24: Admin UI for creating a flow
- Flows have names
- Flows have a schedule, or can be manually triggered

keevan:
Adds the UI for configuring step dependencies, examples below:

image

image

Resolves #24 
Relates to #25


classes/step.php
```diff
@@ -45,12 +45,17 @@ protected static function define_properties(): array {
...
+            //     'null' => NULL_ALLOWED,
+            //     'message' => new \lang_string('invaliddata', 'error'),
+            // ],
         ];
     }
 
     public function __get($name) {
         // Check if it does not exist.
-        if (!isset($this->define_properties()[$name])) {
+        if ($name !== 'id' && !isset($this->define_properties()[$name])) {
```
jaypha, line 58: the property check is already done in get(). It is not needed here.
The same is true for set().

classes/step_form.php
```diff
@@ -34,18 +34,30 @@ class step_form extends \core\form\persistent {
...
         $mform->addElement('text', 'name', get_string('field_name', 'tool_dataflows'));
 
+        // Show a list of other steps as options for depends on.
+        $dataflow = new dataflow($dataflowid);
+        $steps = $dataflow->raw_steps();
+        $options = array_reduce($steps, function ($acc, $step) {
+            $acc[$step->id] = $step->name;
+            return $acc;
+        }, []);
+
```
jaypha, line 57: I suggest a check to exclude the current step from this list.

classes/step.php
keevan: This was added to allow me to use the object like `$dataflow->id` instead of `$dataflow->get('id')`.

Similarly for set() I didn't want to have to set it via the ->set(field, value) method, and since 'id' shouldn't be an explicitly settable property I didn't add that bit there either.

:)

classes/step_form.php
keevan: Thanks @jaypha, this is still WIP code and the plan was to also remove the current step from the list. But your point is correct.

classes/step.php
```diff
@@ -67,4 +72,46 @@ public function __set($name, $value) {
...
+     */
+    protected function validate_dependson($dependencies) {
+        // echo"<pre>";print_r(12223);die;
+        if (false) {
+            return get_string('invaliduserid', 'error');
+        }
+
+        return true;
+    }
 }
```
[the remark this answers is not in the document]
keevan: Ah, is this because internally it only accept strings? Overriding the before_validate feels like it's going against what the persistent model was designed for in a way. :/

classes/step.php
jaypha: That's not what I meant. The `isset($this->define_properties()[$name]` test is not needed as get() and set() perform that test themselves.

classes/step.php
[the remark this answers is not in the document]
jaypha: I know. The moodle code is bad, and this is the best workaround that I could find. Personally I am not a fan of this design pattern (form that maps to a database table). I encountered this pattern when I first started working with websites. In 15 years of working with forms and databases, I have never encountered a situation where I wanted a form that is a direct match to a database. IOW, if it were up to me, I would ditch the persistent form.

classes/step.php
keevan: Yes I understand, however I want to be able to use something like `$step->id` and `$step->name` without doing `$step->get('name')`. If I remove the check, it no longer works.

keevan: Correct, I'll remove this for now. I don't remember why the check was added initially so until that case comes back up, this should simplify things. Thanks Jason.

classes/step.php
```diff
@@ -49,22 +52,87 @@ protected static function define_properties(): array {
...
+    }
+
+    /**
+     * Validate the dependencies
+     *
+     * @param array $dependencies steps that this step is dependent on
+     * @return true|lang_string
+     */
+    protected function validate_dependson($dependencies) {
+        // echo"<pre>";print_r(12223);die;
```
jaypha, line 116: Is this still needed?

lib.php
```diff
@@ -17,12 +17,30 @@
...
+ * @author     Kevin Pham <kevinpham@catalyst-au.net>
+ * @copyright  Catalyst IT, 2022
+ */
+function tool_dataflows_step_types() {
+    return [
+        new step\debugging(),
+        new step\mtrace(),
+        new step\void_step(),
+    ];
+}
```
jaypha, line 46: New line at end of file.

classes/step.php
keevan: Perhaps but I'll re-add it in #25 if that's the case.

jaypha: If you need, then keep it.

PR #60, opened 2022-05-16
25 add step validation

keevan:
- db: add foreign key and step uniqueness constraint on name and flowid
- fix: deps handling for new steps, tidied form and add title / headings
- add input & output stream counts to base class, and add join step type
- test: add dataflow is_dag check and supporting tests

Related to #25

### DIFF
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -17,7 +17,6 @@
 namespace tool_dataflows;
 
 use core\persistent;
-use moodle_exception;
 
 /**
  * Dataflows persistent class
@@ -51,6 +50,70 @@ class dataflow extends persistent {
 
     public function __set($name, $value) {
         return $this->set($name, $value);
+    }
+
+    /**
+     * Validates the dataflow (steps, config, etc)
+     *
+     * This should:
+     * - check if it's in a valid DAG format
+     * - the number of connections (input/output streams) are expected and correct.
+     */
+    public function validate_steps() {
+        global $DB;
+        // Check flows are in valid DAG (no circular dependencies, self referencing, etc).
+        $sql = "SELECT concat(sd.dependson, '|', sd.stepid) as id,
+                       sd.dependson AS src,
+                       sd.stepid AS dest
+                  FROM {tool_dataflows_step_depends} sd
+             LEFT JOIN {tool_dataflows_steps} step ON step.id = sd.stepid
+                 WHERE step.dataflowid = :dataflowid";
+
+        // Note that this works currently because all dependencies are set for each step.
+        $edges = $DB->get_records_sql($sql, ['dataflowid' => $this->id]);
+        // Change this to an array of edges (without the id, keys, etc).
+        $edges = array_map(function ($edge) {
+            return [$edge->src, $edge->dest];
+        }, $edges);
+
+        $isdag = graph::is_dag($edges);
+
+        if ($isdag === false) {
+            return ['dataflowisnotavaliddag' => get_string('dataflowisnotavaliddag', 'tool_dataflows')];
+        }
+
+        // Check if each step is valid based on its own definition of valid (e.g. which could be based on configuration).
+        // TODO: Implement.
+
+        return true;
+    }
+
+
+    /**
+     * Validate Dataflow (steps, dataflow config, etc.)
+     *
+     * Unlike the default validate() method which can not be overriden, this
+     * will also validate any connected steps, configuration and anything
+     * additionally linked with the dataflow.
+     *
+     * @return     true|array true if valid, an array of errors otherwise.
+     * @author     Kevin Pham <kevinpham@catalyst-au.net>
+     * @copyright  Catalyst IT, 2022
+     */
+    public function validate_dataflow() {
+        $stepvalidation = $this->validate_steps();
+        $dataflowvalidation = parent::validate();
+        $errors = [];
+        // If step validation fails, ensure the errors are appended to $errors.
+        if ($stepvalidation !== true) {
+            $errors = array_merge($errors, $stepvalidation);
+        }
+        // If dataflow validation fails, ensure the errors are appended to $errors.
+        if ($dataflowvalidation !== true) {
+            $errors = array_merge($errors, $dataflowvalidation);
+        }
+
+        return empty($errors) ? true : $errors;
     }
 
     /**
@@ -123,7 +186,6 @@ class dataflow extends persistent {
         ]);
         return $steps;
     }
-
 
     /**
      * Method to link another step to this dataflow
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -53,11 +53,13 @@ class dataflow extends persistent {
     }
 
     /**
-     * Validates the dataflow (steps, config, etc)
+     * Validates the steps in the dataflow (steps, step config, etc)
      *
      * This should:
      * - check if it's in a valid DAG format
      * - the number of connections (input/output streams) are expected and correct.
+     *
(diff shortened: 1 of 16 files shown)


classes/dataflow.php
```diff
@@ -53,6 +52,70 @@ public function __set($name, $value) {
...
     }
 
+    /**
+     * Validates the dataflow (steps, config, etc)
+     *
+     * This should:
+     * - check if it's in a valid DAG format
+     * - the number of connections (input/output streams) are expected and correct.
+     */
+    public function validate_steps() {
```
andrewmadden, line 62: Missing return type

classes/dataflow.php
```diff
@@ -53,6 +52,70 @@ public function __set($name, $value) {
...
+        }, $edges);
+
+        $isdag = graph::is_dag($edges);
+
+        if ($isdag === false) {
+            return ['dataflowisnotavaliddag' => get_string('dataflowisnotavaliddag', 'tool_dataflows')];
+        }
+
+        // Check if each step is valid based on its own definition of valid (e.g. which could be based on configuration).
+        // TODO: Implement.
```
andrewmadden, line 86: Should this method be included in the patch if it's not yet implemented?

classes/dataflow.php
```diff
@@ -53,6 +52,70 @@ public function __set($name, $value) {
         return $this->set($name, $value);
     }
 
+    /**
+     * Validates the dataflow (steps, config, etc)
```
andrewmadden, line 56: Was this description duplicated from `validate_dataflow()`?

keevan: Yes, I can clean this up.

classes/dataflow.php
keevan: Typically not, but it's in ALPHA at the moment, so rather than lose track of things, these comments stuck in when committing. 

I can remove them now, they'd be replaced with the implementation in the next few patches anyways. I can do either, but I wouldn't stress.

classes/dataflow.php
```diff
@@ -53,6 +52,70 @@ public function __set($name, $value) {
...
+
+    /**
+     * Validate Dataflow (steps, dataflow config, etc.)
+     *
+     * Unlike the default validate() method which can not be overriden, this
+     * will also validate any connected steps, configuration and anything
+     * additionally linked with the dataflow.
+     *
+     * @return     true|array true if valid, an array of errors otherwise.
+     * @author     Kevin Pham <kevinpham@catalyst-au.net>
```
andrewmadden, line 100: Don't need author etc for function docblock

classes/dataflow.php
andrewmadden: I think better to leave a valid TODO rather than no implementation and no TODO. Fine as is then.

classes/dataflow.php
```diff
@@ -53,6 +52,70 @@ public function __set($name, $value) {
...
+     * @copyright  Catalyst IT, 2022
+     */
+    public function validate_dataflow() {
```
andrewmadden, line 103: Missing return types. Should these have an argument? It seems a little funny to get errors on invalid data without providing anything.

classes/dataflow.php
keevan: Could you please provide a reference for when this should or shouldn't be added? (as I've seen a mixture of them before)

andrewmadden: https://docs.moodle.org/dev/Coding_style#Functions

classes/dataflow.php
keevan: The return types are there, but I can't return the type hint I want because it's mixed between a bool and an array, which is only available in php8+ so I've opted to only have it in the phpdoc above.

classes/dataflow.php
keevan: Thanks @andrewmadden, based on the information you've provided, I'll add another commit to remove them all.

classes/graph.php
```diff
@@ -0,0 +1,102 @@
...
+
+        // Transform to adjacency list.
+        $adjacencylist = [];
+        foreach ($graph as [$src, $dest]) {
+            $adjacencylist[$src][] = $dest;
+        }
+
+        // Perform a depth first search and set and apply various states.
+        foreach (array_keys($adjacencylist) as $src) {
+            if (!isset($discovered[$src])) {
```
andrewmadden, line 54: Isn't $discovered always unset at this point. It's either an empty array or partially populated but $src is unique so would yet have been processed.

classes/graph.php
```diff
@@ -0,0 +1,102 @@
...
+     * Note: the data could be structured as {0, 1}, {1, 2}, {2, 3} using ids,
+     * or names e.g. if planning to validate as part of an import or validating
+     * before storing.
+     *
+     * @param      array $graph array of edges (node connections)
+     * @return     boolean whether or not the edges provided form a valid DAG
+     * @author     Kevin Pham <kevinpham@catalyst-au.net>
+     * @copyright  Catalyst IT, 2022
+     */
+    public static function is_dag($graph) {
```
andrewmadden, line 41: Can add return type to these methods.

classes/step/join.php
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Step type: join
+ *
+ * @package    tool_dataflows
+ * @author     Kevin Pham <kevinpham@catalyst-au.net>
+ * @copyright  Catalyst IT, 2022
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class join extends \tool_dataflows\step\base_step {
```
andrewmadden, line 27: can be shortened to class join extends base_step as it shares namespace

classes/step/join.php
```diff
@@ -0,0 +1,57 @@
...
+     *
+     * For a join step, there should be exactly one output. This is because
+     * without at least one output, there is no need to perform a join.
+     */
+    protected $outputstreams = [1, 1];
+
+    /**
+     * Executes the step
+     *
+     * This will logs the input via debugging and passes the input value as-is to the output.
```
andrewmadden, line 48: typo

db/install.xml
```diff
@@ -31,6 +31,8 @@
       </FIELDS>
       <KEYS>
         <KEY NAME="primary" TYPE="primary" FIELDS="id"/>
+        <KEY NAME="stepname_dataflowid" TYPE="unique" FIELDS="name, dataflowid"/>
```
andrewmadden, line 34: Just double checking we are ok with no upgrade script as this is alpha?

classes/graph.php
keevan: It would appear to be so, but `dfs()` updates this value as well during its call, in particular when it calls upon itself @ https://github.com/catalyst/moodle-tool_dataflows/blob/a70f8fac1b0a69bd3c3f0fb41f29ec76da6592a5/classes/graph.php#L93

I've commented the line you mentioned, re-ran tests and they fail.

db/install.xml
keevan: Yes, this is correct. I believe it will only clutter the upgrade.php at this stage until the fields are good and locked in. The current process would be to uninstall and reinstall this plugin if there were any conflicts.

classes/dataflow.php
keevan: I've added the return type in the phpdoc. As mentioned in the other comment, dual return types are in php8+.

I'm not sure if you want me to add a `mixed` return type, as it's similar (imo) to having an `any` type declaration in typescript - which is not recommended and doesn't really add value.

Please let me know anyways, and this will be made consistent.

Issue #29, opened 2022-04-06
Flow execution manager


brendanheywood:
- Looks at the DAG to find the sinks, leaf nodes in the dag
- Data is always pulled from the end, it isn't pushed from the source(s)

Sinks can be thought of as the ends. but perhaps a more correct way of thinking about them is that they are steps with side effects. Sinks are just another step and they produce output like any other processing step (which is generally unchanged).

If there are multiple sinks, then pull on each in a round robin. This may create asymmetric buffer sizes at tee's but lets ignore that for MVP and assume most use cases will be balanced. In a later version we could store some global stats against pulling from sinks tends to inflate which buffers and then selectively pull from the sinks which reduce the asymmetry.

PR #61, opened 2022-05-17
Issue #29: Basic execution engine.

jaypha:
A very rough prototype for an execution engine.
Currently assumes no T junctions, or connectors as yet.


classes/executor/downstream.php
```diff
@@ -0,0 +1,44 @@
...
+interface downstream {
+    /**
+     * Signals that this flow path has been cancelled.
+     */
+    public function on_cancel(step $step);
+
+    /**
+     * Signals that the step is ready to proceed.
+     */
+    public function on_proceed(step $step, iterator $iterator);
```
keevan, line 38: Perhaps this is better named as on_ready? Or if my understanding is correct, and calling this causes the state to change to FLOWING, perhaps it should be called on_flow instead? (Though it's not exactly 1:1)

classes/executor/iterators/iterator.php
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Iterator interface for use within a dataflow structure.
+ *
+ * @package   tool_dataflows
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
```
keevan, line 27: Can you please remove the unnecessary whitespace here.

classes/executor/iterators/map_iterator.php
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * A mapping iterator that takes in a value, passws it to a function, and then passes it on
+ * to the output.
+ *
+ * @package   tool_dataflows
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
```
keevan, line 28: Can you please remove the unnecessary whitespace here.

keevan (commented):
I know this is still in a draft state, but I had a look over.

Main things I've noticed:
- Some phpdocs are missing info, I've pointed some out.
- Most functions are missing phpdocs, and would be encouraged to add for everything.
- There are many blocks and class properties that don't have supporting comments. If you could add them it would be quicker to grok this information and also when the IDE tries and describes the field.
- Some class names have clashing names as the existing classnames. I'm fine with you changing either, but they should be distinctly different, e.g. step and step, vs step_persistent and step_executor (even if the path has this information, the full path is not always included when declaring/using it). Due to this I've had a difficult time reading the code as I'm not sure when things refer to the base step, the executor step or otherwise. I think the variable names should make this clear as well.

tests/executor/tool_dataflows_basic_execution_test.php
```diff
@@ -0,0 +1,65 @@
...
+ * @package   <insert>
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+
+class tool_dataflows_basic_execution_test extends \advanced_testcase {
+
+    public function test_in_and_out() {
```
keevan, line 34: In layman's terms, can you describe what this test is trying to do? I'm not sure what in_and_out is meant to do and how the following block is used.

```
        array_in_type::$source = json_decode($json);
        array_out_type::$dest = [];
```

tests/executor/tool_dataflows_basic_execution_test.php
```diff
@@ -0,0 +1,65 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_dataflows\executor;
+
+require_once(__DIR__ . "/array_in_type.php"); // This is needed. File will not be automatically included.
+require_once(__DIR__ . "/array_out_type.php"); // This is needed. File will not be automatically included.
```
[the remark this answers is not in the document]
jaypha: These are test step types, not meant for production use.

jaypha:
I have addresses the issues raised. All base functions should now have docs. I haven't added them to overriding functions unless they say something different.
Whitespace issue should be resolved.
Classes have been renamed with an _executor suffix.
More explanatory comments have been added.

tests/executor/tool_dataflows_basic_execution_test.php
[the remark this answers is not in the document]
keevan: It might be better to have steps that we _might_ use in production used for these tests instead. This means more realistic use cases are being tested for. (More realistic tests / use-cases)

Also, not all step types defined are necessarily included in the list of available step types. They need to be added to the list under lib.php file (currently) `function tool_dataflows_step_types() {`. This means it wouldn't be available accidentally in production. (There are controls to prevent production usage if desired, though as the previous point mentioned it probably should be available regardless)

Additionally, using require_once here feels like an anti-pattern. It's hard to know if something is working as expected or broken (because it hasn't been included properly), so I would prefer leaning on the class autoloaders and knowing that it will be included when expected. (Requiring files directly makes the tests less reliable - it assumes it would just work, or the files would also be required_once in places where it is used)

keevan:
I'm getting this in particular, probably missing a method.
` Fatal error: Class tool_dataflows\step\join contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (tool_dataflows\step\base_step::get_iterator) in /var/www/moodle/admin/tool/dataflows/classes/step/join.php on line 0`

And free phpcs fixes you can apply e.g. `phpcbf --ignore=vendor* .` in your PR.

classes/executor/step_executor.php
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * True for flow steps, false for connector steps.
+     *
+     * @return bool
+     */
+    abstract public function is_flow(): bool;
+
+    /**
+     * Initialises teh step.
```
keevan, line 68:
```suggestion
     * Initialises the step.
```

classes/executor/step_executor.php
```diff
@@ -0,0 +1,83 @@
...
+    public $steptype;
+
+    /** @var array  */
+    public $upstreams = [];
+
+    /** @var array  */
+    public $downstreams = [];
+
+    /** @var int  */
+    public $status = dataflow_executor::STATUS_NEW;
```
keevan, line 48: Can you please add a short description for each variable here.

tests/executor/array_out_type.php
```diff
@@ -0,0 +1,46 @@
...
+ * Test writer step type that writes to an array.
+ *
+ * @package   tool_dataflows
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class array_out_type extends base_step {
+
+    /** @var array The output array that the dat is written into. Make sure it is empty before use. */
```
keevan, line 34:
```suggestion
    /** @var array The output array that the data is written into. Make sure it is empty before use. */
```

keevan (commented):
There are various variables that could be better described if possible.

Are you also able to add the relevant coverage information (as this is probably better done now rather than months later - if ever:
```
------------------------------------------------------------------------------------------------------------------------------
------------------------------------------------------------------------------------------------------------------------------
 42 | WARNING | Test method test_in_and_out() is missing any coverage information, own or at class level
------------------------------------------------------------------------------------------------------------------------------
```

tests/executor/tool_dataflows_basic_execution_test.php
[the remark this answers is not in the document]
jaypha: > It might be better to have steps that we might use in production used for these tests instead
When such steps become available, then they can be used. Until then these classes serve to demonstrate that the engine functions.

If you really wan these in the classes/steps directory, I'll move them. But I fail to see any use of them beyond testing.

> Additionally, using require_once here feels like an anti-pattern.
Unfortunately the autoloader doesn't work here. I want and need to be able to define support code for testing, and 'require' is need here. I see nothing wrong with using it.

Issue #65, opened 2022-05-19
Add step id and description fields to DB and table


keevan:
Reasons:
- id - so you can have a name field which is separate to the id field (e.g. `id: read, name: Read user from DB`)
  An additional reason for a separate id field is that it might be used as a reference from other steps, but it should be changed less often than perhaps the name would be. Also it would be good to have this be the unique "key" value when performing the dataflow template export
- description - so you can add more detail to the step and why it has been added
  example: `name: "Call XYZ API", description: "To perform Y and link A to B in system XYZ after user creation"`

brendanheywood:
Maybe id is the wrong thing to call it, perhaps label, name, idnumber or something is better

keevan:
> Maybe id is the wrong thing to call it, perhaps label, name, idnumber or something is better

Yes, for now I've not used this field explicitly but it is referenced as `internalid` within the step class for now. 

Going through the list:
- name is already taken, 
- idnumber makes it sound like it should be a number(?)
- label could work, I had thought of 'reference' at one point which is similar but label is shorter.

The reason for going with id is because other workflow configurations I've used have used id. With the way steps are configured in yaml, there actually isn't a need for a special label for the field (since it takes the value of the step's key).

I'm happy to change this name if the alternative makes it clearer. 

It only needs to be:
- in url-case, 
  - as it needs to be used in a field that would fit in a symfony expression
  - and also as a valid yaml key for exporting configuration/templates
- unique to the dataflow (thus my initial thought of using some 'id')

The name change can be done as another issue if needed. I've already implemented these changes and have gotten them merged in.

PR #66, opened 2022-05-19
add symfony expressions and yaml packages, and basic import support w/ tests

### LINKED ISSUE #19: Import an expression engine, likely symphony
https://symfony.com/doc/current/components/expression_language.html

This will be used just about anywhere that can accept config

keevan:
- feat: add yaml and expression-language support
- db: add internalid and description fields
- feat: add support for new internalid and description fields
- feat: add yaml parsing, import support for dataflows & supporting tests

Resolves #19, #65

Related to #4 - Missing a way to import from the UI, and also an option to export it. Which will come through in another PR.


classes/step.php
```diff
@@ -51,14 +54,35 @@ protected static function define_properties(): array {
...
+     *
+     * @param      string name of the property to update
+     * @param      mixed new value of the property
+     * @return     $this
+     */
     public function __set($name, $value) {
         return $this->set($name, $value);
     }
 
+    protected function set_name($value) {
```
jaypha, line 78: PHPDoc?

Issue #21, opened 2022-04-06
Source sql


brendanheywood:
Easy chunk of sql into recordset

Have an option to store a timestamp of when the query is run, and use that as a parameter in the next run to do delta exports. Should be configurable to the point you could use this with a timestamp, or with any column which is guaranteed to be serial and increasing like an id

Shared test data functionality

Later have the option to use a different db connection

PR #70, opened 2022-05-23
Issue #21: SQL reader

jaypha:
Show how an SQL reader could work.
It extracts its query from the config string.
A cli/test.php file is present to show how it works. It will not be a part of the final submission.

OK to squash.


keevan (commented):
Seems okay, probably better as a phpunit testcase though.

It's also a bit annoying from the default parser, but I think it the configuration should also return in object form, because it makes it easier for expressions to be used consistently.

https://symfony.com/doc/current/components/expression_language.html#passing-in-variables

vs

https://symfony.com/doc/current/components/expression_language/syntax.html#working-with-arrays

jaypha:
Your comments should be addressed now.

classes/step/sql_reader.php
```diff
@@ -0,0 +1,77 @@
...
+                $this->input->close();
+                $this->finished = true;
+            }
+        };
+    }
+
+    /**
+     * Constructs the SQL query from the configuration options.
+     */
+    protected function construct_query(flow_engine_step $step) {
```
keevan, line 59: This function is missing the @param field in the phpdocs

classes/step/base_step.php
```diff
@@ -126,6 +127,32 @@ public function get_number_of_output_streams() {
...
+    /**
+     * Extract the configuration settings from a YAML string.
+     *
+     * @param string $yaml
+     * @return object Parsed configurations as an object.
+     * @throws \moodle_exception
+     */
+    public function extract_config(string $yaml) {
+        $config = Yaml::parse($yaml, Yaml::PARSE_OBJECT_FOR_MAP);
+        if ($this->validate_config($config) !== true) {
```
keevan, line 139: I don't think the validation function call belongs here. This will probably make more sense after my changes, but I think it should be called in the engine validate flow before it continues to process things. Probably not as it's processing the step in the iterator, as this might be a bit late in the process.

classes/step/sql_reader.php
```diff
@@ -0,0 +1,77 @@
...
+ */
+class sql_reader extends flow_step {
+
+    /** @var int[] number of input streams (min, max), zero for readers. */
+    protected $inputstreams = [0, 0];
+
+    /** @var int[] number of output streams (min, max), one for readers. */
+    protected $outputstreams = [1, 1];
+
+    public function get_iterator(flow_engine_step $step): iterator {
```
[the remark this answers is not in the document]
jaypha: fixed

classes/step/base_step.php
jaypha: It should be called upon form validation, as well as importing.

jaypha: extract_config() has been removed as it is not needed.

classes/step/sql_reader.php
jaypha: fixed

lang/en/tool_dataflows.php
```diff
@@ -68,6 +68,8 @@
 $string['stepinvalidinputstreamcount'] = 'Invalid number of input streams for {$a->name} (Found: {$a->found}, Expected between {$a->min} and {$a->max})';
 $string['stepinvalidoutputstreamcount'] = 'Invalid number of output streams for {$a->name} (Found: {$a->found}, Expected between {$a->min} and {$a->max})';
 $string['bad_parameter'] = 'Parameter \'{$a->parameter}\' not supported in \'{$a->classname}\'.';
+$string['sqlnotfound'] = 'Missing \'sql\'';
+$string['invalid_config'] = 'Invalid configuration';
```
keevan, line 72: - It would be better to have the validation / errors keys be consistent. Either all with underscores or without. 
- The `sqlnotfound` string could probably be a generic "Field X is required" and have a way of listing all the required configuration in the step itself. For anything more specific e.g. nested value validation, or query syntax validation, could be a custom lang string, but this feels like it might benefit from being generic.

tests/tool_dataflows_sql_reader_test.php
```diff
@@ -0,0 +1,114 @@
...
+        $writer->name = 'writer';
+        $writer->type = 'tool_dataflows\step\debugging';
+
+        $writer->depends_on([$reader]);
+        $dataflow->add_step($writer);
+
+        // Execute.
+        $engine = new engine($dataflow);
+        $engine->execute();
+        $this->assertDebuggingCalledCount(2, [json_encode($input[0]), json_encode($input[1])]);
```
keevan, line 92: This test seems flaky and might break after a few changes. Keeping in mind the engine needs to eventually support debugging internally and across steps. 

Is there another way to return the instance of the dataflow from the engine, and check if the query results have indeed been processed as expected? (e.g. $instance->writer->total_iterations)

tests/tool_dataflows_sql_reader_test.php
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Test validate_config().
+     *
+     * @covers \tool_dataflows\step\sql_reader::validate_config
+     * @throws \coding_exception
+     */
+    public function test_validate_config() {
+        // Test valid configuration.
+        $config = (object)['sql' => 'arbitrary sql statement'];
```
keevan, line 103: It might be worth having a valid SQL statement here, as this might break down the line only for a valid statement to make it pass again (potentially)

classes/execution/engine.php
```diff
@@ -217,7 +217,8 @@ public function abort(?\Throwable $exception = null) {
             $enginestep->abort();
         }
         $this->status = self::STATUS_ABORTED;
-        // TODO: Do we want to throw the excpetion here, or make it the caller's responsibility?
+        // TODO: We may want to make this the responsibility of the caller.
```
keevan, line 220: Would this TODO still be relevant here as you are now shifting the responsibility to the caller?

jaypha: Actually it performs the throw here. But we may want to change that. This will need to be decided.

lang/en/tool_dataflows.php
jaypha: fixed, now called 'config_field_missing'.

tests/tool_dataflows_sql_reader_test.php
jaypha: fixed

tests/tool_dataflows_sql_reader_test.php
jaypha: As development progresses and more stuff gets added, this can be refactored to use it. Eventually there will be a writer that outputs to an arbitrary stream, and these tests can be changed to use them.

lang/en/tool_dataflows.php
```diff
@@ -68,7 +68,7 @@
 $string['stepinvalidinputstreamcount'] = 'Invalid number of input streams for {$a->name} (Found: {$a->found}, Expected between {$a->min} and {$a->max})';
 $string['stepinvalidoutputstreamcount'] = 'Invalid number of output streams for {$a->name} (Found: {$a->found}, Expected between {$a->min} and {$a->max})';
 $string['bad_parameter'] = 'Parameter \'{$a->parameter}\' not supported in \'{$a->classname}\'.';
-$string['sqlnotfound'] = 'Missing \'sql\'';
+$string['config_field_missing'] = 'Config \'{$a->field}\' missing.';
```
keevan, line 71: This can be simplified by passing `$a` as a string directly, instead of as an array/object.

classes/execution/engine.php
jaypha: An issue is being created to resolve this. For now, please leave it like this as it makes debugging easier.

lang/en/tool_dataflows.php
jaypha: Whattayaknow? You learn something new every day.

Issue #72, opened 2022-05-23
Bug when editing any step and adding dependancies


brendanheywood:

keevan:
Should be fixed in #76

PR #76, opened 2022-05-23
5 expressions and var scopes

keevan:
- fix: add check to ensure invalid dependson value will throw an exception
- feat: add tests and support for expressions & config changes
- feat: add support for updating the dataflow config in the UI
- refactor: change usage of internalid to use the word 'alias' instead

### DIFF
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -38,6 +38,7 @@ class dataflow extends persistent {
     protected static function define_properties(): array {
         return [
             'name' => ['type' => PARAM_TEXT],
+            'config' => ['type' => PARAM_TEXT, 'default' => ''],
             'timecreated' => ['type' => PARAM_INT, 'default' => 0],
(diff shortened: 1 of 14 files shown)


keevan:
Last commit added should resolve #72

classes/dataflow.php
```diff
@@ -38,21 +38,73 @@ class dataflow extends persistent {
...
+     *
+     * @return     array of variables typically passed to the expression parser
+     */
+    public function get_variables(): array {
+        // Test reading a value directly.
+        $variables = [
+            'env' => (object) [
+                'DATAFLOW_ID' => $this->id,
+                'DATAFLOW_RUN_NUMBER' => 0,
+            ],
```
keevan, line 81: This doesn't have to be here, was just using it as an example. Perhaps the dataflow engine should hold these variables instead as it's more relevant after the creation of a dataflow instance.

PR #86, opened 2022-05-24
move and fix page breadcrumb navigation

keevan:
- refactor: move overview.php to index.php
- feat: add breadcrumb navigation to list/info/edit pages

### DIFF
--- a/classes/visualiser.php
+++ b/classes/visualiser.php
@@ -27,6 +27,37 @@ namespace tool_dataflows;
 class visualiser {
 
     /**
+     * Constructs the breadcrumb admin navigation to a nested plugin page
+     *
+     * @param array $crumbs navigation crumbs, in the format of [title, moodle_url] per entry
+     */
+    public static function breadcrumb_navigation($crumbs) {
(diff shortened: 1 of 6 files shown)


classes/visualiser.php
```diff
@@ -26,6 +26,37 @@
...
 
+    /**
+     * Constructs the breadcrumb admin navigation to a nested plugin page
+     *
+     * @param array $crumbs navigation crumbs, in the format of [title, moodle_url] per entry
+     */
+    public static function breadcrumb_navigation($crumbs) {
+        global $PAGE;
+
+        // Admin tools.
```
jaypha, line 37: The crumb trail does not include 'Site Administration' or 'Plugins' links. Is this intentional?

keevan: Yes, this is as intentional as the default crumb of links given by default when going to the index page (Manage Dataflows), which has its bread crumb navigation automatically generated by being listed under the tools section.

image

Also, these items are available already as direct links, I think it would be redundant adding an item for them again.

PR #89, opened 2022-05-24
feat: add functionality to export/import a dataflow + supporting tests

### LINKED ISSUE #4: Export and import stream setups
All processors should have a standard way to store their internal config so we can have a generic way to export and import either a single step or a whole stream.

### LINKED ISSUE #80: Move all dataformat downloads to bottom of the page

### LINKED ISSUE #82: On steps.php remove table sorting and make order explicit
The order should be execution order based on dependencies. There should be a strict deterministic order even for things which are disconnected sub graphs, or with splits and joins

keevan:
Resolves #4 
Resolves #82 

Resolves #80 because the download options are no longer present.

Note: This does NOT include functionality to import/export individual steps, however the capability to do so is there. Just needs a way to wire it up and call it from the UI.

### DIFF
--- a/classes/exportable.php
+++ b/classes/exportable.php
@@ -72,7 +72,6 @@ trait exportable {
         header("Content-Disposition: attachment; filename=\"$filename\"");
     }
 
-
     /**
      * Download the flow/step in YAML format.
      */
--- a/classes/exportable.php
+++ b/classes/exportable.php
@@ -27,6 +27,14 @@ use Symfony\Component\Yaml\Yaml;
  * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 trait exportable {
+
+    /**
+     * Mime type for the downloaded file.
+     *
+     * @var string $mimetype
+     */
+    private $mimetype = 'application/x-yaml';
+
     /**
      * Exports this persistent's data in YAML format.
      *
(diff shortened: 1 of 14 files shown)


classes/exportable.php
```diff
@@ -0,0 +1,102 @@
...
+        } else { // Normal http - prevent caching at all cost.
+            header('Cache-Control: private, must-revalidate, pre-check=0, post-check=0, max-age=0');
+            header('Pragma: no-cache');
+        }
+        header('Expires: '. gmdate('D, d M Y H:i:s', 0) .' GMT');
+        header("Content-Type: $this->mimetype\n");
+        header("Content-Disposition: attachment; filename=\"$filename\"");
+    }
+
+
```
jaypha, line 75: Excess whitespace.

export.php
```diff
@@ -0,0 +1,41 @@
...
+$context = context_system::instance();
+
+// Check for caps.
+require_capability('tool/dataflows:managedataflows', $context);
+
+$dataflowid = required_param('dataflowid', PARAM_INT);
+
+$dataflow = new dataflow($dataflowid);
+$dataflow->download();
+
```
jaypha, line 41: Excess whitespace

lang/en/tool_dataflows.php
```diff
@@ -65,6 +70,8 @@
 $string['new_step'] = 'New Step';
 
 // Warnings / Errors.
+$string['invalidyaml'] = 'The content provided is invalid YAML';
+$string['nodataflowfile'] = 'No dataflow file found';
```
jaypha, line 74: Periods at the end?

keevan:
Just rebased against latest stable branch

Issue #27, opened 2022-04-06
debug logging


brendanheywood:
Each step will have it's own concept of what to do if we are running in dry run mode. Mostly this only applies to sources and targets. A source will read from its test data instead of it's real data. And a target will just spit out something like a log with a curl command rather than making the api call.

At any given point in the progress of a workflow we will keep in memory stats on how many items have streamed through each step. In the summary logs for a workflow run we can show the graph and the number flows.

- [x] Debugging
- [x] Logging

keevan:
Itemized items based on the description above (could be split out and some points may overlap with a different issue)

In dry mode:
- [ ] Source will read from its test data
- [ ] Target will output to log instead of performing the expected write/changes
- [ ] For other steps, it would probably be advised to only perform non permanent changes (e.g. don't insert/update the DB).

Other items which might be split out to a sub task
- [ ] All runs should record and measure the memory usage.
- [ ] Create and store results for a "run"
- [ ] Create a page to display the output/logs for a run

jaypha:
Just how extensive should debugging be?
I currently have it so that a debug message is made for every attempt to execute a step, and also for every iteration on every iterator. For large data sets and complex dataflows, this could result in a TON of debugging messages.

brendanheywood:
I see the dry run mode and the debugging as tangential so maybe we should split those up (see #90 )

Lets focus on the debug logging side for now. The main problem is you don't know what you want to see until something goes wrong in prod and you are trying to figure out what it did and where it busted. So lets focus on getting the mechanics of the debugging done first, the exact detail of what is logged and by what can evolve. I think at minimum we want to know when each connector step executes, and when flows start and stop, and if something barfs which thing barfed. We want generic logs to be emitted by the manager and also the ability for each step to emit its own extra stuff.

I'm on the fence around a log for each iterator record processed but it certainly cant hurt now and we can dial it back as we get closer to shipping. Maybe we want something like an aggregated status every nth record so we still get some sense of progress.

The other thing we need to get right is decoupling emitting logs from where they end up. They might end up in some combination of the cron logs, on screen in dry run mode, possibly persisting to their own log table or to the moodle log table. I think the events which end up in the moodle logs will be very simple, like an event when each dataflow finishes with some overall stats and not much more. The cron logs are probably the least important as they are not useful to an admin in the gui who may not be able to see them.

Echoing above, mvp here is:

* [ ] log interface for each step to use
[...]

PR #100, opened 2022-05-25
Issue #27: Debugging

jaypha:
Add debug statements to the engine execution, and on iterator steps.

dataflow run with debugging

I specifically request comments about the extensiveness of debugging, and how to determine if debugging will occur.
OK to squash.

### DIFF
--- a/classes/execution/engine.php
+++ b/classes/execution/engine.php
@@ -58,6 +58,20 @@ class engine {
     /** @var int Step/dataflow has completely finished. */
     const STATUS_FINALISED = 9;
 
+    /** @var string[] Maps statuses to string indexes. */
+    const STATUS_LABELS = [
+        self::STATUS_NEW => 'engine_status_new',
+        self::STATUS_INITIALISED => 'engine_status_initialised',
+        self::STATUS_BLOCKED => 'engine_status_blocked',
+        self::STATUS_WAITING => 'engine_status_waiting',
+        self::STATUS_PROCESSING => 'engine_status_processing',
+        self::STATUS_FLOWING => 'engine_status_flowing',
+        self::STATUS_FINISHED => 'engine_status_finished',
+        self::STATUS_CANCELLED => 'engine_status_cancelled',
+        self::STATUS_ABORTED => 'engine_status_aborted',
+        self::STATUS_FINALISED => 'engine_status_finalised',
+    ];
+
     /** @var  array The queue of steps to be given a run. */
     public $queue;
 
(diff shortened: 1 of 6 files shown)


keevan (commented):
I suggest changing it from debugging to mtrace as we don't need to see the stack trace that often. I think we'll be selective in where we put debugging statements, but the information provided should be a good starting point for now.

Otherwise I think it's good to merge for now and make to improvements on.

jaypha:
I added a parameter to run.php to provide a trace level. How this gets called is not yet determined.

classes/execution/engine.php
```diff
@@ -17,6 +17,7 @@
 namespace tool_dataflows\execution;
 
 use tool_dataflows\dataflow;
+use tool_dataflows\step\debugging;
```
keevan, line 20: Is this required?

jaypha: Ahh... No. Don't knwo why I put it in.
Removed.

classes/execution/engine.php
```diff
@@ -206,21 +229,38 @@ public function finalise() {
...
         throw $exception;
     }
 
+    /**
+     * Display a trace message for debugging.
+     *
+     * @param string $message Message to display.
+     * @param int $tracelevel The trace level of the message.
+     */
+    public function trace(string $message, int $tracelevel = self::TRACE_ALL) {
```
brendanheywood, line 258: Maybe this should have an optional param which takes a step object instance, and then we have a ->trace() function in the base step class which is what steps use which populates it for free and then we can do some more nuanced output depending on where we end up sending the logs to

jaypha:
I added a trace() function to engine_step.

keevan (commented):
Are you able to please change `trace` to `log` instead? It seems a bit more intuitive as it's currently serving as just a logger.

So two things:
- change `trace()` to `log()`
- change the engine property (boolean) to be `logging` instead (from trace) or similar, to make it more distinguishable from the method name.

keevan (commented):
Okay, so having taken a look at this I think this needs a few small adjustments.

Here are some points of consideration:
- Logging should always be enabled. So perhaps remove the variable.
- Where things are logged will differ depending on how it is run (but it will still be logged):
  - For a dry-run, it will only log to the screen
  - For a manual run (non-dry), it will log to the screen and also to a log table linked to that instance's run.
  - A normal run triggered by webhook/cron/etc, will NOT output to the screen and will log it as above.

I think doing point 1 and enabling logging for now throughout by default will be enough for this PR. And adding a place to store the logs for the other run types mentioned above will be something added in the next PR.

PR #108, opened 2022-05-26
fix: steps listing which was not returning detached nodes with tests

### LINKED ISSUE #105: Not all steps are shown in table
image

keevan:
Resolves #105


keevan:
Checked and made compatible for 7.1 based on the changes below.

https://wiki.php.net/rfc/spread_operator_for_array#advantages_over_array_merge

Also fixed `moodle_database::get_in_or_equal() does not accept empty arrays` error and added a test case specific for it.

Issue #106, opened 2022-05-26
Be able to delete a step


brendanheywood:
* [x] MVP is just delete it which might put the graph into an invalid state #85 
* [x] If it is a flow step with valid ins and outs then remove it and rewire the two cuts ends back together so it remains valid
* [x] If it is a connector with valid ins and outs then remove it and rewire the two cuts ends back together so it remains valid

keevan:
Noting that this Is a duplicate of #85 with additional requirements

brendanheywood:
Lets make this one the more advanced stuff then with the relinking

keevan:
I'm okay with that.

So I'm understanding it correctly, with the join / split steps in mind (points 2 and 3), I'm guessing it should check and since the connection counts post-wiring will be invalid, it will stay detached?

brendanheywood:
If the ins and outs have the same count and type then we can safely rewire them, if count or types are different then we can't so it will just cut it into two sub graphs and force the admin to wire it back correctly.

PR #117, opened 2022-05-26
85 add step deletion

### LINKED ISSUE #85: Add the ability to remove a step
There _might_ be cases where this might not be entirely desired, but the absence of this is also an issue.

Should be shown on the steps (table), and individual step edit page.

Later on perhaps added to the dataflow diagram, but nothing worth considering at this point.

Should redirect to a confirm page before the action is performed.

keevan:
- feat: add ability to remove a step with supporting tests
- feat: add web ui for step deletion

Resolves #85 
Related to #106  - should cover the MVP.

### DIFF
--- a/classes/step.php
+++ b/classes/step.php
@@ -296,4 +296,16 @@ class step extends persistent {
 
         return $yaml;
     }
+
+    /**
+     * Removes any dependencies before removing the step.
+     */
+    protected function before_delete() {
+        global $DB;
+
+        // Remove dependencies other steps have on this step.
+        // NOTE: Later this should check and attempt to rewire/connect steps previously linked to this step.
+        $DB->delete_records('tool_dataflows_step_depends', ['stepid' => $this->id]);
+        $DB->delete_records('tool_dataflows_step_depends', ['dependson' => $this->id]);
+    }
 }
--- a/classes/step.php
+++ b/classes/step.php
@@ -304,7 +304,7 @@ class step extends persistent {
         global $DB;
 
         // Remove dependencies other steps have on this step.
-        // NOTE: Later this should check and attempt to rewire/connect steps previously linked to this step.
+        // TODO: this should check and attempt to rewire/connect steps previously linked to this step.
         $DB->delete_records('tool_dataflows_step_depends', ['stepid' => $this->id]);
         $DB->delete_records('tool_dataflows_step_depends', ['dependson' => $this->id]);
     }
--- a/remove-step.php
+++ b/remove-step.php
@@ -1,0 +1,52 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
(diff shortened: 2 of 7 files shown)


keevan:
image

classes/step.php
```diff
@@ -296,4 +296,16 @@ public function get_export_data(): array {
...
     }
+
+    /**
+     * Removes any dependencies before removing the step.
+     */
+    protected function before_delete() {
+        global $DB;
+
+        // Remove dependencies other steps have on this step.
+        // NOTE: Later this should check and attempt to rewire/connect steps previously linked to this step.
```
brendanheywood, line 307: Micro thing, using // TODO makes all these easier to find and come back to and has ide support

remove-step.php
```diff
@@ -0,0 +1,84 @@
...
+
+visualiser::breadcrumb_navigation([
+    [get_string('pluginmanage', 'tool_dataflows'), new moodle_url('/admin/tool/dataflows/index.php')],
+    [$dataflow->name, new moodle_url('/admin/tool/dataflows/edit.php', ['id' => $dataflow->id])],
+    [get_string('steps', 'tool_dataflows'), $dataflowdetailsurl],
+    [get_string('remove_step', 'tool_dataflows'), $url],
+]);
+
+// The initial request just shows the confirmation page; doing nothing until confirmation.
+if (!$confirm) {
```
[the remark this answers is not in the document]
brendanheywood: and then often these kinda of 'actions' an be distilled down into a tiny case section in the top of the calling page instead of its own page, saves heaps of boilerplate

classes/step.php
keevan: Thanks Brendan, my IDE has support for notes as well, but in this instance TODO makes more sense than a general notice. I'll update this accordingly and make a mental note for future comments. Much appreciated.

remove-step.php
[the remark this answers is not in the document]
keevan: Looking much better now (and less code!). Thanks for the feedback + suggestion.

image

image

Do you think the "Run now" action should behave in a similar way? Or is having its own page okay for that as the output is also displayed on that page? Could argue both ways, then again, could also have the best of both worlds. (Copying a link for a run, and sharing that will display the confirmation - due to missing sesskey - but the button from the dataflow details page should trigger it directly).

remove-step.php
```diff
@@ -0,0 +1,52 @@
...
+ * @copyright  Catalyst IT, 2022
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+use tool_dataflows\step;
+
+require_once(dirname(__FILE__) . '/../../../config.php');
+
+// Prepare and accept page params.
+$stepid = required_param('stepid', PARAM_RAW_TRIMMED);
```
brendanheywood, line 31: can this just be param_int?

keevan: Yes it can. I've fixed this now.

remove-step.php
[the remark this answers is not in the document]
brendanheywood: Good question - In future I can see some trigger steps requiring additional input in order to have the data needed to trigger, so it would be another set of optional formslib defintions which we'd insert on that page. So leave it for the moment me thinks

PR #128, opened 2022-05-27
restructure pages and correct nav

### LINKED ISSUE #109: 'run' page should be a nav child of its flow
There is no way back from the run page to the flow page, add a direct link as well as the crumb hierarchy

### LINKED ISSUE #110: Editing an existing flow page should be a child under the flows view page

### LINKED ISSUE #125: Be able to edit a workflow from the workflow details page

### LINKED ISSUE #129: Dataflows table name to go to details page and add options under manage to edit

keevan:
- refactor: move steps.php to view.php as it displays dataflow details
- fix: run now behaviour for exceptions and nav links
- fix: dataflow edit page nav hierarchy

Resolves #109
Resolves #110 
Resolves #125 
Resolves #129

### DIFF
--- a/edit.php
+++ b/edit.php
@@ -90,16 +90,18 @@ if (($data = $form->get_data())) {
 }
 
 // Display the mandatory header and footer.
+$crumbs = [[get_string('pluginmanage', 'tool_dataflows'), new moodle_url('/admin/tool/dataflows/index.php')]];
 $heading = get_string('new_dataflow', 'tool_dataflows');
 if (isset($persistent)) {
     $heading = get_string('update_dataflow', 'tool_dataflows');
+    $crumbs[] = [$persistent->name, new moodle_url('/admin/tool/dataflows/view.php', ['dataflowid' => $persistent->id])];
 }
+$crumbs[] = [$heading, $url];
 
 // Configure the breadcrumb navigation.
-visualiser::breadcrumb_navigation([
-    [get_string('pluginmanage', 'tool_dataflows'), new moodle_url('/admin/tool/dataflows/index.php')],
-    [$heading, $url],
-]);
+// New    - Dataflows > Manage Flows > New dataflow.
+// Update - Dataflows > Manage Flows > $dataflow->name > Update dataflow.
+visualiser::breadcrumb_navigation($crumbs);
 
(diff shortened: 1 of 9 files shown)


view.php
```diff
@@ -43,7 +43,7 @@
 // Check for caps.
 require_capability('tool/dataflows:managedataflows', $context);
```
brendanheywood, line 44: This page I think should be the root of all data flow pages but it is inconstent:

/admin/tool/dataflows/view.php?dataflowid=2

has a crumb of:

Admin tools / Dataflows / Manage flows / sql + mtrace / Steps

But I think the 'steps' node is redudant. 

If you click on 'sql + mtrace' then you end up in a different place, and you can toggle between the two places clicking on the same crumb node

edit.php?id=2 should be a child node under view.php?dataflowid=2

Also I think view.php?dataflowid=2 should just be view.php?id=2

edit.php
```diff
@@ -90,16 +90,18 @@
 }
 
 // Display the mandatory header and footer.
+$crumbs = [[get_string('pluginmanage', 'tool_dataflows'), new moodle_url('/admin/tool/dataflows/index.php')]];
 $heading = get_string('new_dataflow', 'tool_dataflows');
 if (isset($persistent)) {
     $heading = get_string('update_dataflow', 'tool_dataflows');
+    $crumbs[] = [$persistent->name, new moodle_url('/admin/tool/dataflows/view.php', ['dataflowid' => $persistent->id])];
 }
+$crumbs[] = [$heading, $url];
```
brendanheywood, line 99: If you cancel from edit.php?id=2 then you should end up at the view.php?dataflowid=2 but you end up at /index.php - this needs a return url optional param

keevan:
Relating to #125 

image

edit.php
keevan: Added this in https://github.com/catalyst/moodle-tool_dataflows/pull/128/commits/fe2f73eadfaf1a2bbbc8d37724a6a2ac4167d0e3 which defaults to the index page of dataflows (Overview page)

view.php
keevan: This should be fixed now, going through and doing a retest

PR #137, opened 2022-05-27
feat: add validation for steps (step types) for dataflow details page

### LINKED ISSUE #118: Step specific configurations need to be validated.
Step type classes have a `validate_config()` method which needs to be called at some point.
Flows should not be able to run if the steps do not have valid configurations.
Some means of communicating invalid steps to the user needs to be implemented.

keevan:
Also added a step name prefix to messages to make it stand out which steps are the ones with issues.

Resolves #118


keevan:
Preview of what it looks like:

image

keevan:
NTS: Failing tests due to step types actually being tested against now :')

Will need to refactor the tests to have valid types and configuration (if required) for those types

PR #138, opened 2022-05-27
Implement a stream writer with encoding abilities.

jaypha:
Add encoders for json and csv.

I need something like this for developing dry run, but am submitting this separately, so it can be picked apart separately.

### DIFF
--- a/classes/formats/encoder_base.php
+++ b/classes/formats/encoder_base.php
@@ -1,0 +1,56 @@
+<?php
+// This file is part of Moodle - https://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_dataflows\formats;
+
+/**
+ * Encodes json.
+ *
+ * @package   tool_dataflows
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+abstract class encoder_base {
+    /** @var bool Has data been added yet? */
+    protected $sheetdatadded = false;
+
+    /**
+     * Return the start of the output.
+     *
+     * @return string
+     */
+    public function start_output(): string {
+        return '';
+    }
+
+    /**
+     * Encode a single record
+     *
+     * @param mixed $record
+     * @param int $rownum
+     */
+    abstract public function encode_record($record, int $rownum): string;
+
+    /**
+     * Return the end of the input.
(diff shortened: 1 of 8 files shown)


classes/formats/encoder_base.php
```diff
@@ -0,0 +1,56 @@
...
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_dataflows\formats;
```
brendanheywood, line 17: Moodle already bundles waaay too many different libraries which encode to csv or json and various other things. csv is also not as trivial as you'd think. There is a whole api devoted just to this:

https://docs.moodle.org/dev/Data_formats

it can accept an iterator and a path to write to and a format:

https://github.com/moodle/moodle/blob/master/lib/classes/dataformat.php#L106-L116

So the step config could be configured with whatever dataformat plugins happen to be installed. Have a look here at the one you get for free:

/admin/settings.php?section=managedataformats

BUT there is a big caveat which is that the data format expects the incoming records to be flat, like a csv, or a db recordset, so it won't be able to handle nested json structures. 

At a glance it looks like your csv encoder below won't handle nested json either, it will need to have some sort of config to flatten it, or perhaps just ignore anything nested. If and when this is implemented then you will be able to push it into any data format.

So I think the end game here is 2 different things:
1) we want a special json encoder which can handle anything with nested structures - what you have done here
2) we want a generic dataformat writer which needs to know how to flatten data. Less critical for now. It would be configured with a set of expressions so you know what columns and values get exported. This is issue https://github.com/catalyst/moodle-tool_dataflows/issues/23

So I'll leave it up to you if you want to land this as is and then evolve the csv one into a dataformat one later. It's probably worth a quick look to see how aligned the two apis are to make sure this is viable etc.

jaypha: I tried using the dataformat API first, and found it to have many problems. It is hardwired to output to a browser.

While I could get the JSON encoder to work by wrapping everything in ob_start() and ob_end(), I could not find a way to get the CSV writer to work.

The CSV writer will not get initialised without a call to send_http_headers(). The underlying 'spinput' code was slightly better, but still assumes that you are outputting to a file or the broswer. The code basically sucks.

If one of the other libraries doesn't have this problem, let me know.

brendanheywood: ok fair enough we'll come back to this later, maybe paired with some core trackers to improve that

brendanheywood: in the mean time then, how about the csv exporter simply drops all values which are not a primitive, eg anything nested

jaypha: The core idea is a stream writer with a configurable encoding. The encoding is internal to the step, so can be refactored if needed. If the idea is OK, then I'll flesh it out with unit tests of PHPDocs.

I made the temporary fix to the CSV encodcer.

brendanheywood: all good

classes/formats/encoders/json.php
```diff
@@ -0,0 +1,65 @@
...
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class json extends encoder_base {
+    /**
+     * Return the start of the output.
+     *
+     * @return string
+     */
+    public function start_output(): string {
+        return '[';
```
brendanheywood, line 36: bonus points I think it would be nicer if the json was pretty printed

Issue #145, opened 2022-05-31
Decide on naming conventions for suites of steps and their class names


brendanheywood:
* [x] move all classes to local namespace
* [x] classes should either a prefix like 'source_xyz', 'connector, xyz', or nested ones like flow_filter_xyz
* [x] most of these rough groupings of steps should inherit from a generic class which sets the inputs and outputs
* [x] have methods which say what type or 'group' this step is which will be a language string used to group the steps in the step chooser page
* [ ] we might also need a sort order method for the groups. Both of these will be set in the classes we inherit

https://etherpad.catalyst.net.nz/dataflows-different-types-of-steps

keevan:
Currently worked on in https://github.com/catalyst/moodle-tool_dataflows/pull/146

I've implemented most of the other points, but the 1st and last. The last one I didn't end up doing because I've rendered the ordering of options via a mustache template so it has a fixed order for now, but this can change if needed.

With the first point, what would the correct namespace be? 

For example, `namespace tool_dataflows\step;` is currently used for `class reader_sql extends reader_step`. Should this just nest under another local folder? `namespace tool_dataflows\local\step;` Reference (rule 1 2 and 3 @ https://docs.moodle.org/dev/Coding_style#Namespaces)

brendanheywood:
yes this: namespace tool_dataflows\local\step;

'execution' and 'formats' also should move under local

keevan:
No particular sort for the groups, it's just explicitly shown on the step-chooser page.

If a sort is needed, please open a new issue for it.

PR #146, opened 2022-05-31
Refactor step types

keevan:
Initial changes to classnames and current structure based on https://github.com/catalyst/moodle-tool_dataflows/issues/145

- [x] classes should either a prefix like 'source_xyz', 'connector, xyz', or nested ones like flow_filter_xyz
- [x] most of these rough groupings of steps should inherit from a generic class which sets the inputs and outputs (e.g. flow_step, connector_step will be base classes for those various step types)
- [x] have methods which say what type or 'group' this step is which will be a language string used to group the steps in the step chooser page
image


classes/step/base_step.php
```diff
@@ -40,11 +36,19 @@ abstract class base_step {
      */
     protected $component = 'tool_dataflows';
 
-    /** @var int[] number of input streams (min, max) */
-    protected $inputstreams = [0, 1];
+    /* Input and Output flow and connectors, default all things to zero */
```
brendanheywood, line 39: As changing namespace breaks all the existing test flows lets also move this to local and get it done so we don't have to change it again

templates/dropdown.mustache
```diff
@@ -0,0 +1,107 @@
...
+    * label the text or html content of the main button
+    * hastriggers whether or not the there were triggers provided
+    * hasconnectors whether or not the there were connectors provided
+    * hasflows whether or not the there were flows provided
+    * items collection of known categories of step types (trigger, flow, connectors)
+
+    Example context (json):
+    { "status": "OK"}
+}}
+<span class="badge badge-secondary">{{status}}</span>
```
brendanheywood, line 38: This is pretty broken:
image

We are going to end up with lots of step types so whatever popup we have need to be able to scroll internal like a modal.

Maybe the simplest step is just make the chooser its own page and we can make it js later when we get to filtering the chooser based on various constraints (eg I want to add a step that starts with a flow input)

templates/dropdown.mustache
```diff
@@ -0,0 +1,107 @@
...
+        aria-haspopup="true"
+        aria-expanded="false">
+        {{{label}}}
+    </button>
+    <div class="dropdown-menu" aria-labelledby="tool_dataflows_dropdown">
+        <!-- triggers -->
+        {{#hastriggers}}
+            <h6 class="dropdown-header">{{#str}} stepgrouptriggers, tool_dataflows {{/str}}</h6>
+            {{#items.triggers}}
+            <a class="dropdown-item" href="{{{href}}}">{{label}}</a>
```
brendanheywood, line 55: It doesn't matter that you choose, the selection doesn't end up pin the form

image

templates/dropdown.mustache
keevan: I agree this is odd behaviour. The default bootstrap dropdown actually drops down and doesn't drop... in the middle? It'd be good if there was a dropdown + filter so you can just filter down to the name you want, but another page might work as well.

keevan: Also _note to self_ to fix up the mustache comments/example.

templates/dropdown.mustache
brendanheywood: For the time being lets split it into a new page

keevan:
Step chooser page:
image

remove-step.php
```diff
@@ -23,7 +23,7 @@
  * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 
-use tool_dataflows\step;
+use tool_dataflows\local\step;
```
brendanheywood, line 26: step deletion no longer works:

Exception - Class 'tool_dataflows\local\step' not found

[More information about this error](https://docs.moodle.org/401/en/error/moodle/generalexceptionmessage)

run.php
```diff
@@ -25,7 +25,7 @@
 
 use tool_dataflows\dataflow;
 use tool_dataflows\visualiser;
-use tool_dataflows\execution\engine;
+use tool_dataflows\local\execution\engine;
```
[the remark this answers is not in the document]
brendanheywood: This might be a bug specific the cron trigger step and not a generic bug, maybe this can wait

remove-step.php
keevan: Fixed + pushed

PR #150, opened 2022-06-02
Issue #90: Add support for dryruns.

### LINKED ISSUE #90: Dry run mode
Rough outline: (split from #27)

* [ ] each step will declare if it changes state
* [ ] if it changes state when each step implements its own behavior, possibly with extra config, to say what it should do in dry run mode. 
* [ ] eg an email step could send the email to an alternate email
* [ ] other steps like updating a db could just do nothing, but log what they would have done
* [ ] the same logic applies to source steps, when in dry run mode each decides what they could do. A db reading step could instead just spit out a tiny subset of the data, eg the first N rows. or it could spit out some static data configured in the step setup

jaypha:
Resolves #90 

Add isdryrun property for engine.
Add dryrunstreamname support to writer_stream.
Add EOL to output in JSON encoder.

### DIFF
--- a/classes/local/execution/engine.php
+++ b/classes/local/execution/engine.php
@@ -102,13 +102,11 @@ class engine {
      * @param dataflow $dataflow The dataflow to be executed, as defined in the editor.
      * @param bool $trace Set to true to enable log output.
      */
-    public function __construct(dataflow $dataflow) {
+    public function __construct(dataflow $dataflow, bool $isdryrun = false) {
         $this->dataflow = $dataflow;
 
-        $config = $dataflow->config;
-
         // TODO: For now, extract this from the config, but this should really be taken from global variables.
-        $this->isdryrun = !empty($config->isdryrun);
+        $this->isdryrun = $isdryrun;
 
         // Create engine steps for each step in the dataflow.
         foreach ($dataflow->steps as $stepdef) {
--- a/classes/local/execution/engine.php
+++ b/classes/local/execution/engine.php
@@ -105,7 +105,6 @@ class engine {
     public function __construct(dataflow $dataflow, bool $isdryrun = false) {
         $this->dataflow = $dataflow;
 
-        // TODO: For now, extract this from the config, but this should really be taken from global variables.
         $this->isdryrun = $isdryrun;
 
         // Create engine steps for each step in the dataflow.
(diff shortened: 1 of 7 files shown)


jaypha:
Here is a demo workflow to use it.

[dry-run-demo_20220602_0647.txt](https://github.com/catalyst/moodle-tool_dataflows/files/8821495/dry-run-demo_20220602_0647.txt)

classes/local/execution/engine.php
```diff
@@ -102,6 +105,11 @@ class engine {
     public function __construct(dataflow $dataflow) {
         $this->dataflow = $dataflow;
 
+        $config = $dataflow->config;
+
+        // TODO: For now, extract this from the config, but this should really be taken from global variables.
+        $this->isdryrun = !empty($config->isdryrun);
```
keevan, line 111: Could this instead be something determined when the engine is executed? This would ideally be part of instance variables that are set when the dataflow runnable instance is created and the engine runs (just as an example, the instance id, run number, dry mode, etc)

I think it would be perhaps part of a dropdown menu (split button) with the main run now available, and options to start it with dry mode, etc.
https://getbootstrap.com/docs/5.2/components/dropdowns/#split-button

Though it might be quicker to just add a separate button specifically for dry-runs initially.

tests/tool_dataflows_stream_writer_test.php
```diff
@@ -80,6 +80,9 @@ public function test_writer_json($inputdata) {
 
         $resultdata = file_get_contents($tmpfilename);
 
+        // Strip the whitespace from the result data.
+        $resultdata = str_replace(PHP_EOL, '', $resultdata);
```
keevan, line 84: This test feels like it could break if we applied any additional formatting changes in the future or change the output style. Perhaps it would be enough to ensure the data itself is correct, so maybe comparing the results of the data both re-encoded.

classes/local/execution/engine.php
jaypha: I could not get split buttons to work, so I added a 'dry run' button.

tests/tool_dataflows_stream_writer_test.php
jaypha: Fixed

keevan (approved):
Looks great. I'm merging this in now :+1:

Issue #107, opened 2022-05-26
Style nodes and arrows depending on type


brendanheywood:
All styles should be both a shape and not just a color, but also usually a color

All these styles I've made up on the spot (mostly looking at other engines and charts in the wild) the specifics aren't critical and we can evolve it later. Some of the combinations might be too cluttered so just implement and see and tweak.

* [x] Triggers should be oval / round pill boxes, and blue
* [x] Connectors should be square boxes and grey
* [x] Flow steps should be green and with small rounded corners
* [x] Connectors or flow steps which mutate state in the outside world are parallelograms instead of rectangle
* [x] Arrows connecting connectors should be a solid black arrow
* [x] Arrows Connecting flows should be a dotted arrow
* [ ] Nodes which 
* [x] Nodes which are invalid are shown in red

I think this should be largely automatic and handled by the manager code, but possibly we will need to have a method where a step can add some styles in graphvis format (want to avoid in case we change renderer for the graph)

keevan:
There's one remaining point which might have been unintentional. This issue will close after #152 merges so please feel free to open another issue if you recall the missing item.

PR #151, opened 2022-06-02
74 declare side effects

### LINKED ISSUE #74: Allow steps to declare if they change external state or not
eg if you have a workflow which pulls data, and then does 5 filters on it, then we would consider that a valid workflow. But it doesn't actually do anything. We want to be able to say have a step which updates something, or sends an email etc. These steps would be declared as changing state. A valid workflow must have all finishing steps in the graph change state?? Maybe this is a warning not an error, we will want to run half authored workflows which might not do anything yet.

* [ ] This can apply to flow steps or connector steps
* [ ] show these in the table, maybe something like a lightning icon
* [ ] Show these in the graph visually somehow, maybe a color or dotted border (will also depend on if it is a connector or flow step)

keevan:
- fix: issue if exception is thrown and engine is not defined to log it
- feat: add declaration of side effects and checks
- feat: add basic icon outputs to the different types of steps
- feat: add different styles for different step types in the visual graph

Resolves #74 
Related to #107 - still needs arrow styles

### DIFF
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -312,9 +312,9 @@ class dataflow extends persistent {
     }
 
     /**
-     * Returns a dotscript of the dataflow, false if no connections are available
+     * Returns a dotscript of the dataflow
      *
-     * @return     string dotscript or false if not a valid flow
+     * @return     string dotscript
      */
     public function get_dotscript(): string {
         global $DB;
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -18,6 +18,7 @@ namespace tool_dataflows;
 
 use core\persistent;
 use Symfony\Component\Yaml\Yaml;
+use tool_dataflows\local\step\base_step;
 
 /**
  * Dataflows persistent class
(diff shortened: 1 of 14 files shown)


keevan:
image

Example flow:
[newww_20220602_0706.yml.txt](https://github.com/catalyst/moodle-tool_dataflows/files/8821503/newww_20220602_0706.yml.txt)

keevan:
After most recent push:

image

image

classes/dataflow.php
```diff
@@ -313,38 +314,17 @@ public function validate_dataflow() {
     /**
      * Returns a dotscript of the dataflow, false if no connections are available
      *
-     * @return     string|false dotscript or false if not a valid flow
+     * @return     string dotscript or false if not a valid flow
```
jaypha, line 317: This needs to be updated if the function no longer returns false.

keevan: Ah nice catch. I'll fix this up.

PR #152, opened 2022-06-03
107 arrow styles

### LINKED ISSUE #107: Style nodes and arrows depending on type

keevan:
- feat: add arrow styles based on their connections

Continues from #151

Example

image

Resolves #107

### DIFF
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -18,7 +18,7 @@ namespace tool_dataflows;
 
 use core\persistent;
 use Symfony\Component\Yaml\Yaml;
-use tool_dataflows\local\step\base_step;
+use tool_dataflows\local\step\flow_step;
 
 /**
  * Dataflows persistent class
@@ -317,8 +317,6 @@ class dataflow extends persistent {
      * @return     string dotscript
      */
     public function get_dotscript(): string {
-        global $DB;
-
         // Fetch the dot script node from each step to construct them.
         $steps = $this->steps;
         $nodes = [];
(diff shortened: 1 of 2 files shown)


classes/dataflow.php
```diff
@@ -328,38 +326,50 @@ public function get_dotscript(): string {
...
+                // possible.
+                $localstyles['arrowhead'] = 'none';
+                $localstyles['color'] = 'red';
+            } else {
+                $srcsteptype = new $srcstep->type();
+                $deststeptype = new $deststep->type();
+                if ($srcsteptype instanceof flow_step && $deststeptype instanceof flow_step) {
+                    // If this is a flow to flow, show a dashed/dotted arrow indicating trickling data.
+                    $localstyles['style'] = 'dashed';
+                    $localstyles['color'] = '#008196';
```
jaypha, line 355: I would consider this optional, but
it may be a good idea to have these colours put into the string table, similar to how time formats are in core. Particularly if they are to be reused.

keevan: Ah that sounds like an okay idea. That said, I don't think we'd want to introduce too much customizability with how the dataflow looks this early on.

Perhaps it would be a good idea to have these styles, and similarly for step related styles referenced somewhere more centralised, more graph focused and not just adding more to the persistent class. I think that sort of refactor probably belongs as a separate issue though, as it will touch on the other style bits in the step type classes.

Issue #87, opened 2022-05-24
Cannot save step.


jaypha:
There are a couple issue regarding version compatibility.

When trying to add a step to an empty dataflow, the following error occurs.

`Coding error detected, it must be fixed by a programmer: Unexpected property 'dependson' requested. `
This is specific to v3.5.

jaypha:
I create a dataflow from the manage flows page. I give it some name like 'tutu', and save it.
I then click on the 'steps' column cell, to go to the steps management page.
I click on create new step. I fill in the name, description and alias fields. I leave the type as 'debugging'.
I click save, and kapow!

keevan:
I'm unable to reproduce this issue, can you please see if reinstalling the plugin still causes the issue for you?

![image](https://user-images.githubusercontent.com/9924643/169948927-a74ddc52-c4e4-4a13-929b-1509be508c72.png)

jaypha:
I am using version 3.5 of Moodle. I installed 3.11, and the problem did not occur. This appears to be a problem of older versions of Moodle.

jaypha:
Just noticed that the icon i/upload is not a part of 3.5. It is a part of 3.11.

PR #155, opened 2022-06-06
Issue #87: Fix saving step bug.

jaypha:
Change step.php to create step object after removing 'dependson' field.


keevan (approved):
Looks good, nice find @jaypha

PR #156, opened 2022-06-06
Issue #131: Change run now button to match validation

### LINKED ISSUE #131: If run is valid, make run now green, if invalid make it danger / red as well as disabled

jaypha:
Resolves #131 

Change button colour to be danger for invalid, success for valid.
Change icon to show stop icon when invalid.
Screenshot from 2022-06-06 11-31-28
Screenshot from 2022-06-06 11-31-58

### DIFF
--- a/classes/visualiser.php
+++ b/classes/visualiser.php
@@ -193,7 +193,7 @@ class visualiser {
             $buttoncolour = 'btn-success';
             $buttonicon = 't/go';
         } else {
-            $buttoncolour = 'btn-danger ';
+            $buttoncolour = 'btn-danger';
             $buttonicon = 't/stop';
         }
         $runurl = new \moodle_url(
--- a/classes/visualiser.php
+++ b/classes/visualiser.php
@@ -189,17 +189,24 @@ class visualiser {
         $validation = $dataflow->validate_dataflow();
 
         // Display the run now button (disabling it if dataflow is not valid).
+        if ($validation === true) {
+            $buttoncolour = 'btn-success';
+            $buttonicon = 't/go';
+        } else {
+            $buttoncolour = 'btn-danger ';
+            $buttonicon = 't/stop';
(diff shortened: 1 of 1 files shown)


classes/visualiser.php
```diff
@@ -189,17 +189,24 @@ public static function display_steps_table(int $dataflowid, steps_table $table,
         $validation = $dataflow->validate_dataflow();
 
         // Display the run now button (disabling it if dataflow is not valid).
+        if ($validation === true) {
+            $buttoncolour = 'btn-success';
+            $buttonicon = 't/go';
+        } else {
+            $buttoncolour = 'btn-danger';
+            $buttonicon = 't/stop';
+        }
```
keevan, line 198:
```suggestion
        $buttoncolour = 'btn-danger';
        if ($validation === true) {
            $buttoncolour = 'btn-success';
        }
```
As mentioned in internal comms, I'm not sure about the icon change, seems like it would stop/pause a running dataflow which isn't the intention.

But otherwise the changes look okay.

jaypha: I don't know where the internal comms are, but I think it should not display a 'go' icon if you cannot actually go. Maybe a different icon to what's used here, but not 'go'.

jaypha: Changed it to a 'banned' icon.

keevan: > I don't know where the internal comms are, but I think it should not display a 'go' icon if you cannot actually go. Maybe a different icon to what's used here, but not 'go'.

Internal comms = Matrix. 

That's fine. The reason why I think the icon was fine unchanged was because the disabled state and colour change should have been enough to indicate that this button is the "Run now" button but it has been disabled and cannot be clicked. But it seems you prefer something different which I'm happy with also - so long as it's clear.

keevan (approved):

PR #157, opened 2022-06-06
123 add step custom forms

### LINKED ISSUE #123: Allow each step to define it's own form elements and validation
Most likely three methods
* [ ] define standard form fields for a step
* [ ] a validaiton method
* [ ] optional extra form fields which are only for dry run. This way the fact that a step has any dry run options can be managed centrally

keevan:
- feat: ability to set up custom form inputs per step type
- fix: updating non-custom mstep configuration & prefixed form rel methods

Resolves #123

### DIFF
--- a/classes/local/step/writer_stream.php
+++ b/classes/local/step/writer_stream.php
@@ -16,6 +16,7 @@
 
 namespace tool_dataflows\local\step;
 
+use Symfony\Component\Yaml\Yaml;
 use tool_dataflows\local\execution\flow_engine_step;
 use tool_dataflows\local\execution\iterators\iterator;
 use tool_dataflows\local\execution\iterators\map_iterator;
@@ -33,6 +34,18 @@ use tool_dataflows\local\execution\iterators\map_iterator;
  * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 class writer_stream extends writer_step {
+
+    /**
+     * Return the definition of the fields available in this form.
+     *
+     * @return array
+     */
+    protected static function form_define_fields(): array {
+        return [
+            'streamname' => ['type' => PARAM_TEXT],
+            'format' => ['type' => PARAM_TEXT],
+        ];
+    }
 
     /**
      * Get the iterator for the step, based on configurations.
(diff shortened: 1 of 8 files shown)


keevan:
Example of writer stream step inputs and a submission with invalid data

image

brendanheywood:
Not getting far creating a new step:

/admin/tool/dataflows/step.php?dataflowid=1&type=tool_dataflows%5Clocal%5Cstep%5Cwriter_stream

efault exception handler: Exception - Class '' not found Debug: 
Error code: generalexceptionmessage
* line 142 of /admin/tool/dataflows/classes/form/step_form.php: Error thrown
* line 113 of /lib/classes/form/persistent.php: call to tool_dataflows\\form\\step_form->get_default_data()
* line 85 of /admin/tool/dataflows/step.php: call to core\\form\\persistent->__construct()
, referer: http://moodle.local/admin/tool/dataflows/step-chooser.php?id=1

classes/local/step/writer_stream.php
```diff
@@ -116,16 +128,33 @@ public function abort() {
...
+     * conflicts with any existing fields.
+     *
+     * @param \MoodleQuickForm &$mform
+     */
+    public function form_add_custom_inputs(\MoodleQuickForm &$mform) {
+        // Stream name.
+        $mform->addElement('text', 'config_streamname', get_string('writer_stream:streamname', 'tool_dataflows'));
+
+        // Format.
+        $mform->addElement('text', 'config_format', get_string('writer_stream:format', 'tool_dataflows'));
```
brendanheywood, line 158: Can you please make this a drop down

lang/en/tool_dataflows.php
```diff
@@ -137,3 +137,10 @@
 $string['engine_status_cancelled'] = 'Cancelled';
 $string['engine_status_aborted'] = 'Aborted';
 $string['engine_status_finialised'] = 'Finalised';
+
+// Writer stream.
+$string['writer_stream:streamname'] = 'Stream Name';
```
brendanheywood, line 142: Lets polish this a bit more with both a field name and an example of a field with help text under it with an example like file:///path/to/file.txt

classes/local/step/writer_stream.php
keevan: Yes, but I don't think all the formats are available up-front, and I'm not sure if we want to support (probably) external formats at this current stage.

This feels like it would need to behave in a similar way to the collecting of step types from external plugins - but instead, a collecting of encoders from external plugins.

If this feels like the correct path then I'm happy to go for it.

brendanheywood: There is currently exactly 2 formats, csv and json and they must have code for them here:

https://github.com/catalyst/moodle-tool_dataflows/tree/MOODLE_35_STABLE/classes/local/formats/encoders

If and when we implement a different way of writing using the dataformat api then we can ask that api for the list of formats which are installed and enabled.

keevan: I've added this.

image

lang/en/tool_dataflows.php
keevan: I've added this.

image

Issue #149, opened 2022-06-02
Add validation for connection in out vs flow in out for dataflows


keevan:
Currently this is only checking IO of flows, but it is also assuming the connections are to and from flows. 

This is NOT the case, and it should also check the type of the step (flow vs connector), and also determine its particular connection requirements.

keevan:
As a quick example, this writer step should be an invalid dataflow:

![image](https://user-images.githubusercontent.com/9924643/172280169-af8bd161-0c11-4b72-9587-6b5fd0ed2aa2.png)

PR #159, opened 2022-06-07
Issue #127: Display existing links and link requirements on step edit page

### LINKED ISSUE #127: When editing a step show it linkage expectations
ie if it want's 0-1 connectors, or 0-1 flow steps

Show what is currently linked both in and out

jaypha:
Resolves #127 

Create mustache template for the summary
Add functions to visualizer to generate messages.

### DIFF
--- a/css/step-summary.css
+++ b/css/step-summary.css
@@ -1,0 +1,31 @@
+
+.dataflow-step-summary {
+    width:700px;
+    margin:auto;
+    border: 1px solid #ced4da;
+    padding: 15px;
+    margin-bottom: 15px;
+}
+
+.dataflow-step-summary > div:first-child {
+    display:flex;
+}
+
+.dataflow-step-summary > div:first-child > div {
+    flex-grow: 1;
+    flex-basis: 100px
+}
+
+.dataflow-step-summary > div > div:nth-child(2) {
+    flex-basis: 300px;
+    text-align: center;
+}
+
+.dataflow-step-summary h6 {
+    text-align: center;
+}
+
+.dataflow-step-summary > div:last-child {
+    padding-top: 5px;
+}
+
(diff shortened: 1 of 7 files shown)


jaypha:
Screenshot from 2022-06-07 12-18-50

keevan:
Noting that the check for validation added here, is the check needed to close this issue:

https://github.com/catalyst/moodle-tool_dataflows/issues/149

jaypha:
I'm not sure what you are saying.

keevan:
> I'm not sure what you are saying.

The validation messages and logic that have been added here, e.g. below

image

Is the one missing in the dataflow->validate_steps method, and so #149 was created. If this can be shared that'd be great. If not then it's a loose link such that whomever picks up the related issue can check this and get it working, in particular something similar to the `get_link_expectations` change made in this PR.

css/step-summary.css
```diff
@@ -0,0 +1,31 @@
+
+.dataflow-step-summary {
```
brendanheywood, line 2: wondering if most of this could be done inside a mustache template and use out of the box bootstrap classes instead of custom ones

not critical

jaypha: I did use a mustache template. If there are bootstrap classes to define the layout that would be preferable.

brendanheywood: .. just read to the bottom an realised there is a new template, so yeah ideally move as much of this as possible to bootstrap

Not remotely polished but a rough start:

```
<div class="dataflow-step-summary card shadow p4 center">
    <div class="card-header">
        <div>
            <h6>Inputs</h6>
                <p>No inputs</p>
        </div>
        <div>
            connector_curl<br>
            <span class="text-muted small">(tool_dataflows\local\step\connector_curl)</span>
        </div>
        <div>
            <h6>Outputs</h6>
                <a href="http://moodle.local/admin/tool/dataflows/step.php?id=2">Users</a>
        </div>
    </div>
    <div class="card-body">
        Requires 1 input connector link.<br>
        Requires between 0 and 1 connector flow links.
    </div>
</div>
```

brendanheywood: image

PR #160, opened 2022-06-07
5 variable setting from engine run

keevan:
- feat: add optional query fragments and variable setting for reader_sql
- feat: add expressions support to the dataflow engine and sql reader
- fix: reduce noisy expression parsing by splitting methods
- test: fix/add tests and skip empty evaluated sql expressions

Related to #5

### DIFF
--- a/classes/local/execution/engine_step.php
+++ b/classes/local/execution/engine_step.php
@@ -189,7 +189,13 @@ abstract class engine_step {
      */
     public function set_var($name, $value) {
         // Check if this is an available field to set in the step type.
-        // TODO: implement.
+        if (!$this->steptype->is_field_valid($name)) {
+            $stepurl = new \moodle_url('/admin/tool/dataflows/step.php', ['id' => $this->stepdef->id]);
+            throw new \moodle_exception('variablefieldnotexpected', 'tool_dataflows', $stepurl, (object) [
+                'field' => $name,
+                'steptype' => $this->steptype->get_id(),
+            ]);
+        }
 
         // Check if this field can be updated or not, e.g. if this was forced in config, it should not be updatable.
         // TODO: implement.
--- a/classes/local/execution/engine_step.php
+++ b/classes/local/execution/engine_step.php
@@ -173,4 +173,35 @@ abstract class engine_step {
     public function log(string $message) {
         mtrace('Engine \'' . $this->engine->name . '\', step \'' . $this->name . '\': ' . $message);
     }
+
+    /**
+     * Sets a variable, at the path provided.
+     *
+     * Currently, this will set it into the step's configuration.
+     *
+     * TODO: later this might only set it into the instance's history record,
+     * such that failed execution can be continued without creating a new 'run'.
+     * For example, set_var always goes to instance vars, set_config always
+     * updates step's config.
+     *
+     * @param      string name or path to name of field e.g. 'some.nested.fieldname'
+     * @param      mixed value
+     */
+    public function set_var($name, $value) {
+        // Check if this is an available field to set in the step type.
+        // TODO: implement.
+
+        // Check if this field can be updated or not, e.g. if this was forced in config, it should not be updatable.
+        // TODO: implement.
+
+        $previous = $this->stepdef->config->{$name};
+        $this->log("Setting '$name' to '$value' (from '{$previous}')");
+        $this->stepdef->set_var($name, $value);
+
+        // Persists the variable to the dataflow step config.
+        // NOTE: This is skipped during a dry-run. Variables 'should' still be accessible as per normal.
+        if (!$this->engine->isdryrun) {
+            $this->stepdef->save();
+        }
+    }
 }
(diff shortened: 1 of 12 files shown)


keevan:
Example import file with example query

[jun7_20220607_0725.yml.txt](https://github.com/catalyst/moodle-tool_dataflows/files/8850993/jun7_20220607_0725.yml.txt)

keevan:
Example run screenshot:

image

Noting the following:
- flowcap step name
- any 'var setting' is logged

Example query format:
```
SELECT *
   FROM {user}
where 1=1
[[ AND id > ${{countervalue}} ]]
```

You can substitute other expression values as needed, e.g. step.id, dataflow.steps.somestep.somefield

tests/tool_dataflows_sql_reader_test.php
```diff
@@ -112,4 +112,94 @@ public function test_validate_config() {
...
+     * Test expression usage throughout a engine lifecycle/run
+     *
+     * @covers \tool_dataflows\local\step\reader_sql
+     */
+    public function test_expressions_within_a_run() {
+        global $DB;
+        // The gist:
+        // - Read some records,
+        // - ensure the counter (iterator checkpoint) is set after each iteration, based on & if the field is provided.
+        // - Perform the run again, and this time, expect the record to start from the next value, according to the SQL defined.
```
keevan, line 126: NTS: clean up comments here

classes/local/execution/engine_step.php
```diff
@@ -173,4 +173,35 @@ protected function proceed_status(): int {
...
+     * such that failed execution can be continued without creating a new 'run'.
+     * For example, set_var always goes to instance vars, set_config always
+     * updates step's config.
+     *
+     * @param      string name or path to name of field e.g. 'some.nested.fieldname'
+     * @param      mixed value
+     */
+    public function set_var($name, $value) {
+        // Check if this is an available field to set in the step type.
+        // TODO: implement.
```
keevan, line 192: NTS: Add these checks. Any additional fields should be available in the form fields

classes/local/step/reader_sql.php
```diff
@@ -49,6 +54,7 @@ protected static function form_define_fields(): array {
      * @throws \moodle_exception
      */
     public function get_iterator(flow_engine_step $step): iterator {
+        $this->flowenginestep = $step;
         $query = $this->construct_query($step);
```
jaypha, line 58: This needs to be updated.

classes/local/step/reader_sql.php
```diff
@@ -30,6 +31,8 @@
  */
 class reader_sql extends reader_step {
 
+    private $flowenginestep;
```
jaypha, line 34: If a step type is going to be aware of its engine counterpart, this should probably be implemented in all step types. i.e. I suggest this be put into base_step.

classes/local/execution/iterators/php_iterator.php
```diff
@@ -79,6 +79,15 @@ public function next() {
         if ($this->finished) {
             return false;
         }
+
+        // Only performs this check on the first iteration, and aborts if the
+        // iterator's position is valid (e.g. has no data).
+        if ($this->iterationcount === 0 && !$this->input->valid()) {
+            $this->step->log('Aborting at iteration ' . $this->iterationcount . '. No data?');
+            $this->abort();
```
jaypha, line 87: I'm not certain what your intention is here, but as an FYI, calling an iterator's abort( ) function (I thinking it's a bad choice of name), will not end the dataflow. If you intend to terminate the dataflow, you need to throw an exception. The engine will pick it up and abort all steps and iterators.

keevan: I don't necessarily think an empty flow should terminate the whole dataflow, which is why I've opted for a similar check to the one further down. The issue comes when the value provided by the first iteration is invalid, because the records returned initially is empty.

e.g. instead of something like `{id: 1, name: "Kevin"}` it's returning `{"scalar": []}` or similar

classes/local/step/reader_sql.php
keevan: Can you please provide a bit more context - e.g. what exactly needs to be changed?

jaypha: You removed the parameter from the function definition of construct_query().

classes/local/execution/iterators/php_iterator.php
jaypha: Fair enough, as long as this is intentional, and not an accident.

classes/local/step/reader_sql.php
keevan: Ah thanks. Sorry the github highlighting had the focus on the wrong line, so it was confused.

image

classes/local/execution/iterators/php_iterator.php
brendanheywood: To be clear I expect it to be quite normal for an iterator in real life to have no records and thats fine. eg you might make a curl to grab a delta, turns that into iterator, has nothing in it, and then a connector at the end which sends an email which says 'there was 0 records processed'. At no point will there be an error or abort

PR #161, opened 2022-06-08
feat: add global and dataflow level variable setting

### LINKED ISSUE #5: Allow any processor step to set variables in various scopes
Each a simple processor could count the records that flow through the stream and then set the count in the stream scope which can be access in any expression. These are distinct from record scope variables which change for each record.

Each processor just gets a setter and a name, all variables are stored namespaced under the instance name of the step. A single processor plugin might  have many instances and both sets of vars need to be available. Each step will be given a default serial id but it can be set explicitly to something more human readable. 

There will be multiple 'scopes' that a var can be stored in:

1) 'step' scope / dataflow instance - this is essentially a temporary variable which is only in scope for the duration of a specific dataflow instance. If an instance is suspended, or finishes, then all this state will be persisted so we can see what happened. If suspended then we need the state in order to hibernate the instance and continue it in a new process. There is not an expectation that the state will be continously persisted so it will not be visible from outside the running process. So effectively an in memory set of vars.

2) 'flow' dataflow scope - this is a scope for a dataflow, rather than an instance of a dataflow. So for example if a step wants to do delta imports it must store the last timestamp it processed and that needs to persist across multiple instances of a dataflow. 

3) 'global' scope - these are for variables which span all dataflows. They can be read and written by any step but generally they would be set

Vars can be set by the admin author of a dataflow at all three levels. Vars at any level can also be force set in config.php so they are read only.

There will be namespaces for all vars which applies to all;

a) each step has a unique id within a dataflow, this id will be the namespace for its vars.

eg a step called 'logsql' which does delta's would store call setValue('flow', 'timestamp', now()) and this would end up in

flow.logsql.timestamp =123456789

b) the flow will have it's own namespace of 'config' for vars set by the admin at authoring time

flow.config.foo = bar

c) global vars eg shared connection string params

global.config.hrdb = xyz
global.config.hrpassword = xyz

Thinking these should be another persistent, and we should record who last edited it as well as if it was set from a running instance of manually.

keevan:
- add supporting tests
- reader_sql_variable_setter class contains an example of how it can be set, and the tool_dataflows_variables_test test class shows how it can be retrieved (directly) and via an expression.

Resolves #5

### DIFF
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -76,8 +76,11 @@ class dataflow extends persistent {
      * @return     array of variables typically passed to the expression parser
      */
     public function get_variables(): array {
+        $globalconfig = Yaml::parse(get_config('tool_dataflows', 'config'), Yaml::PARSE_OBJECT_FOR_MAP) ?: new \stdClass;
+
         // Test reading a value directly.
         $variables = [
+            'global' => $globalconfig,
             'env' => (object) [
                 'DATAFLOW_ID' => $this->id,
                 'DATAFLOW_RUN_NUMBER' => 0,
@@ -491,5 +494,22 @@ class dataflow extends persistent {
         }
 
         return $yaml;
+    }
+
+    /**
+     * Updates the value stored in the dataflow's config
+     *
+     * @param  string name or path to name of field e.g. 'some.nested.fieldname'
+     * @param  mixed value
+     */
+    public function set_var($name, $value) {
+        // Grabs the current config.
+        $config = $this->config;
+
+        // Updates the field in question.
+        $config->{$name} = $value;
+
+        // Updates the stored config.
+        $this->config = Yaml::dump((array) $config);
     }
 }
--- a/classes/local/execution/engine.php
+++ b/classes/local/execution/engine.php
@@ -16,6 +16,7 @@
 
 namespace tool_dataflows\local\execution;
 
+use Symfony\Component\Yaml\Yaml;
 use tool_dataflows\dataflow;
 use tool_dataflows\local\step\flow_cap;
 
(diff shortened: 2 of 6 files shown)


classes/local/execution/engine.php
```diff
@@ -297,4 +298,57 @@ public function __get($parameter) {
...
+     * referenced from other steps.
+     *
+     * TODO: add instance support.
+     *
+     * @param      string name of the field
+     * @param      mixed value
+     */
+    public function set_dataflow_var($name, $value) {
+        // Check if this field can be updated or not, e.g. if this was forced in config, it should not be updatable.
+        // TODO: implement.
```
jaypha, line 316: Still relevant?

classes/dataflow.php
```diff
@@ -76,8 +76,11 @@ public function __set($name, $value) {
      * @return     array of variables typically passed to the expression parser
      */
     public function get_variables(): array {
+        $globalconfig = Yaml::parse(get_config('tool_dataflows', 'config'), Yaml::PARSE_OBJECT_FOR_MAP) ?: new \stdClass;
```
jaypha, line 79: This doesn't make sense. Is it correct? I did not notice a unit test for this function.

jaypha: Forget it. I just read bout it in the manual.

classes/local/execution/engine_step.php
```diff
@@ -201,7 +201,7 @@ public function set_var($name, $value) {
         // TODO: implement.
```
jaypha, line 201: Still relevant?

classes/local/execution/engine.php
keevan: Yes. This will need to be updated when we start forcing dataflows via configuration, so the values they might hold can't be updated at any point.

classes/local/execution/engine_step.php
keevan: Yes. As above.

PR #162, opened 2022-06-08
Issue #149: Add link validataions

### LINKED ISSUE #149: Add validation for connection in out vs flow in out for dataflows

### LINKED ISSUE #166: Bug all readers eg sql should have no flow inputs

jaypha:
Resolve #149 
Resolves #166 

Add methods to step to validate input and output links.

### DIFF
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -166,13 +166,26 @@ class dataflow extends persistent {
         $steps = $this->steps;
         $adjacencylist = graph::to_adjacency_list($edges);
 
+        // Prepare an array of steps by id, which makes local lookups instantaneous.
+        $stepidhashmap = [];
+        foreach ($steps as $step) {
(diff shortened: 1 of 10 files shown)


jaypha:
I'm looking for a way to avoid multiple database reads.

jaypha:
Screenshot from 2022-06-09 09-58-47

brendanheywood:
@jaypha can you please also highlight any nodes which are not valid in red here:

https://github.com/catalyst/moodle-tool_dataflows/blob/MOODLE_35_STABLE/classes/local/step/base_step.php#L343-L344

jaypha:
Screenshot from 2022-06-09 14-04-44
Screenshot from 2022-06-09 14-04-58

PR #163, opened 2022-06-08
149 add validation to dataflow details

### LINKED ISSUE #149: Add validation for connection in out vs flow in out for dataflows

keevan:
- fix: add validation for input output counts for connector and flows
- fix: add missing langstrings for new checks
- fix: adjust trigger step input connector count to zero

Resolves #149


keevan:
Closing this one as it will be fixed in https://github.com/catalyst/moodle-tool_dataflows/pull/162 instead

PR #179, opened 2022-06-09
Issue #174: Change to use display names for step types.

### LINKED ISSUE #174: Each step should declare a lang string for its name
* [x] method in base which most steps will use
* [x] steps defined in another plugin will need to override it (or detect the component the step comes from?)
* [x] use this in the chooser page
* [x] use this in the step table, hide the class name and raw name
* [x] on chooser page each step should be its own name inside its own shape

jaypha:
Resolves #174

### DIFF
--- a/classes/steps_table.php
+++ b/classes/steps_table.php
@@ -104,7 +104,6 @@ class steps_table extends \table_sql {
         // Icons relating to the step type and its functions. Currently
         // colours/shapes which will match up closely with the current dataflow
         // diagram.
-        $classname = $record->type;
         $icons = [];
         if (class_exists($classname)) {
             $steptype = new $classname();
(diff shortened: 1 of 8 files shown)


jaypha:
Screenshot from 2022-06-09 11-47-16

classes/steps_table.php
```diff
@@ -104,28 +104,28 @@ public function col_type(\stdClass $record): string {
         // Icons relating to the step type and its functions. Currently
         // colours/shapes which will match up closely with the current dataflow
         // diagram.
-        $classname = $record->type;
         $icons = [];
         if (class_exists($classname)) {
             $steptype = new $classname();
             // Add lightning to all steps with side effects.
             $steptype->has_side_effect() && $icons[] = \html_writer::tag('span', '⚡', [
```
brendanheywood, line 111: I think we can fully ditch the side effect icon here now that it is in the node style

Issue #164, opened 2022-06-08
Have the execution manager populate a suite of variables exposing the inner state of steps 


brendanheywood:
This data will be used for a couple things, but typically it could be used in summary steps at the end which say what happened. It will also be used in some advanced steps later https://github.com/catalyst/moodle-tool_dataflows/issues/120 and https://github.com/catalyst/moodle-tool_dataflows/issues/119

### State transition variables
* [x] every time a step changes state record the state and the time that it changed. All timestamps are microtime

Later issues: 
* make a new color for every state https://github.com/catalyst/moodle-tool_dataflows/issues/53

See also https://github.com/catalyst/moodle-tool_dataflows/issues/165

See also
 #226

brendanheywood:
Right now we have output in the run like this:
```
Engine 'sql': Created
Engine 'sql': Initialised. Dry run: No
Engine 'sql', step 'debug': status Blocked
Engine 'sql', step 'sql': status Waiting
Engine 'sql', step 'cron': status Finished
Engine 'sql', step 'sql': status Flowing
Engine 'sql': Finished
Engine 'sql': Finalised
```

This should be one and the same logic, so once this is implement I'd expect to also see all the timestamps and states in the logs as well as the variables.

Also in that example there is a lot of state transitions missing, ie debug is set to blocked but never flows or finishes. Step 'sql' never finishes.

keevan:
The output has been fixed in #185.

PR #185, opened 2022-06-09
164 add state times on change

keevan:
- feat: wip version of storing timing information for state changes
- fix: refactored engine labels, timing data stored & internal references

Related to #164

### DIFF
--- a/classes/step.php
+++ b/classes/step.php
@@ -117,11 +117,32 @@ class step extends persistent {
         return $this->set($name, $value);
     }
 
-    public function get_variables() {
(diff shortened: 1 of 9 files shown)


classes/step.php
```diff
@@ -83,7 +118,7 @@ public function __set($name, $value) {
     }
 
     public function get_variables() {
-        $dataflow = new dataflow($this->dataflowid);
+        $dataflow = $this->dataflow ?? new dataflow($this->dataflowid);
```
jaypha, line 121: You should probably have a get_dataflow() to handle this.

PR #195, opened 2022-06-10
Issue #190: Enhance abort logging

### LINKED ISSUE #190: Bug in error logging of stream writer
When something goes wrong, it doesn't tell you anything useful like what stream it was trying to open and any details that might help fix it.

```
2022 Jun  9, 16:51:18.343 Engine 'sql': moodle_exception: error/[[unable_to_open_stream]] in /var/www/moodle.local/admin/tool/dataflows/classes/local/step/writer_stream.php:103
Stack trace:
#0 /var/www/moodle.local/admin/tool/dataflows/classes/local/step/writer_stream.php(94): class@anonymous->__construct()
#1 /var/www/moodle.local/admin/tool/dataflows/classes/local/execution/flow_engine_step.php(59): tool_dataflows\local\step\writer_stream->get_iterator()
#2 /var/www/moodle.local/admin/tool/dataflows/classes/local/execution/engine.php(203): tool_dataflows\local\execution\flow_engine_step->go()
#3 /var/www/moodle.local/admin/tool/dataflows/classes/local/execution/engine.php(180): tool_dataflows\local\execution\engine->execute_step()
#4 /var/www/moodle.local/admin/tool/dataflows/run.php(109): tool_dataflows\local\execution\engine->execute()
#5 {main} 
```

Also I think we can improve the way this behaves. If a file can't be written, I would consider this a 'normal' error in that it is the kinda of thing you would typically hit while authoring a flow, or even hit while in production and when something in the inf changes under you.The stack trace might be helpful now but is pretty low value in terms of trying to fix the config of the flow. ie it isn't a code level bug. I would prefer it if in this case for example the step itself emitted a log with something useful and then aborted.

jaypha:
Resolves #190 

Add exception message to engine abort log.
Add log calls to writer_stream when errors occur.
Add extra checks to writer_stream when errors occur.

### DIFF
--- a/lang/en/tool_dataflows.php
+++ b/lang/en/tool_dataflows.php
@@ -151,6 +151,10 @@ $string['invalid_config'] = 'Invalid configuration';
 $string['non_reader_steps_must_have_flow_upstreams'] = 'Non reader steps must have at least one flow step dependency.';
 $string['format_not_supported'] = 'The format \'{$a}\' is not supported.';
 
+// Stream errors.
+$string['writer_stream:failed_to_open_stream'] = 'Failed to open stream "{$a}".';
+$string['writer_stream:failed_to_write_stream'] = 'Failed to write to stream "{$a}".';
+
 // Capabilities.
 $string['dataflows:managedataflows'] = 'Create and configure dataflows';
 $string['dataflows:exportdataflowhistory'] = 'Download / export history of dataflows';
--- a/classes/local/execution/engine.php
+++ b/classes/local/execution/engine.php
@@ -251,7 +251,7 @@ class engine {
             $enginestep->abort();
         }
         $this->status = self::STATUS_ABORTED;
-        $this->log('Aborted');
+        $this->log('Aborted: ' . $exception->getMessage());
         // TODO: We may want to make this the responsibility of the caller.
         throw $exception;
     }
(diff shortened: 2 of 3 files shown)


lang/en/tool_dataflows.php
```diff
@@ -151,6 +151,10 @@
 $string['non_reader_steps_must_have_flow_upstreams'] = 'Non reader steps must have at least one flow step dependency.';
 $string['format_not_supported'] = 'The format \'{$a}\' is not supported.';
 
+// Stream writer errors.
+$string['failed_to_open_stream'] = 'Failed to open stream "{$a}".';
+$string['failed_to_write_stream'] = 'Failed to write to stream "{$a}".';
```
keevan, line 156: I think this belongs under the stream writer group of messages, so perhaps something like `writer_stream:failed_to_open_stream` since it should only be relevant to this particular step type. 

Reason being I can see this potentially growing and I'm not sure whether or not it's better to break them into step type messages, or keep warning/errors in a big single grouped list, as it might not be as easy to determine which ones happen (i.e. haven't been caught) in a particular step type.

jaypha: I'll use the tag 'stream_error:'

Is that an OK compromise?

PR #198, opened 2022-06-10
Issue #189: Fix ordering of steps.

### LINKED ISSUE #189: When creating or editing steps, the depends on field should be in the same order as the step table

jaypha:
Resolves #189 

Refactor step_form to use ordered list.

### DIFF
--- a/classes/form/step_form.php
+++ b/classes/form/step_form.php
@@ -72,13 +72,12 @@ class step_form extends \core\form\persistent {
 
         // Show a list of other steps as options for depends on.
         $dataflow = new dataflow($dataflowid);
-        $steps = $dataflow->raw_steps();
-        $options = array_reduce($steps, function ($acc, $step) {
-            if ((int) $step->id !== $this->get_persistent()->id) {
-                $acc[$step->id] = $step->name;
-            }
-            return $acc;
-        }, []);
+        $steps = $dataflow->steps;
+        $options = [];
+        foreach ($steps as $step) {
+            $options[$step->id] = $step->name;
+        }
+        unset($options[$this->get_persistent()->id]); // We never want a step to depend on itself.
 
         $select = $mform->addElement(
             'select',


keevan (changes requested):
Unfortunately I don't think many of the changes here are needed at all.

`dataflow->steps` should return a list of steps in the order they should be in, if step details are required. For the step summary dependencies, you can probably get the same results using a combination of the `dataflow->edges` and `dataflow->steps` for details.

fwiw, I checked and saw that the step class doesn't have a reference to its connected dataflow yet, but it will soon enough, which should make this a bit less expensive to query.

It's somewhat teased in #185 but it would probably be more ideal to have a get_dataflow which internally handles creation of a dataflow instance if one was not set prior.

I also think by default the order should just be consistent, that also removes the need to pass whether or not we want them ordered or not, and internally, it should be smart about fetching, caching and/or evaluating the order (e.g. if things change), but that's probably a thing we can deal with later on.

classes/step.php
```diff
@@ -209,16 +210,26 @@ public function dependencies() {
...
         ]);
+        if ($ordered) {
+            $dataflow = new dataflow($this->dataflowid);
+            $ordereddeps = [];
+            foreach ($dataflow->step_order as $stepid) {
+                if (isset($deps[$stepid])) {
+                    $ordereddeps[$stepid] = $deps[$stepid];
+                }
+            }
+            return $ordereddeps;
```
keevan, line 221: Potentially, this can be a query which fetches just the step dependencies, and it aggregates the matching deps from `$dataflow->steps` and returns them directly.

classes/form/step_form.php
```diff
@@ -72,13 +72,12 @@ public function definition() {
...
-                $acc[$step->id] = $step->name;
-            }
-            return $acc;
-        }, []);
+        $steps = $dataflow->raw_steps(true);
+        unset($steps[$this->get_persistent()->id]); // We never want a step to depend on itself.
+        $options = [];
+        foreach ($steps as $step) {
+            $options[$step->id] = $step->name;
+        }
```
keevan, line 80: Here it would probably be $dataflow->steps, looping through and adding id and name to options, excluding itself if it exists.

jaypha:
It has been changed to do strictly what was asked in the issue, and uses ->steps so nothing else needs to be changed.

keevan (approved):

PR #200, opened 2022-06-13
Issue #126: Disable workflow flag

### LINKED ISSUE #126: Be able to disable and enable a workflow
What this means is mostly dependent on the implementation of the trigger step, so each trigger will need to honor this in some way
* [x] regardless of how it is triggered it will not run
* [ ] each trigger step can take extra steps to avoid it being run in the first place, eg cron would filter these out in sql
* [x] dry run still works
* [x] manual run still works, eg you'd test stuff manually then when ready enable it to allow the trigger to take over
* [x] until the triggers are implemented later this really just sets a flag and is shown in the master index of workflows as a disabled row cell

Peterburnett:
Closes #126

### DIFF
--- a/classes/dataflows_table.php
+++ b/classes/dataflows_table.php
@@ -150,8 +150,25 @@ class dataflows_table extends \table_sql {
         $exportbtn = str_replace('btn-secondary', 'btn-link bn-sm p-0', $exportbtn);
         $content .= $exportbtn;
 
+        // Display the standard enable and disable icon.
+        if ($record->enabled) {
+            $icon = $OUTPUT->render(new \pix_icon('t/show', get_string('disable'), 'moodle'));
+            $action = 'disable';
+        } else {
+            $icon = $OUTPUT->render(new \pix_icon('t/hide', get_string('enable'), 'moodle'));
+            $action = 'enable';
+        }
+        $url = new \moodle_url('/admin/tool/dataflows/dataflow-action.php',
+            ['id' => $record->id, 'action' => $action, 'sesskey' => sesskey()]);
+        $btnuid = 'enablebuttoncontents';
+        $btn = $OUTPUT->single_button($url, $btnuid, 'post');
+        $enablebtn = str_replace($btnuid, $icon, $btn);
+        $enablebtn = str_replace('btn-secondary', 'btn-link bn-sm p-0', $enablebtn);
+        $content .= $enablebtn;
+
         // Delete dataflow icon.
-        $deleteurl = new \moodle_url('/admin/tool/dataflows/remove-dataflow.php', ['id' => $record->id, 'sesskey' => sesskey()]);
+        $deleteurl = new \moodle_url('/admin/tool/dataflows/dataflow-action.php',
+            ['id' => $record->id, 'action' => 'remove', 'sesskey' => sesskey()]);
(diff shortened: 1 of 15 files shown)


Peterburnett:
Couple notes from developing this in isolation. There was no previous upgrade steps, but I think its at the point now where you dont have to nuke your workflows to upgrade and integrate features, so I felt it was time to start doing upgrade.php insertions.

Also, see comment on the attached issue for some context on the decisions made for the execution decisions.

I also refactored delete-workflows into a shared workflow actions page, because I dont see the point in having seperate URLs for it, as its just rewriting the boilerplate for the permissions and redirection handling, when its much cleaner shared.

Peterburnett:
image

jaypha:
I noticed that, on the dataflows list page, the disabled ones are listed above the enabled ones. Is this intentional? I would have thought it would be the other way around.

classes/dataflows_table.php
```diff
@@ -169,4 +186,17 @@ public function col_manage(\stdClass $record): string {
...
+
+    /**
+     * Get row level classes for output
+     *
+     * @param \stdClass $row
+     * @return string
+     */
+    public function get_row_class($row): string {
+        if (!$row->enabled) {
+            return 'table-dark';
```
brendanheywood, line 198: This is a little too dark for my taste. Maybe wrap the title in <strike> instead?

brendanheywood: image

classes/dataflows_table.php
```diff
@@ -150,8 +150,25 @@ public function col_manage(\stdClass $record): string {
         $exportbtn = str_replace('btn-secondary', 'btn-link bn-sm p-0', $exportbtn);
         $content .= $exportbtn;
 
+        // Display the standard enable and disable icon.
+        if ($record->enabled) {
```
brendanheywood, line 154: Can you please make the buttons consistent, some are links and some are buttons, I reckon they can all be links

image

render glitch due to this:

image

classes/form/dataflow_form.php
```diff
@@ -45,6 +45,10 @@ public function definition() {
         // Name of the dataflow.
         $mform->addElement('text', 'name', get_string('field_name', 'tool_dataflows'));
 
+        // Enable/Disable control.
+        $mform->addElement('advcheckbox', 'enabled', get_string('enabled', 'tool_dataflows'));
+        $mform->setDefault('enabled', 0);
```
brendanheywood, line 50: Really torn on whether they should default on or off. I'm leaning towards on

brendanheywood: So torn, nah leave them off

classes/dataflows_table.php
jaypha: Or you could use text dimming, which is consistent with other plugins like cloudmetrics
Screenshot from 2022-06-14 10-49-47
.

Peterburnett: The reason I didnt initially use text dimming, was because the buttons should still be displayed as usable, because they are, but I think it will be best yeah.

Peterburnett:
> I noticed that, on the dataflows list page, the disabled ones are listed above the enabled ones. Is this intentional? I would have thought it would be the other way around.

This is not the case. Its just the order that the workflows were created in. I have not touched the ordering. Thats out of scope for this change

Peterburnett:
image

Issue #202, opened 2022-06-15
Copy from or to from s3


brendanheywood:
Split out from https://github.com/catalyst/moodle-tool_dataflows/issues/42

Reasons for it being split out:
1) make it easy to only load the sdk when we need to with this step
2) make it more explicit by having an s3 related step
3) we need a place to insert the s3 keys and config (even if optional using IAM)
4) security, this will eventually only be able to read from the sandbox and/or s3 and nowhere else

Tasks:
* [x] step config for s3 keys
* [x] copy from path (can be local or s3)
* [x] copy to path (can be local or s3)

PR #203, opened 2022-06-15
Issue #202: Added S3 copy connector step

Peterburnett:


Peterburnett:
image

Peterburnett:
I had to do a little bit of wiring on this one, as currently there isnt a great mechanism for the connector to get the configuration of the step instance from the raw execute method. I chose to pass the execution context object directly into the execute function, as that way it would be the most up to date object, but I also considered storing the engine_step in the step type at construction time. Happy to hear any thoughts about this.

Issue #204, opened 2022-06-15
Clean up references to enginestep in step types


jaypha:
When I first started putting execution code into step types, I tried to keep it isolated. That has proven unrealistic.
A reference to the engine step is now a class property, but the same reference is passed to several methods within the class.

- [x] Ensure that the enginestep property is set in the get_engine_step() method for all step types. Future proof this.
- [x] Remove all engine step parameters from class methods and use the class property instead.

PR #205, opened 2022-06-15
Issue #94: Add cron trigger step

### LINKED ISSUE #94: Have a workflow triggered at a regular schedule by cron
* [ ] each workflow would have it's own schedule, defined either as a frequency or as a crontab entry
* [ ] a single master cron which runs every minute and figures out which workflows to run / are overdue
* [ ] MVP simple just run them all in serial

jaypha:
Resolves #94 

Flesh out trigger_cron and process_dataflows classes.
Add timestr config to trigger_cron.

### DIFF
--- a/classes/local/execution/engine.php
+++ b/classes/local/execution/engine.php
@@ -114,6 +114,18 @@ class engine {
         $this->isdryrun = $isdryrun;
         $this->automated = $automated;
 
+        // Refuse to run if dataflow is not enabled.
+        if (!($dataflow->enabled || !$automated || $isdryrun)) {
+            $this->abort(new \moodle_exception('running_disabled_dataflow', 'tool_dataflows'));
+            return;
+        }
+
+        // Refuse to run if dataflow is invalid.
+        if ($dataflow->validate_dataflow() !== true) {
+            $this->abort(new \moodle_exception('running_invalid_dataflow', 'tool_dataflows'));
+            return;
+        }
+
         // Create engine steps for each step in the dataflow.
         foreach ($dataflow->steps as $stepdef) {
             $classname = $stepdef->type;
@@ -145,14 +157,20 @@ class engine {
     }
 
     public function initialise() {
-        foreach ($this->enginesteps as $enginestep) {
-            $enginestep->initialise();
-        }
+        try {
+            $this->status_check(self::STATUS_NEW);
 
-        // Add sinks to the execution queue.
-        $this->queue = $this->sinks;
-        $this->set_status(self::STATUS_INITIALISED);
-        $this->log('Initialised. Dry run: ' . ($this->isdryrun ? 'Yes' : 'No'));
+            foreach ($this->enginesteps as $enginestep) {
+                $enginestep->initialise();
+            }
+
+            // Add sinks to the execution queue.
+            $this->queue = $this->sinks;
+            $this->set_status(self::STATUS_INITIALISED);
+            $this->log('Initialised. Dry run: ' . ($this->isdryrun ? 'Yes' : 'No'));
+        } catch (\Throwable $thrown) {
+            $this->abort($thrown);
+        }
     }
 
     /**
@@ -181,21 +199,12 @@ class engine {
      */
     public function execute() {
         $this->initialise();
+        $this->status_check(self::STATUS_INITIALISED);
 
-        // Check the execution conditions to ensure we can safely execute.
-        $execute = $this->dataflow->get('enabled');
-        // If not enabled, we can only execute under certain conditions.
-        if (!$execute) {
-            // We can only execute in a manual (non-automated) run, or a dry run.
-            $execute = !$this->automated || $this->isdryrun;
-        }
-
-        if ($execute) {
-            while ($this->status != self::STATUS_FINISHED) {
-                $this->execute_step();
-                if ($this->status == self::STATUS_ABORTED) {
-                    return;
-                }
+        while ($this->status != self::STATUS_FINISHED) {
+            $this->execute_step();
+            if ($this->status == self::STATUS_ABORTED) {
+                return;
             }
         }
 
@@ -250,28 +259,42 @@ class engine {
      * Finalises the execution. Any remaining resources should be released.
      */
     public function finalise() {
-        foreach ($this->enginesteps as $enginestep) {
-            $enginestep->finalise();
+        try {
+            $this->status_check(self::STATUS_FINISHED);
+
+            foreach ($this->enginesteps as $enginestep) {
+                $enginestep->finalise();
+            }
+            $this->set_status(self::STATUS_FINALISED);
+            $this->log('Finalised');
+        } catch (\Throwable $thrown) {
+            $this->abort($thrown);
         }
-        $this->set_status(self::STATUS_FINALISED);
-        $this->log('Finalised');
     }
 
     /**
      * Stops execution immediately. Gracefully stops all processors and iterators.
      *
-     * @param \Throwable|null $exception
+     * @param \Throwable|null $reason
      * @throws \Throwable
      */
-    public function abort(?\Throwable $exception = null) {
-        $this->exception = $exception;
+    public function abort(?\Throwable $reason = null) {
+        if (!is_null($reason)) {
+            $message = $reason->getMessage();
+        } else {
+            $message = '';
+        }
+        $this->exception = $reason;
         foreach ($this->enginesteps as $enginestep) {
             $enginestep->abort();
         }
         $this->set_status(self::STATUS_ABORTED);
-        $this->log('Aborted: ' . $exception->getMessage());
+        $this->log('Aborted: ' . $message);
+
         // TODO: We may want to make this the responsibility of the caller.
-        throw $exception;
+        if (!is_null($reason)) {
+            throw $reason;
+        }
     }
 
     /**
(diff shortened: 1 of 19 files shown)


classes/local/step/trigger_cron.php
```diff
@@ -33,4 +36,83 @@ class trigger_cron extends trigger_step {
...
+    /**
+     * Validate the configuration settings.
+     *
+     * @param object $config
+     * @return true|\lang_string[] true if valid, an array of errors otherwise
+     */
+    public function validate_config($config) {
+        $errors = [];
+        if (empty($config->timestr)) {
+            $errors['config_timestr'] = get_string('config_field_missing', 'tool_dataflows', 'timestr', true);
```
Peterburnett, line 69: You can do this in form_add_custom_inputs using addRule('required')

classes/local/step/trigger_cron.php
```diff
@@ -33,4 +36,83 @@ class trigger_cron extends trigger_step {
...
+     * @param step $step
+     */
+    public function on_save(step $stepdef) {
+        // The next run timestamp is determined by evaluating the time string using the previous run timestamp as a reference.
+        // If there is no previous timestamp, the the current time is used as the next time instead.
+        $oldtime = scheduler::get_last_time($stepdef->dataflowid);
+        $newtime = $oldtime ? scheduler::determine_next_time(
+            $stepdef->config->timestr,
+            scheduler::get_last_time($stepdef->dataflowid)
+        ) : time();
```
Peterburnett, line 90: Im not sure this logic is correct. If its never been run before, and I have a heavy workflow, how could I schedule it to run at 2am the first time

classes/manager.php
```diff
@@ -61,4 +61,18 @@ public static function get_encoders(): array {
...
+    /**
+     * Return a list of steps of the given type.
+     *
+     * @param string $steptype
+     * @return array
+     */
+    public static function get_steps_of_type(string $steptype): array {
+        global $DB;
+
+        $records = $DB->get_records('tool_dataflows_steps', ['type' => $steptype]);
```
Peterburnett, line 74: nitpick, cleaner if you just `return $DB->get_records`. Personal preference perhaps

classes/task/process_dataflows.php
```diff
@@ -39,6 +45,15 @@ public function get_name() {
      * Processes dataflows.
      */
     public function execute() {
-        // TODO: Process dataflows.
+        $steps = manager::get_steps_of_type('tool_dataflows\local\step\trigger_cron');
+        foreach ($steps as $step) {
```
Peterburnett, line 49: I think this should be delegated to the logic of the trigger_cron step. Eg I think each step should be responsible for figuring its own guts out, and instead this cron task handler should basically be `get all enabled, foreach execute` and thats it. No time logic etc

classes/local/step/connector_step.php
```diff
@@ -62,7 +62,8 @@ public function get_group(): string {
      * @return engine_step
      */
     public function get_engine_step(engine $engine, \tool_dataflows\step $stepdef): engine_step {
-        return new connector_engine_step($engine, $stepdef, $this);
+        $this->enginestep = new connector_engine_step($engine, $stepdef, $this);
```
Peterburnett, line 65: This means we can probably remove the $enginestep parameter in execute due to pass by reference. I wasnt sure if you wanted to take this approach with both connectors and flows, this was a gap in my architecture understanding. I dont think there should be any functional difference, and it might be easier if we centralise all enginestep objects to be all set at constructor time

jaypha: https://github.com/catalyst/moodle-tool_dataflows/issues/204

classes/local/step/trigger_cron.php
jaypha: Thanks for the tip. I will use addRule as it gives a visual aid to the form. However, validation is performed in many places other than form submission (e.g. importing), therefore this check needs to be here as well.

classes/manager.php
jaypha: fixed

classes/local/scheduler.php
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * Manages run schedules for dataflow tasks.
+ *
+ * @package   tool_dataflows
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class scheduler {
+    public const TABLE = 'tool_dataflows_schedule';
```
brendanheywood, line 28: Most triggers should be fairly self contained, but given that cron is going to be the dominant trigger step used in the real world I'm fine with this just being a field in the existing dataflow table. If it isn't a cron triggered workflow it can just be null. Happy either way

classes/task/process_dataflows.php
```diff
@@ -39,6 +45,15 @@ public function get_name() {
...
+            $dataflow = new dataflow($step->dataflowid);
+            if ($dataflow->enabled && scheduler::get_next_time($dataflow->id) < time() && $dataflow->validate() === true) {
```
brendanheywood, line 51: I'd split this logic so if it isn't valid we log something. Or we blindly run it and the engine should validate it and log it instead rather than us. I think we want triggers doing as little as possible so its all shared

classes/task/process_dataflows.php
```diff
@@ -39,6 +45,15 @@ public function get_name() {
      * Processes dataflows.
      */
     public function execute() {
-        // TODO: Process dataflows.
+        $steps = manager::get_steps_of_type('tool_dataflows\local\step\trigger_cron');
```
brendanheywood, line 48: Can you do the time filtering in sql so most of the time it returns an empty set and we move on

classes/task/process_dataflows.php
```diff
@@ -39,6 +45,15 @@ public function get_name() {
...
+        foreach ($steps as $step) {
+            $dataflow = new dataflow($step->dataflowid);
+            if ($dataflow->enabled && scheduler::get_next_time($dataflow->id) < time() && $dataflow->validate() === true) {
+                mtrace("Running dataflow $dataflow->name ($dataflow->id)");
```
brendanheywood, line 52: can you also log what time it was supposed to run which might be much earlier than now

classes/local/scheduler.php
jaypha: Are you suggesting that the schedule table be merged into the dataflow table?

brendanheywood: Yes, If you think it makes sense. I was thinking about weird edge cases that could break this, like in theory you could build a temporarily invalid workflow which had 2 cron trigger steps. But we simply shouldn't allow that kinda of voodoo in the first place. If you want to keep it as two tables then that's also fine.

classes/task/process_dataflows.php
jaypha: I have moved the time logic to the scheduler class, which handles most of the time logic anyway. The validation is now done inside the engine, as requested by Brendan. The task now only checks for enabled.

classes/task/process_dataflows.php
jaypha: It now uses a function in scheduler which filters based on time.

classes/task/process_dataflows.php
jaypha: This has been done.  It now only checks for enabled.

classes/local/execution/engine.php
```diff
@@ -145,6 +157,10 @@ public function __construct(dataflow $dataflow, bool $isdryrun = false, $automat
     }
 
     public function initialise() {
+        if ($this->status !== self::STATUS_NEW) {
+            return;
```
keevan, line 161: Seems odd to want to call this twice, thus the guard. Can we also log a message to indicate this was called and skipped since the status isn't new. Should help point us towards where this would be firing more than once.

classes/local/execution/engine.php
```diff
@@ -250,6 +260,10 @@ public function execute_step() {
      * Finalises the execution. Any remaining resources should be released.
      */
     public function finalise() {
+        if ($this->status !== self::STATUS_FINISHED) {
+            return;
```
keevan, line 264: This also seems like a good time to log something to indicate why it might have returned early. e.g. You typically wouldn't want this to run without the engine being in a finished state, so logging the fact it returned early should help correct any misuse.

classes/local/execution/engine_step.php
```diff
@@ -94,6 +94,7 @@ public function initialise() {
      */
     public function finalise() {
         $this->set_status(engine::STATUS_FINALISED);
+        $this->steptype->on_finalise();
```
keevan, line 97: More of note for self, since I've juggled the idea of adding this to the step type vs an iterator class. 

No action required here on your part.

lang/en/tool_dataflows.php
```diff
@@ -205,3 +209,8 @@
 $string['reader_sql:counterfield_help'] = 'Field in which the counter value is derived from. For example, the userid field in user related query';
 $string['reader_sql:countervalue'] = 'Counter Value';
 $string['reader_sql:countervalue_help'] = 'Current value from the counter field';
+
+// Cron trigger.
+$string['trigger_cron:cron'] = 'Scheduled task';
+$string['trigger_cron:timestr'] = 'Time schedule';
+$string['trigger_cron:timestr_help'] = 'A string compatible with the PHP function strtotime().';
```
keevan, line 216: Can you please add an example of the expected time? 

For example, tomorrow, yesterday, next monday. As I'm not sure the value expected for this field and how it would translate that into the next time it would run.

tests/local/execution/tool_dataflows_engine_test.php
```diff
@@ -174,4 +175,56 @@ public function dataflow_provider(): array {
 
         return [$dataflow, $steps, $expectedsinks, $expectedflowsinks];
     }
+
+    /**
+     * Tests that you cannot run an invalid dataflow.
+     *
+     * @throws \coding_exception
```
keevan, line 182: Probably an IDE thing but not sure if you'd need `@throws` hints in unit tests. There are a few more examples further down and it appears to be more so noise than useful information. 

Do you know if the moodle docs mention that this needs to be included?

classes/local/execution/engine.php
jaypha: It now throws an exception.

classes/local/execution/engine.php
jaypha: These status checks now throw an exception if failed. If this occurs inside an execution it will likely end up resulting in a an abort. Otherwise it is the responsibility of the caller to handle.

lang/en/tool_dataflows.php
jaypha: done

tests/local/execution/tool_dataflows_engine_test.php
jaypha: done

PR #224, opened 2022-06-21
194 support more optional fragments and help text

keevan:
- feat: add support for multiline conditional SQL fragments
- docs: add help text to the SQL input, also providing a short explainer

### DIFF
--- a/classes/local/step/reader_sql.php
+++ b/classes/local/step/reader_sql.php
@@ -153,7 +153,15 @@ class reader_sql extends reader_step {
      */
     public function form_add_custom_inputs(\MoodleQuickForm &$mform) {
         // SQL.
+        $sqlexamples = "
+            SELECT id, username
+              FROM {user}
+          [[ WHERE id > \${{countervalue}} ]]
+          ORDER BY id ASC
+             LIMIT 10";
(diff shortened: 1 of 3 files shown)


classes/local/step/reader_sql.php
```diff
@@ -153,7 +153,14 @@ public function validate_config($config) {
      */
     public function form_add_custom_inputs(\MoodleQuickForm &$mform) {
         // SQL.
+        $sqlexamples = "
+            SELECT id, username
+              FROM {user}
+            [[ WHERE \${{ countervalue }} > id ]]
+             LIMIT 10 ORDER BY id ASC";
```
brendanheywood, line 160: this SQL is broken in a couple ways, this works:
```
SELECT id, username
FROM {user}
[[ WHERE id > ${{countervalue}} ]]
ORDER BY id ASC
LIMIT 3
```

keevan: Ah thanks I see the issue. I'll fix that up.

keevan: Now I've updated it to output the below code, which works when I pasted it back into the flow and ran it.
```
            SELECT id, username
              FROM {user}
          [[ WHERE id > ${{countervalue}} ]]
          ORDER BY id ASC
             LIMIT 10
```

PR #234, opened 2022-06-24
Issue #219: Refactor cron trigger to use crontab format.

### LINKED ISSUE #219: Allow schedules based on a crontab
strtotime works really well for schedules which are only onces a day but at specific times, but it doesn't seem to have any specificity to handle multiple runs within a day and on neat time boundaries. So we need to get cronspec working.

IF it is convenient to use the Moodle cron api then great, but there is no requirement to use it. It would not suprise me if its easy to grab our own library eg this one looks ok

https://github.com/dragonmantank/cron-expression

strtotime can still be convenient in some cases so if it possible to support both at the same time, either two fields or one, then that is preferable. In both cases it should still show the next few expected run times https://github.com/catalyst/moodle-tool_dataflows/issues/215

jaypha:
resolves #219

### DIFF
--- a/classes/local/step/trigger_cron.php
+++ b/classes/local/step/trigger_cron.php
@@ -33,8 +33,28 @@ class trigger_cron extends trigger_step {
 
     protected static function form_define_fields(): array {
         return [
-            'timestr' => ['type' => PARAM_TEXT],
+            'minute' => ['type' => PARAM_TEXT],
+            'hour' => ['type' => PARAM_TEXT],
+            'day' => ['type' => PARAM_TEXT],
+            'month' => ['type' => PARAM_TEXT],
+            'dayofweek' => ['type' => PARAM_TEXT],
+            'disabled' => ['type' => PARAM_TEXT],
         ];
+    }
+
+    /**
+     * Get the default data.
+     *
+     * @return stdClass
+     */
+    public function form_get_default_data(&$data) {
+        parent::form_get_default_data($data);
+        $fields = array('minute', 'hour', 'day', 'month', 'dayofweek');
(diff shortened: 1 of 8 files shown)


classes/local/step/trigger_cron.php
```diff
@@ -47,9 +67,41 @@ protected static function form_define_fields(): array {
...
+            $times = scheduler::get_scheduled_times($this->stepdef->dataflowid);
+            if (!(bool)$this->stepdef->dataflow->enabled) {
+                $nextrun = get_string('trigger_cron:flow_disabled', 'tool_dataflows');
+            } else if ($times->nextruntime > time()) {
+                $nextrun = userdate($times->nextruntime);
+            } else {
+                $nextrun = get_string('asap', 'tool_task');
+            }
+
+            $mform->addElement('static', 'lastrun', get_string('lastruntime', 'tool_task'), $times->lastruntime ? userdate($times->lastruntime) : get_string('never'));
```
brendanheywood, line 82: might be nicer to group all elements together into a single line

https://docs.moodle.org/dev/lib/formslib.php_Form_Definition#addGroup

keevan: Worth mentioning there might be some extra work you encounter if you use grouping, which is mentioned here:

https://docs.moodle.org/dev/Persistent_form#Grouping_fields

Would be interested to know if it affects this in any way.

jaypha: I have done this, but I have not figured out how to get the labels to display.

jaypha: Screenshot from 2022-06-27 12-14-49

PR #235, opened 2022-06-24
170 setup and store runs

### LINKED ISSUE #236: when a run in run, either by trigger or by manual run it will generate a guui

### LINKED ISSUE #237: save a record for this instance when it starts, and update it when it has finished

keevan:
- feat(db): add log and runs dataflow tables
- feat: add run class and to init and entrypoints for updating state
- feat: store run data when the engine is run (not dry)

Resolves #236 
Resolves #237

### DIFF
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -194,14 +194,21 @@ class dataflow extends persistent {
             }
         }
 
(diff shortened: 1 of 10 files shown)


keevan:
Privacy tests are red, I'll fix them up.

PR #238, opened 2022-06-26
Issue #225: Remove test line that was causing a value mismatch.

### LINKED ISSUE #225: Issue with scheduling test, which causes CI to fail sometimes
https://github.com/catalyst/moodle-tool_dataflows/runs/6980502857?check_suite_focus=true#step:3:2039
```
There was 1 failure:
1) tool_dataflows\tool_dataflows_scheduler_test::test_get_scheduled_time
Failed asserting that 1655799581 matches expected 1655799580.
```

jaypha:
Resolves #225

### DIFF
--- a/classes/local/scheduler.php
+++ b/classes/local/scheduler.php
@@ -42,64 +42,6 @@ class scheduler {
     }
 
     /**
-     * Get the last time a dataflow was scheduled to run.
-     *
-     * @param int $dataflowid
-     * @return int Timestamp. Defaults to zero (effectively never) if no entry exists.
-     * @throws \dml_exception
-     */
-    public static function get_last_scheduled_time(int $dataflowid): int {
-        global $DB;
-
-        $record = $DB->get_record( self::TABLE, ['dataflowid' => $dataflowid]);
-        if ($record === false) {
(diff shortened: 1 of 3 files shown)


keevan (commented):
Okay I've had a review of the code involved. I think there are 2 approaches we can make, I'll flag the one I'm leaning towards.

Background: in scheduler::get_next_scheduled_time (admin/tool/dataflows/classes/local/scheduler.php), the function accepts a dataflowid and returns the next run time otherwise returns the current timestamp using the time() function. Checking all usage, there is only one place which calls this method.

So these are the 2 alternative approaches I'm thinking of:
- Pass a second optional parameter to the next scheduled time function, which sets/returns the time the next run should happen, if the next scheduled run time hadn't already been set.
- Or more simply, always return the time scheduled for, otherwise return NULL, and have the caller implement what would be the fallback or handling should be if the function didn't return a non-null value. (currently leaning towards this one based on amount of usage)

Either approach would work, the 2nd might be simpler in this case, and would not be over-engineering it like alternative 1 would suggest doing, since we don't have a real need for it yet other than passing in the current timestamp.

jaypha:
The relevant functions have become redundant and were removed, fixing the problem. This PR also does a bit of cleanup.

PR #241, opened 2022-06-27
170 ui for runs

### LINKED ISSUE #170: Saved metadata on run instances
This is a precursor to the full logging

* [x] #236
* [x] #237
* [ ] make a basic interface for seeing the run instances of a workflow and drill into them to see when it ran and its timing info

Probably makes sense to do this after https://github.com/catalyst/moodle-tool_dataflows/issues/187

keevan:
Added UI for dataflow runs, fixing some issues with data stored and added some (more) color to the new pages.

Testing colors for now, each run represents a different state. Can be tweaked later.

Dataflow details page:
image

Color Palette:
image

Runs view page:
image

All runs page:

image

NTS: 
- [x] fix page breadcrumbs
- [x] fix data fill issues with start / finish states
- [x] fix all runs default display (css rules,prioritisation), should be more like:
image

Resolves #170

### DIFF
--- a/classes/runs_table.php
+++ b/classes/runs_table.php
@@ -112,9 +112,7 @@ class runs_table extends \table_sql {
      * @return string
      */
     public function col_timestarted(\stdClass $record): string {
-        $defaulttimezone = date_default_timezone_get();
-        $formatteddate = date_format_string($record->timestarted, get_string('strftimedatetimeaccurate'), $defaulttimezone);
-        return $formatteddate;
+        return userdate($record->timestarted, get_string('strftimedatetimeaccurate'));
     }
 
     /**
@@ -124,9 +122,7 @@ class runs_table extends \table_sql {
      * @return string
      */
     public function col_timefinished(\stdClass $record): string {
-        $defaulttimezone = date_default_timezone_get();
-        $formatteddate = date_format_string($record->timefinished, get_string('strftimedatetimeaccurate'), $defaulttimezone);
-        return $formatteddate;
+        return userdate($record->timefinished, get_string('strftimedatetimeaccurate'));
     }
 
     /**
--- a/classes/runs_table.php
+++ b/classes/runs_table.php
@@ -1,0 +1,142 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_dataflows;
+
+use tool_dataflows\local\execution\engine;
+
+/**
+ * Display a table runs for a particular dataflow.
+ *
+ * @package    tool_dataflows
+ * @author     Kevin Pham <kevinpham@catalyst-au.net>
+ * @copyright  Catalyst IT, 2022
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class runs_table extends \table_sql {
+
+    const COLUMNS = [
+        'name',
+        'userid',
+        'status',
+        'timestarted',
+        'timefinished',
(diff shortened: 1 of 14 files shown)


keevan:
Worth mentioning if the run aborts, errors and the such, are not logged against the run record, but will be shown added in the logs table at a later stage:

image

db/install.xml
```diff
@@ -73,9 +73,9 @@
         <FIELD NAME="dataflowid" TYPE="int" LENGTH="10" NOTNULL="false" SEQUENCE="false" COMMENT="This is the dataflow that this run is originally based on."/>
         <FIELD NAME="userid" TYPE="int" LENGTH="10" NOTNULL="false" SEQUENCE="false" COMMENT="The user who started this particular run"/>
         <FIELD NAME="status" TYPE="char" LENGTH="10" NOTNULL="false" SEQUENCE="false" COMMENT="The current state of the run as a string, e.g. running, pausing, paused, finished, etc."/>
-        <FIELD NAME="timestarted" TYPE="int" LENGTH="11" NOTNULL="false" SEQUENCE="false" COMMENT="Timestamp of when the run started"/>
-        <FIELD NAME="timepaused" TYPE="int" LENGTH="11" NOTNULL="false" SEQUENCE="false"/>
-        <FIELD NAME="timefinished" TYPE="int" LENGTH="11" NOTNULL="false" SEQUENCE="false"/>
+        <FIELD NAME="timestarted" TYPE="number" LENGTH="14" NOTNULL="false" SEQUENCE="false" DECIMALS="4" COMMENT="Timestamp of when the run started"/>
```
brendanheywood, line 76: why is this non upgradable?

keevan: https://user-images.githubusercontent.com/9924643/176653277-f3d97402-8892-4b6c-afba-78c0810a0b36.mp4

I can add an upgrade script for this, and a version bump too (whilst I'm at it)

lang/en/tool_dataflows.php
```diff
@@ -70,6 +73,17 @@
...
 $string['steps'] = 'Steps';
 
+// Dataflows runs (table).
+$string['run'] = 'Run';
+$string['runs'] = 'Runs';
+$string['field_status'] = 'Status';
+$string['field_timestarted'] = 'Started';
+$string['field_timefinished'] = 'Finished';
+$string['field_duration'] = 'Duration';
+$string['startstate'] = 'Start state';
```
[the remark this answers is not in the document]
keevan: The fields due to how the tables have been set up, are somewhat fixed to be `field_<name_of_field>`. This is consistent with the table classes we currently use.

In retrospect, `startstate` on the other hand might be better under field_startstate (and similarly the others), since they belong to the run table & persistent classes.

db/install.xml
keevan: `\d mdl_tool_dataflows_runs` output:

Before
```
timestarted  | bigint                |           |          | 
timepaused   | bigint                |           |          | 
timefinished | bigint                |           |          | 
```

After:
```
timestarted  | numeric(14,4)         |           |          |
timepaused   | numeric(14,4)         |           |          |
timefinished | numeric(14,4)         |           |          |
```

Seems to have worked okay.

classes/runs_table.php
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    /**
+     * Return the time started as a formatted datetime string
+     *
+     * @param \stdClass $record
+     * @return string
+     */
+    public function col_timestarted(\stdClass $record): string {
+        $defaulttimezone = date_default_timezone_get();
```
brendanheywood, line 115: userdate

classes/runs_table.php
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    /**
+     * Return the time finished as a formatted datetime string
+     *
+     * @param \stdClass $record
+     * @return string
+     */
+    public function col_timefinished(\stdClass $record): string {
+        $defaulttimezone = date_default_timezone_get();
```
brendanheywood, line 127: userdate

classes/runs_table.php
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    /**
+     * Return the duration in seconds of the run.
+     *
+     * @param \stdClass $record
+     * @return string
+     */
+    public function col_duration(\stdClass $record): string {
+        $totalsecs = number_format($record->timefinished - $record->timestarted, 4);
```
brendanheywood, line 139: 4 decimal points is a little odd, I think its better to stick with either milliseconds or microseconds 3 or 6, in this case 3 is fine

keevan: Reason for this was the output of microtime(true) seems to always output to an accuracy of 4 decimals. 

I've updated the display in the table to only show 3 decimals to get this over the line.

PR #243, opened 2022-06-28
Issue #215: Add more visibility for cron triggers.

### LINKED ISSUE #215: Improve cron trigger visibilty
* [x] show in the UI in one or more places when it will run next
* [x] editing the trigger step in the help text show then next time it is supposed to run
* [x] because cron will be 90% of the use cases I think it is a special case and could be shown in the master list of all flows like 'Will run in 15 ins' 'Was supposed to run 2 mins ago'

jaypha:
Resolves #215 
Add column 'Extra' to dataflow and step tables.
Add get_extra_info() to step type classes.
Get trigger_cron::get_extra_info() to return next scheduled run time.
Fill 'extra' column with content returned by get_extra_info().


jaypha:
Screenshot from 2022-06-28 12-38-55
Screenshot from 2022-06-28 12-39-12

jaypha:
If anyone has a better name than 'Extra', I'm all ears.

brendanheywood:
"Details" might be better? rest of the code looks great and merge at will

Issue #167, opened 2022-06-08
Prod ready curl step


brendanheywood:
If it is easier then break this up into discrete commits / prs, in roughly this order / groups:

group 1

* [x] should have no flow inputs or outputs
* [x] optional connectors in or out, it could be the start and download a file, or the end where it uploads one
* [x] has config for the url to curl
* [x] has config for the file to download it to, if not present discards output

group 2

* [x] mvp is if the response is 40x or 50x then send an abort to the execution manager. later improve this
* [x] has config textarea for headers, raw key: value roughly exactly as it goes over the wire. can contain expressions

group 3 

* [x] has config for the get / post / put method
* [x] has config for a text area which is raw post data
* [x] has a config checkbox which specifies if this has side effects. If it is post or put then it overrides this and will have side effects

group 4

* [x] after it executes populates variables with:
* [x] the http error code and http response eg '404' and 'content not found'
* [x] a variable with the time it took to connect and download
* [x] a variable with the bytes sent
* [x] logs errors like ssl errors or timeouts

marcghaly:
Hi @brendanheywood,

I have not clearly understood what you meant on unchecked point `mvp is if the response is 40x or 50x then send an abort to the execution manager. later improve this`, also should there be a specific timeout for curl call ?

Concerning file download, should it be same step or another one and what should be the download location ?

https://github.com/catalyst/moodle-tool_dataflows/commit/f7321a60181a4d86e16be92199411d71afca6189#diff-cf8d12918f4729abaeec73d8b5c28ddb1d00c306ee84d79ce193c5ea0dc8308aR114

Regards,

Marc

keevan:
Hi @marcghaly, regarding the later point of file download. I believe the _download_ behaviour should be in the same step. The download location could be configurable or fixed. But it would be ideal to work out of the scratch directory mentioned here https://github.com/catalyst/moodle-tool_dataflows/issues/158 . I think fixing it initially might make the intentions clearer when working with the dataflow at least initially.

There would then be another reader step that lets this data _flow_ across into other flow steps (if it's an array of data). If it's a file download that's not iterable (e.g. a picture) another connector step can handle it.

Not sure if this was mentioned but if the cURL request was to GET e.g. user information so it's data for a single record in JSON format, then perhaps instead of an output path (to avoid file io), it could store the response in the step as an instance variable (instead of discarded), and can be referenced by other steps later if needed using an expression.

brendanheywood:
[...]

PR #245, opened 2022-06-29
Issue #167: curl step

marcghaly:
Hi @brendanheywood,

This is PR for issue #167,

Regards,

Marc

### DIFF
--- a/classes/local/step/connector_curl.php
+++ b/classes/local/step/connector_curl.php
@@ -16,12 +16,14 @@
 
 namespace tool_dataflows\local\step;
 use tool_dataflows\local\execution\engine_step;
+use Symfony\Component\Yaml\Yaml;
 
 /**
  * CURL connector step type
  *
  * @package    tool_dataflows
  * @author     Kevin Pham <kevinpham@catalyst-au.net>
+ * @author     Ghaly Marc-Alexandre <marc-alexandreghaly@catalyst-ca.net>
  * @copyright  Catalyst IT, 2022
  * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
@@ -30,15 +32,180 @@ class connector_curl extends connector_step {
     /** @var bool whether or not this step type (potentially) contains a side effect or not */
     protected $hassideeffect = true;
 
+    /** @var int Time after which curl request is aborted */
+    protected $timeout = 60;
+
+    /**
+     * Return the definition of the fields available in this form.
+     *
+     * @return array
+     */
+    protected static function form_define_fields(): array {
+        return [
+            'curl' => ['type' => PARAM_URL],
+            'destination' => ['type' => PARAM_PATH],
+            'headers' => ['type' => PARAM_RAW],
+            'method' => ['type' => PARAM_TEXT],
+            'rawpostdata' => ['type' => PARAM_RAW],
+            'sideeffects' => ['type' => PARAM_RAW],
+            'timeout' => ['type' => PARAM_INT],
+        ];
+    }
+
+    /**
+     * Allows each step type to determine a list of optional/required form
+     * inputs for their configuration
+     *
+     * It's recommended you prefix the additional config related fields to avoid
+     * conflicts with any existing fields.
+     *
+     * @param \MoodleQuickForm &$mform
+     */
+    public function form_add_custom_inputs(\MoodleQuickForm &$mform) {
+        $jsonexample = [
+            'header-key' => 'header value',
+            'another-key' => '1234'
+        ];
+        $ex['json'] = \html_writer::nonempty_tag('pre', json_encode($jsonexample, JSON_PRETTY_PRINT));
+        $ex['yaml'] = '<pre>header-key: header value<br>another-key: 1234</pre>';
+        $mform->addElement('text', 'config_curl', get_string('connector_curl:curl', 'tool_dataflows'),
+                ['cols' => 50, 'rows' => 7]);
+        $mform->addElement('text', 'config_destination', get_string('connector_curl:destination', 'tool_dataflows'));
+        $mform->addHelpButton('config_destination', 'connector_curl:destination', 'tool_dataflows');
+        $mform->addElement('textarea', 'config_headers', get_string('connector_curl:headers', 'tool_dataflows'),
+                ['cols' => 50, 'rows' => 7]);
+        $mform->addElement('static', 'headers_help', '', get_string('connector_curl:field_headers_help', 'tool_dataflows', $ex));
+        $mform->addElement('select', 'config_method', get_string('connector_curl:method', 'tool_dataflows'),
+                ['get' => 'GET', 'post' => 'POST', 'put' => 'PUT']);
+        $mform->addElement('textarea' , 'config_rawpostdata', get_string('connector_curl:rawpostdata', 'tool_dataflows'),
+                ['cols' => 50, 'rows' => 7]);
+        $mform->addElement('checkbox', 'config_sideeffects', get_string('connector_curl:sideeffects', 'tool_dataflows'),
+                get_string('yes'));
+        $mform->hideIf('config_rawpostdata', 'config_method', 'eq', 'get');
+        $mform->disabledIf('config_rawpostdata', 'config_method', 'eq', 'get');
+        $mform->addElement('text', 'config_timeout', get_string('connector_curl:timeout', 'tool_dataflows'));
+        $mform->addHelpButton('config_timeout', 'connector_curl:timeout', 'tool_dataflows');
+    }
+
+    /**
+     * Validate the configuration settings.
+     *
+     * @param object $config
+     * @return true|\lang_string[] true if valid, an array of errors otherwise
+     */
+    public function validate_config($config) {
+        $errors = [];
+        if (empty($config->curl)) {
+            $errors['config_curl'] = get_string('config_field_missing', 'tool_dataflows', 'curl', true);
+        }
+        if (empty($config->rawpostdata) && ($config->method === 'put' || $config->method === 'post')) {
+            $errors['config_rawpostdata'] = get_string('config_field_missing', 'tool_dataflows', 'rawpostdata', true);
+        }
+        return empty($errors) ? true : $errors;
+    }
+
     /**
      * Executes the step
      *
-     * TODO: This will perform a curl call
+     * Performs a curl call according to given parameters.
      *
      * @return bool Returns true if successful, false otherwise.
      */
     public function execute(): bool {
-        // TODO: Implement.
+        // Get variables.
+        $config = $this->enginestep->stepdef->config;
+        $isdryrun = $this->enginestep->engine->isdryrun;
+        $method = $config->method;
+        $dbgcommand = 'curl -X ' .  strtoupper($method) . ' ' . $config->curl;
+        $result = null;
+
+        if (!empty($config->timeout)) {
+            $this->timeout = (int) $config->timeout;
+            $dbgcommand .= ' --max-time ' . $config->timeout;
+        }
+
+        $options = ['CURLOPT_TIMEOUT' => $this->timeout];
+
+        if ($method === 'post') {
+            $options['CURLOPT_POST'] = 1;
+        }
+
+        if ($method === 'put') {
+            $options['CURLOPT_PUT'] = 1;
+        }
+
+        $curl = new \curl();
+
+        $this->hassideeffect = !empty($config->sideeffects);
+
+        // For put and post methods automatically overrides/has side effects.
+        if ($config->method != 'get') {
+            $this->hassideeffect = true;
+        }
+
+        // Provided a header is specified add header to request.
+        if (!empty($config->headers)) {
+            $headers = $config->headers;
+            if (!is_array($headers)) {
+                $headers = json_decode($headers, true);
+            }
+            if (is_null($headers)) {
+                $headers = Yaml::parse($config->headers);
+            }
+            $curl->setHeader($headers);
+            $dbgcommand .= ' -H \'' . $config->headers . '\'';
+        }
+
+        // Sets post data.
+        if (!empty($config->rawpostdata)) {
+            $options['CURLOPT_POSTFIELDS'] = $config->rawpostdata;
+            $dbgcommand .= ' -d \'' . $config->rawpostdata . '\'';
+        }
+
+        // Download response to file provided destination is set.
+        if (!empty($config->destination) && !$isdryrun) {
+            if ($config->destination[0] === '/') {
+                $config->destination = ltrim($config->destination, '/');
+            }
+            $config->destination = $this->enginestep->engine->resolve_path($config->destination);
+            $file = fopen($config->destination, 'w');
+            $options['CURLOPT_FILE'] = $file;
+        }
+
+        // Perform call.
+        if (!$isdryrun) {
+            $result = $curl->$method($config->curl, [], $options);
+        }
+
+        if (!empty($file)) {
+            fclose($file);
+        }
+
+        $info = $curl->get_info();
+        // Stores response to be reusable by other steps.
+        // TODO : Once set_var api is refactored add response.
+        $response = $curl->getResponse();
+        $httpcode = $info['http_code'] ?? null;
+        $connecttime = $info['connect_time'] ?? null;
+        $totaltime = $info['total_time'] ?? null;
+        $sizeupload = $info['size_upload'] ?? null;
+        $destination = !empty($config->destination) ? $config->destination : null;
+        $errno = $curl->get_errno();
+
+        if (($httpcode >= 400 || empty($response) || $errno == 28) && !$isdryrun) {
+            throw new \moodle_exception($httpcode . ':' . $result);
+        }
+
+        if (!$isdryrun) {
+            $this->enginestep->stepdef->set_var('result', $result);
+            $this->enginestep->stepdef->set_var('httpcode', $httpcode);
+            $this->enginestep->stepdef->set_var('connecttime', $connecttime);
+            $this->enginestep->stepdef->set_var('totaltime', $totaltime);
+            $this->enginestep->stepdef->set_var('sizeupload', $sizeupload);
+            $this->enginestep->stepdef->set_var('destination', $destination);
+        } else {
+            $this->enginestep->stepdef->set_var('dbgcommand', $dbgcommand);
+        }
         return true;
     }
 }
--- a/lang/en/tool_dataflows.php
+++ b/lang/en/tool_dataflows.php
@@ -282,3 +282,15 @@ $string['connector_s3:target_help'] = 'Path to the target file. Similar to the s
 $string['connector_s3:missing_s3_source_or_target'] = 'At least one source or target path must reference a location in S3.';
 $string['connector_s3:source_is_a_directory'] = 'The source path is a directory but a file path is expected.';
 
+// Connector cURL.
+$string['connector_curl:curl'] = 'Client URL';
+$string['connector_curl:destination'] = 'File / Response destination';
+$string['connector_curl:destination_help'] = 'Should this field be left blank,
+response will be stored in step definition to be reused subsequently by other steps.';
+$string['connector_curl:headers'] = 'Headers';
+$string['connector_curl:method'] = 'HTTP request method';
+$string['connector_curl:rawpostdata'] = 'Raw post data';
+$string['connector_curl:sideeffects'] = 'Does this request have side effects?';
+$string['connector_curl:timeout'] = 'Timeout';
+$string['connector_curl:timeout_help'] = 'Time after which curl request will abort, default is 60 seconds.';
+$string['connector_curl:field_headers_help'] = 'Headers should be in the following JSON format: {$a->json} You can also use the following YAML format: {$a->yaml}';
--- a/tests/tool_dataflows_curl_connector_test.php
+++ b/tests/tool_dataflows_curl_connector_test.php
@@ -1,0 +1,208 @@
+<?php
+// This file is part of Moodle - https://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_dataflows;
+
+use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
+use Symfony\Component\Yaml\Yaml;
+use tool_dataflows\local\step\connector_curl;
+use tool_dataflows\local\execution\engine;
+use tool_dataflows\step;
+use tool_dataflows\dataflow;
+
+/**
+ * Unit test for the curl connector step.
+ *
+ * @package   tool_dataflows
+ * @author    Ghaly Marc-Alexandre <marc-alexandreghaly@catalyst-ca.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class tool_dataflows_curl_connector_test extends \advanced_testcase {
+    /**
+     * Set up before each test
+     */
+    protected function setUp(): void {
+        parent::setUp();
+        $this->resetAfterTest();
+    }
+
+    /**
+     * Tests the execute() function.
+     *
+     * @covers \tool_dataflows\local\step\connector_curl::execute
+     */
+    public function test_execute() {
+        global $CFG;
+
+        $connectorcurl = new connector_curl();
+        $testgeturl = $this->getExternalTestFileUrl('/h5puuid.json');
+
+        $stepdef = new step();
+        $dataflow = new dataflow();
+        $dataflow->name = 'connector-step';
+        $dataflow->enabled = true;
+        $dataflow->save();
+        // Tests get method.
+        $stepdef->config = Yaml::dump([
+            'curl' => $testgeturl,
+            'destination' => '',
+            'headers' => '',
+            'method' => 'get'
+        ]);
+        $stepdef->name = 'connector';
+        $stepdef->type = 'tool_dataflows\local\step\connector_curl';
+        $dataflow->add_step($stepdef);
+        ob_start();
+        $engine = new engine($dataflow, false, false);
+        $engine->execute();
+        ob_get_clean();
+        $variables = $engine->get_variables()['steps']->connector->config;
+        // Result can be anything but for readability decoded to see vars.
+        $result = json_decode($variables->result, true);;
+        $this->assertEquals($result['uuid'], '3d188fbf-d0b7-4d4e-ae4d-4b5548df824e');
+
+        $this->assertEquals($variables->httpcode, 200);
+        $this->assertObjectHasAttribute('connecttime', $variables);
+        $this->assertObjectHasAttribute('totaltime', $variables);
+        $this->assertObjectHasAttribute('sizeupload', $variables);
+
+        $testurl = $this->getExternalTestFileUrl('/test_post.php');
+
+        // Tests post method.
+        $stepdef->config = Yaml::dump([
+            'curl' => $testurl,
+            'destination' => '',
+            'headers' => '',
+            'method' => 'post',
+            'rawpostdata' => 'data=moodletest'
+        ]);
+        $dataflow->add_step($stepdef);
+        ob_start();
+        $engine = new engine($dataflow, false, false);
+        $engine->execute();
+        ob_get_clean();
+        $variables = $engine->get_variables()['steps']->connector->config;
+
+        $this->assertEmpty($variables->result);
+        $this->assertEquals($variables->httpcode, 200);
+        $this->assertObjectHasAttribute('connecttime', $variables);
+        $this->assertObjectHasAttribute('totaltime', $variables);
+        $this->assertObjectHasAttribute('sizeupload', $variables);
+
+        // Tests put method.
+        $stepdef->config = Yaml::dump([
+            'curl' => $testurl,
+            'destination' => '',
+            'headers' => '',
+            'method' => 'put',
+            'rawpostdata' => 'data=moodletest'
+        ]);
+        $dataflow->add_step($stepdef);
+        ob_start();
+        $engine = new engine($dataflow, false, false);
+        $engine->execute();
+        ob_get_clean();
+        $variables = $engine->get_variables()['steps']->connector->config;
+
+        $this->assertEmpty($variables->result);
+        $this->assertEquals($variables->httpcode, 200);
+        $this->assertObjectHasAttribute('connecttime', $variables);
+        $this->assertObjectHasAttribute('totaltime', $variables);
+        $this->assertObjectHasAttribute('sizeupload', $variables);
+
+        // Tests debug command when dry run.
+        $stepdef->config = Yaml::dump([
+            'curl' => $testurl,
+            'destination' => '',
+            'headers' => '',
+            'method' => 'post',
+            'rawpostdata' => '{
+                "name": "morpheus",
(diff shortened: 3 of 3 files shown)


classes/local/step/connector_curl.php
```diff
@@ -26,22 +26,183 @@
  * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 class connector_curl extends connector_step {
+    /** @var int[] number of input flows (min, max). */
+    protected $inputflows = [0, 0];
```
brendanheywood, line 30: It is valid to have a flow step which finishes up, and then a connector step after it. The defaults have been fixed in the connector_step base so you can remove most of these

lang/en/tool_dataflows.php
```diff
@@ -229,3 +229,13 @@
 $string['trigger_cron:crontab_desc'] = 'The schedule is edited as five values: minute, hour, day, month and day of month, in that order. The values are in crontab format.';
 $string['trigger_cron:strftime_datetime'] = '%d %b %Y, %H:%M';
 $string['trigger_cron:next_run_time'] = 'Next run time: {$a}';
+
+// Connector cURL.
+$string['connector_curl:curl'] = 'Client URL';
```
brendanheywood, line 234: need a version.php bump to pick these up

classes/local/step/connector_curl.php
```diff
@@ -26,22 +26,183 @@
...
+     * Allows each step type to determine a list of optional/required form
+     * inputs for their configuration
+     *
+     * It's recommended you prefix the additional config related fields to avoid
+     * conflicts with any existing fields.
+     *
+     * @param \MoodleQuickForm &$mform
+     */
+    public function form_add_custom_inputs(\MoodleQuickForm &$mform) {
+        $mform->addElement('textarea', 'config_curl', get_string('connector_curl:curl', 'tool_dataflows'),
```
brendanheywood, line 74: This should be a single line not a text area

classes/local/step/connector_curl.php
```diff
@@ -26,22 +26,183 @@
...
+                ['cols' => 50, 'rows' => 7]);
+        $mform->addElement('textarea', 'config_destination', get_string('connector_curl:destination', 'tool_dataflows'));
```
brendanheywood, line 76: single line not textarea

brendanheywood: also I cannot get this to work, every time I save a path eg '/tmp/postman.html'  then after I have saved it then it mangles it into destination: tmppostman.html

classes/local/step/connector_curl.php
```diff
@@ -26,22 +26,183 @@
...
+        $this->enginestep->stepdef->config->response = $response;
+
+        $httpcode = $info['http_code'];
+        $connecttime = $info['connect_time'];
+        $totaltime = $info['total_time'];
+        $sizeupload = $info['size_upload'];
+        $errno = $curl->get_errno();
+        if ($httpcode >= 400 || empty($response) || $errno == 28) {
+            if (!empty($response)) {
+                $codestatustext = $this->get_error_status($httpcode);
```
brendanheywood, line 200: Why do we need this lookup? We should use the text that we get from the response which might not be the same as this

brendanheywood:
@marcghaly can you please rebase this, was a few clashes. Comments in the pr

classes/local/step/connector_curl.php
marcghaly: Fixed.

classes/local/step/connector_curl.php
marcghaly: I used wrong param type PARAM_FILE instead of PARAM_PATH, this is fixed.

classes/local/step/connector_curl.php
marcghaly: I have remove this - used text from response.

marcghaly:
Hi @brendanheywood,

I have fixed issues above, path issue and `download_one` method issue which by default is not outputting proper response for `put` and `post`. It is rebased over latest master.

Regards,

Marc

marcghaly:
Hi @keevan,

This is a PR for issue #167, as you had that under radar is it possible for you to have a look at this ?

Thank you

Regards,

Marc

keevan:
@marcghaly Sure thing. I can take a look

classes/local/step/connector_curl.php
```diff
@@ -26,19 +26,166 @@
...
+     * @param \MoodleQuickForm &$mform
+     */
+    public function form_add_custom_inputs(\MoodleQuickForm &$mform) {
+        $mform->addElement('text', 'config_curl', get_string('connector_curl:curl', 'tool_dataflows'),
+                ['cols' => 50, 'rows' => 7]);
+        $mform->addElement('text', 'config_destination', get_string('connector_curl:destination', 'tool_dataflows'));
+        $mform->addElement('textarea', 'config_headers', get_string('connector_curl:headers', 'tool_dataflows'),
+                ['cols' => 50, 'rows' => 7]);
+        $mform->addElement('select', 'config_method', get_string('connector_curl:method', 'tool_dataflows'),
+                ['GET', 'POST', 'PUT']);
```
brendanheywood, line 74: Can I suggest this is changed so the data saved is actually the method name and not an index number. Will make it much easier to debug and future proof when we add new methods like PATCH or reorder the UI here

lang/en/tool_dataflows.php
```diff
@@ -231,3 +231,13 @@
...
+$string['connector_curl:destination'] = 'File destination';
+$string['connector_curl:headers'] = 'Headers';
+$string['connector_curl:method'] = 'HTTP request methods';
+$string['connector_curl:rawpostdata'] = 'Raw post data';
+$string['connector_curl:sideeffects'] = 'Has this request side effects ?';
```
brendanheywood, line 241: This this out a little, eg

Does this request have side effects ?

Set this to true for API calls which change state on the remove server. This only affects whether the curl will be done or not in dry run mode or not.

classes/local/step/connector_curl.php
```diff
@@ -26,19 +26,166 @@
...
+                $config->destination = substr($config->destination, 1);
+            }
+            if (isset($config->rawpostdata)) {
+                $options['CURLOPT_POSTFIELDS'] = $postdata;
+                if ($method == 'post') {
+                    $options['CURLOPT_POST'] = 1;
+                }
+                if ($method == 'put') {
+                    $options['CURLOPT_PUT'] = 1;
+                }
```
keevan, line 152: Do these options need to be nested under the postdata check? Are we able to reduce the amount of nesting here?

classes/local/step/connector_curl.php
```diff
@@ -26,19 +26,166 @@
...
+     *
+     * @param object $config
+     * @return true|\lang_string[] true if valid, an array of errors otherwise
+     */
+    public function validate_config($config) {
+        $errors = [];
+        if (empty($config->curl)) {
+            $errors['config_curl'] = get_string('config_field_missing', 'tool_dataflows', 'curl', true);
+        }
+        if (($config->method == 1 || $config->method == 2) && empty($config->rawpostdata)) {
```
keevan, line 96: Can you please create class constants or similar to hold these different methods so they can be more easily identified rather than being magic numbers?

e.g. 
```
class ... {
    const METHOD_GET = 0;
    const METHOD_POST = 1;
    const METHOD_PUT = 2;
}
```

classes/local/step/connector_curl.php
```diff
@@ -26,19 +26,166 @@
...
+        $sizeupload = $info['size_upload'] ?? null;
+        $errno = $curl->get_errno();
+        if ($httpcode >= 400 || empty($response) || $errno == 28) {
+            if (is_string($result)) {
+                $resultarr = json_decode($result, true);
+                if (isset($resultarr['error'])) {
+                    $codestatustext = $httpcode . '/' . $resultarr['error'];
+                } else {
+                    $codestatustext = $httpcode . '/' . $result;
+                }
```
keevan, line 185:
```suggestion
                $result = json_decode($result, true)['error'] ?? $result;
                $codestatustext = $httpcode . '/' . $result;
```

Would this work?

keevan (commented):
Hey @marcghaly I've done a pass of the code. I haven't tested the functionality of the code yet but I'll do that shortly.

classes/local/step/connector_curl.php
```diff
@@ -26,19 +26,166 @@
...
+                $this->hassideeffect = true;
+            } else {
+                $result = $curl->get($config->curl, [], $options);
+            }
+        }
+
+        $info = $curl->get_info();
+        // Stores response to be reusable by other steps.
+        $response = $curl->getResponse();
+        $this->enginestep->stepdef->config->response = $response;
```
keevan, line 171: Instead of setting the response here directly, could you use `$this->enginestep->set_var()` instead? (which was intended for similar purposes)

keevan:
Doing a bit more functional testing (from the UI):

1. There is no mention of where the response is stored if a file destination is not set. This information could be useful if you're creating steps in order.
2. There should be some handling for dry mode, in particular dry mode shouldn't create any files even if the file path was set, and it shouldn't perform the cURL request if it's marked as having a side effect in configuration.

brendanheywood:
> There should be some handling for dry mode, in particular dry mode shouldn't create any files even if the file path was set, and it shouldn't perform the cURL request if it's marked as having a side effect in configuration.

To refine this a little, if curl downloads a file into the local scratch dir then this should NOT be considered having a side effect. It is not changing state outside of its own temporary memory and storages. If that file was then copied in another step to path which was outside the scratch dir then that would have side effects. So in theory the only time a curl step should have side effects is if the remote host is marked as changing state

@marcghaly this is a little tricky as the scratch directory pr hasn't landed yet:

https://github.com/catalyst/moodle-tool_dataflows/pull/252/files

Also if you are in dry run mode, and normally you are curling a remote api and it grabs a file which is then used in later steps, then our workflow will now abort in dry run mode when the later step can't find the file it needs. So what we need is an additional setting which is the contents of the file it will write in dry run mode when it will not actually do the real curl call. This can just be a textarea or a file upload which ever is easiest.

keevan:
I'm having trouble referencing the response from a subsequent step for when the destination path is not set. 

e.g. using the expression `steps.curl.config.response`

To ease with testing, are you able to export the dataflow you've been using to test with?

classes/local/step/connector_curl.php
marcghaly: Fixed.

classes/local/step/connector_curl.php
```diff
@@ -26,19 +26,166 @@
...
+        if (isset($config->headers) && !empty($config->headers)) {
+            $headers = $config->headers;
+            if (!is_array($headers)) {
+                $headers = json_decode($headers, true);
+            }
+            $curl->setHeader($headers);
+        }
+        if (isset($config->rawpostdata)) {
+            $postdata = json_decode($config->rawpostdata, true);
+        }
```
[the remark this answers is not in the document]
marcghaly: Fixed.

classes/local/step/connector_curl.php
```diff
@@ -26,19 +26,166 @@
...
+        $options = ['CURLOPT_TIMEOUT' => $this->timeout];
+        $curl = new \curl();
+
+        if (!empty($config->sideeffects)) {
+            $this->hassideeffect = true;
+        } else {
+            $this->hassideeffect = false;
+        }
+
+        if (isset($config->headers) && !empty($config->headers)) {
```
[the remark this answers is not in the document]
marcghaly: Fixed.

classes/local/step/connector_curl.php
```diff
@@ -26,19 +26,166 @@
...
+            $headers = $config->headers;
+            if (!is_array($headers)) {
+                $headers = json_decode($headers, true);
+            }
+            $curl->setHeader($headers);
+        }
+        if (isset($config->rawpostdata)) {
```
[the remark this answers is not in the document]
marcghaly: If data is to be stored directly how do you handle multiple keys being posted ? for example :

{
    "name": "doe",
    "surname": "john",
    "id" : "57"
}

I have added a comment

classes/local/step/connector_curl.php
```diff
@@ -26,19 +26,166 @@
...
+     * Performs a curl call according to given parameters.
      *
      * @return bool Returns true if successful, false otherwise.
      */
     public function execute(): bool {
-        // TODO: Implement.
+        global $CFG, $PAGE;
+        // Get variables.
+        $config = $this->enginestep->stepdef->config;
+        $requesttype = [0 => 'get', 1 => 'post', 2 => 'put'];
```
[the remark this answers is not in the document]
marcghaly: I have removed this array as the select value is now the method name

classes/local/step/connector_curl.php
```diff
@@ -26,19 +26,166 @@
...
      * Executes the step
      *
-     * TODO: This will perform a curl call
+     * Performs a curl call according to given parameters.
      *
      * @return bool Returns true if successful, false otherwise.
      */
     public function execute(): bool {
-        // TODO: Implement.
+        global $CFG, $PAGE;
```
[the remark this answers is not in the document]
marcghaly: Fixed.

classes/local/step/connector_curl.php
```diff
@@ -26,19 +26,166 @@
...
+        if ($httpcode >= 400 || empty($response) || $errno == 28) {
+            if (is_string($result)) {
+                $resultarr = json_decode($result, true);
+                if (isset($resultarr['error'])) {
+                    $codestatustext = $httpcode . '/' . $resultarr['error'];
+                } else {
+                    $codestatustext = $httpcode . '/' . $result;
+                }
+            }
+            $this->enginestep->engine->abort(new \moodle_exception($codestatustext));
```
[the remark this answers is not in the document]
marcghaly: Fixed - just thrown exception

lang/en/tool_dataflows.php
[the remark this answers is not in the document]
marcghaly: Fixed

classes/local/step/connector_curl.php
```diff
@@ -26,19 +26,166 @@
...
+                $headers = json_decode($headers, true);
+            }
+            $curl->setHeader($headers);
+        }
+        if (isset($config->rawpostdata)) {
+            $postdata = json_decode($config->rawpostdata, true);
+        }
+        // Perform call or download/call.
+        if (!empty($config->destination)) {
+            // Download.
```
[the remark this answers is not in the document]
marcghaly: Fixed.

classes/local/step/connector_curl.php
```diff
@@ -26,19 +26,166 @@
...
+            }
+            if (isset($config->rawpostdata)) {
+                $options['CURLOPT_POSTFIELDS'] = $postdata;
+                if ($method == 'post') {
+                    $options['CURLOPT_POST'] = 1;
+                }
+                if ($method == 'put') {
+                    $options['CURLOPT_PUT'] = 1;
+                }
+            }
```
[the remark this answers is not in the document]
marcghaly: Fixed.

classes/local/step/connector_curl.php
marcghaly: Added them at the top as $options is used for every method. Fixed.

classes/local/step/connector_curl.php
marcghaly: Replaced those with method names as value are now direct names

classes/local/step/connector_curl.php
marcghaly: Works better ! thank you

classes/local/step/connector_curl.php
marcghaly: Fixed.

classes/local/step/connector_curl.php
```diff
@@ -26,19 +26,166 @@
...
+
+        $info = $curl->get_info();
+        // Stores response to be reusable by other steps.
+        $response = $curl->getResponse();
+        $this->enginestep->stepdef->config->response = $response;
+
+        $httpcode = $info['http_code'] ?? null;
+        $connecttime = $info['connect_time'] ?? null;
+        $totaltime = $info['total_time'] ?? null;
+        $sizeupload = $info['size_upload'] ?? null;
```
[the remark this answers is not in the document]
marcghaly: @brendanheywood mentioned in issue #167 that he wanted those stored variables - should I get rid of those or use set_vars instead ?

marcghaly:
Hi @keevan,

Thank you for the review, I made appropriate changes - are you able to use data from a former step this way `$this->enginestep->stepdef->get_variables()['steps']`, not sure if this `steps.curl.config.response` is yaml expression in which case I am not used to, do you have an example ?

Regards,

Marc

classes/local/step/connector_curl.php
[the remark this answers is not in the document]
keevan: Please set them via set_vars for now.

classes/local/step/connector_curl.php
```diff
@@ -26,19 +26,166 @@
...
+        }
+        if (isset($config->rawpostdata)) {
+            $postdata = json_decode($config->rawpostdata, true);
+        }
+        // Perform call or download/call.
+        if (!empty($config->destination)) {
+            // Download.
+            if ($config->destination[0] === '/') {
+                $config->destination = substr($config->destination, 1);
+            }
```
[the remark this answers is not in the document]
brendanheywood: Jason's patch hopefully will land today. If it does then lets use the scratch dir api, if it doesn't then lets land this PR regardless and make a follow up issue to convert curl to use the scratch dir

classes/local/step/connector_curl.php
```diff
@@ -27,18 +27,176 @@
...
+        if (!empty($config->destination) && !$isdryrun) {
+            // Download the response.
+            if ($config->destination[0] === '/') {
+                $config->destination = substr($config->destination, 1);
+            }
+            if (isset($postdata)) {
+                $options['CURLOPT_POSTFIELDS'] = $postdata;
+            }
+            $file = fopen($CFG->dirroot . '/' . $config->destination, 'w');
+            $result = $curl->$method($config->curl, $postdata, $options);
```
brendanheywood, line 165: Why do we pull the downloaded file into memory (file could be huge) and then write it to a file? we can use CURLOPT_FILE to get curl to write it directly

brendanheywood: After this is resolved I suspect this nested if's can be unravelled and we'll end up with only a single place which actually calls curl and the rest is just setting up options

classes/local/step/connector_curl.php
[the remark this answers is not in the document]
brendanheywood: @marcghaly I'm ok to land this bit as is and we can clean this up in a followup issue.

This is what I would like, but the set_var api needs to be improved to allow it:

https://github.com/catalyst/moodle-tool_dataflows/issues/260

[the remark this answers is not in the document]
brendanheywood: Sorry to clarify, I do eventually want all this timing data available, but wether it is sorted here or later I'm not too fussed. If this isn't working then comment it out and put a TODO for the later pr

classes/local/step/connector_curl.php
```diff
@@ -27,18 +27,176 @@
...
+        $httpcode = $info['http_code'] ?? null;
+        $connecttime = $info['connect_time'] ?? null;
+        $totaltime = $info['total_time'] ?? null;
+        $sizeupload = $info['size_upload'] ?? null;
+        $errno = $curl->get_errno();
+
+        if (($httpcode >= 400 || empty($response) || $errno == 28) && !$isdryrun) {
+            $codestatustext = null;
+            if (is_string($result)) {
+                $result = json_decode($result, true)['error'] ?? $result;
```
brendanheywood, line 192: This appears to be expecting that the result from the curl will be json which is not true. Curl should not care in the slightest what the download is, it could be csv or binary

classes/local/step/connector_curl.php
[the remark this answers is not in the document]
keevan: Just a heads up, this PR has been merged so I'd use the new api now.

https://github.com/catalyst/moodle-tool_dataflows/pull/252

What you want is something like:
```
$this->enginestep->engine->resolve_path($destination);
```

classes/local/step/connector_curl.php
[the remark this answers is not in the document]
keevan: +1 adding a TODO comment above the block is fine for me.

classes/local/step/connector_curl.php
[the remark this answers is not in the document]
brendanheywood: Broadly the curl step should not make any assumptions about the data it is posting, or the data it is getting back. 

If the request post data happens to be json then it should be equivalent to this curl cli command:

```
curl -X POST https://reqbin.com/echo/post/json
   -H 'Content-Type: application/json'
   -d '{"login":"my_login","password":"my_password"}'
```

if it is a form post then it could be like this:

```
curl -X POST https://reqbin.com/echo/post/form
   -H "Content-Type: application/x-www-form-urlencoded" 
   -d "param1=value1&param2=value2" 
```

When you run it in dry run mode I would like the step to output a command like the above that can be cut and paste and run on the command line manually to make this dirt easy to debug

classes/local/step/connector_curl.php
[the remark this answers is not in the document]
marcghaly: Adding a TODO

classes/local/step/connector_curl.php
marcghaly: Used CURLOPT_FILE and removed nested if's

classes/local/step/connector_curl.php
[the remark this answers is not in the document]
marcghaly: Fixed

classes/local/step/connector_curl.php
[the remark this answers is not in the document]
marcghaly: Fixed, and added command

classes/local/step/connector_curl.php
marcghaly: Fixed.

marcghaly:
Hi @keevan,

I have fixed points cited above could you have another look if possible ?

Thank you 

Marc

keevan:
Hi @marcghaly , one other thing. I would like each type of step to have at least one unit test of some sort. Are you able to include one for this step please? I'll continue the review now.

classes/local/step/connector_curl.php
keevan: Seems like these are still here. I believe you can remove both $inputflows and $inputconnectors properties from this class.

classes/local/step/connector_curl.php
```diff
@@ -27,18 +27,181 @@
...
+            $options['CURLOPT_PUT'] = 1;
+        }
+
+        $curl = new \curl();
+
+        if (!empty($config->sideeffects)) {
+            $this->hassideeffect = true;
+        } else {
+            $this->hassideeffect = false;
+        }
```
keevan, line 141:
```suggestion
        $this->hassideeffect = !empty($config->sideeffects);
```

classes/local/step/connector_curl.php
```diff
@@ -27,18 +27,181 @@
...
+            $options['CURLOPT_FILE'] = $file;
+        }
+
+        // Perform call.
+        if (!$isdryrun) {
+            $result = $curl->$method($config->curl, [], $options);
+            // For put and post methods automatically overrides/has side effects.
+            if ($config->method != 'get') {
+                $this->hassideeffect = true;
+            }
```
keevan, line 175: What is the reason this field needs to be updated?

In either case, if it's needed I think the handling for this doesn't need to be here in particular, and can join the snippet further above where it's initially set. This should be deterministic before the curl is run, so should be further above before any actual handling.

classes/local/step/connector_curl.php
```diff
@@ -27,18 +27,181 @@
...
+        }
+
+        if (!empty($config->destination) && !$isdryrun) {
+            fclose($file);
+        }
```
keevan, line 180: I think this can simply be 
```suggestion
        if (!empty($file)) {
            fclose($file);
        }
```

keevan (commented):
image

Also, can you please add some explainer/help text to the "headers" field I'm not sure what the format should be in. I assume it would be something like:
```
header-key: header value
another-key: 1234
...
etc.
```

classes/local/step/connector_curl.php
```diff
@@ -27,18 +27,181 @@
...
      * Executes the step
      *
-     * TODO: This will perform a curl call
+     * Performs a curl call according to given parameters.
      *
      * @return bool Returns true if successful, false otherwise.
      */
     public function execute(): bool {
-        // TODO: Implement.
+        global $CFG;
```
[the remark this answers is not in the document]
marcghaly: Fixed

classes/local/step/connector_curl.php
```diff
@@ -27,18 +27,181 @@
...
+        if (!empty($config->rawpostdata)) {
+            $options['CURLOPT_POSTFIELDS'] = $config->rawpostdata;
+            $dbgcommand .= ' -d \'' . $config->rawpostdata . '\'';
+        }
+
+        // Download response to file provided destination is set.
+        if (!empty($config->destination) && !$isdryrun) {
+            if ($config->destination[0] === '/') {
+                $config->destination = substr($config->destination, 1);
+            }
```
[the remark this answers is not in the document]
marcghaly: used ltrm instead

classes/local/step/connector_curl.php
```diff
@@ -27,18 +27,181 @@
...
+        $result = null;
+
+        if (!empty($config->timeout)) {
+            $this->timeout = (int) $config->timeout;
+            $dbgcommand .= ' --max-time ' . $config->timeout;
+        }
+
+        $options = ['CURLOPT_TIMEOUT' => $this->timeout];
+
+        if ($method == 'post') {
```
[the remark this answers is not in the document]
marcghaly: fixed

classes/local/step/connector_curl.php
```diff
@@ -27,18 +27,181 @@
...
+            $options['CURLOPT_POST'] = 1;
+        }
+
+        if ($method == 'put') {
```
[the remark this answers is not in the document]
marcghaly: fixed

classes/local/step/connector_curl.php
marcghaly: Fixed

classes/local/step/connector_curl.php
marcghaly: If method is `post` or `put` it will have side effects but setting it above will prevent the execution of the curl request

marcghaly:
Hi @keevan, I have added unit tests - just unsure concerning the API I am using, should it be like a "mock api" so it is always "online", this is also rebased over last master.

Regards

Marc

lang/en/tool_dataflows.php
keevan: You'll need to rebump the version again since the changes might have been lost in your rebase.

classes/local/step/connector_curl.php
keevan: Sorry I'm not following. I've tried moving the if block up and making it look like so, and it seems to work okay:
```
        $this->hassideeffect = !empty($config->sideeffects);
        if ($config->method != 'get') {
            $this->hassideeffect = true;
        }
```

The other question was whether or not $this->hassideeffect was used anywhere in this step in particular, since it's only used for display purposes and an example is located here https://github.com/catalyst/moodle-tool_dataflows/blob/MOODLE_35_STABLE/classes/local/step/flow_step.php#L109

I think if it can change, based on configuration values, then perhaps the has_side_effect function in base_step needs to be overriden in the connector_curl step, to include the extra checks. It doesn't seem like the block I mentioned above does anything in particular, but if you think it should could you please clarify?

keevan:
> Hi @keevan, I have added unit tests and fixed issues - just unsure concerning the API I am using, should it be like a "mock api" so it is always "online", this is also rebased over last master.
> 
> Regards
> 
> Marc

Hi Marc, I noticed there are various unaddressed comments on the files tab. Can you please respond to those open issues.

Also regarding the API test. It's much more reliable (and fast) if we mock the response instead of relying on a third party. There might be examples in core that do this already. From a high level, I'm expecting a combination of test things like:
- is the method set being called correctly?
- is the response being stored as a file correctly?
- is the response being stored into a variable correctly (where it can be referenced from future steps)
- shape of input/output data if any (more generally speaking)
- And anything else you'd want to check if someone were to ask, "How do you know this step works correctly?"

lang/en/tool_dataflows.php
marcghaly: Fixed.

marcghaly:
Hi @keevan, I have updated unit tests let me know status concerning side effects issue

Regards,

Marc

lang/en/tool_dataflows.php
```diff
@@ -251,3 +251,16 @@
...
+$string['connector_curl:destination'] = 'File / Response destination';
+$string['connector_curl:destination_help'] = 'Should this field be left blank,
+response will be stored in step definition to be reused subsequently by other steps.';
+$string['connector_curl:headers'] = 'Headers';
+$string['connector_curl:method'] = 'HTTP request method';
+$string['connector_curl:rawpostdata'] = 'Raw post data';
+$string['connector_curl:sideeffects'] = 'Does this request have side effects?';
+$string['connector_curl:timeout'] = 'Timeout';
+$string['connector_curl:timeout_help'] = 'Time after which curl request will abort, default is 60 seconds.';
+$string['field_headers_help'] = 'Headers should be in the following JSON format: <br> {<br>"header-key": "header value", <br>"another-key": "1234"<br>}';
```
keevan, line 266: Could you please move the code example to the caller (since code snippets shouldn't be translated), and wrap them inside a pre tag to preserve formatting.

Currently it displays as follows:

image

I'm expecting something like:

image

Also, have you considered the possibility of having YAML support for these headers? Since YAML is a superset of JSON, I think it should work with JSON, but yaml might be nicer display-wise

e.g.

image

(taken from https://www.json2yaml.com/ - ps: you can ignore line 1 of the yaml version)

keevan:
HI @marcghaly, could you please 
- update the PR description to have something like "Resolves #167" as this will ensure the linked issue is automatically closed off when this PR merges. Also it allows you to access this PR from the kanban board.
- resolve the version.php conflicts.
- Could you please add an `@author` tag to the curl connector file, as you've written the code and deserve the kudos
- there might be one extra point I've made prior to this comment, so if you could fix that up as well.

With regards the side effects setting of state. I think though it's not used currently, it might see some use later on, but it's quite harmless to leave in there. If anything, it would be moved to a `has_side_effects` function (overriding the base class), which will then affect how it displays in the UI, but that could be in a separate issue and addresssed when needed.

But for now, I'm quite happy with what it's currently demonstrating so if there are any bugs / issues once used with the json reader step then we'll iron those out then as well.

Thanks @marcghaly.

tests/tool_dataflows_curl_connector_test.php
```diff
@@ -0,0 +1,210 @@
...
+namespace tool_dataflows;
+
+use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
+use Symfony\Component\Yaml\Yaml;
+use tool_dataflows\local\step\connector_curl;
+use tool_dataflows\local\execution\engine;
+use tool_dataflows\step;
+use tool_dataflows\dataflow;
+
+defined('MOODLE_INTERNAL') || die();
```
keevan, line 26: This is not required, so if you could please remove this line.

```

FILE: /admin/tool/dataflows/tests/tool_dataflows_curl_connector_test.php
--------------------------------------------------------------------------------------------------------------------
--------------------------------------------------------------------------------------------------------------------
 26 | WARNING | [x] Unexpected MOODLE_INTERNAL check. No side effects or multiple artifacts detected.
--------------------------------------------------------------------------------------------------------------------
--------------------------------------------------------------------------------------------------------------------
```

lang/en/tool_dataflows.php
marcghaly: I have fixed display and move string to caller, also added yaml support and its corresponding example

tests/tool_dataflows_curl_connector_test.php
marcghaly: Fixed

marcghaly:
Hi @keevan, points above have been fixed - branch is rebased over latest main one.

Regards,

Marc

keevan (approved):
Looks good now, there'll be more adjustments made as we go along since there are other things available but this is good to be merged in.

keevan:
Test comment:

Resolves #167

PR #246, opened 2022-06-29
Issue #207: JSON reader step

### LINKED ISSUE #207: Create a JSON reader step
* [ ] This is more or less like the csv reader except it will start with json
* [ ] The step will start by parsing the entire json document into memory, so not great for super massive api calls
* [ ] an optional expression which will be a sort by, when present it will sort the array by this expression
* [ ] it will need an expression to reach into the json to pull out the bit which is an iterator. In the example below we want to pull out the "users" array and turn that into a flow

```
{
"users": [{
  ...,
}],
  "modified":  1654058940,
  "errors": []
}
```
* [ ] #265

petersistrom:
Resolves #207

### DIFF
--- a/classes/local/step/reader_json.php
+++ b/classes/local/step/reader_json.php
@@ -16,7 +16,9 @@
 
 namespace tool_dataflows\local\step;
 
+use html_writer;
 use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
+use tool_dataflows\helper;
 use tool_dataflows\local\execution\iterators\iterator;
 use tool_dataflows\local\execution\iterators\dataflow_iterator;
 
@@ -47,7 +49,6 @@ class reader_json extends reader_step {
      * Get the iterator for the step, based on configurations.
      *
      * @return iterator
-     * @throws \moodle_exception
      */
     public function get_iterator(): iterator {
         $jsonarray = $this->parse_json();
@@ -97,6 +98,7 @@ class reader_json extends reader_step {
      * Parses stream to json string.
      *
      * @return string
+     * @throws \moodle_exception
      */
     protected function get_json_string(string $path): string {
         $jsonstring = file_get_contents($this->enginestep->engine->resolve_path($path));
@@ -138,8 +140,13 @@ class reader_json extends reader_step {
      */
     public function validate_config($config) {
         $errors = [];
-        if (empty($config->pathtojson)) {
+        if (!isset($config->pathtojson)) {
             $errors['config_pathtojson'] = get_string('config_field_missing', 'tool_dataflows', 'pathtojson', true);
+        } else {
+            $error = helper::path_validate($config->pathtojson);
+            if ($error !== true) {
+                $errors['config_pathtojson'] = $error;
+            }
         }
         return empty($errors) ? true : $errors;
     }
@@ -157,7 +164,8 @@ class reader_json extends reader_step {
         // JSON array source.
         $mform->addElement('text', 'config_pathtojson', get_string('reader_json:pathtojson', 'tool_dataflows'));
         $mform->addElement('static', 'config_json_path_help', '',
-            get_string('reader_json:json_path_help', 'tool_dataflows', '<code>file:///path/to/file.json</code>'));
+            get_string('reader_json:json_path_help', 'tool_dataflows',
+                html_writer::nonempty_tag('code', 'file:///path/to/file.json')));
 
         // Array iterator value.
         $arrayexample = (object) [
@@ -171,16 +179,18 @@ class reader_json extends reader_step {
             'modified' => [1654058940],
             'errors' => [],
         ];
-        $jsonexample = '<br/><code>'.json_encode($arrayexample, JSON_PRETTY_PRINT).'</code>';
-        $expression = '<code>data.list.users</code>';
+        $jsonexample = html_writer::empty_tag('br').html_writer::nonempty_tag('code', json_encode($arrayexample, JSON_PRETTY_PRINT));
+        $expression = html_writer::nonempty_tag('code', 'data.list.users');
 
         $mform->addElement('text', 'config_arrayexpression', get_string('reader_json:arrayexpression', 'tool_dataflows'));
         $mform->addElement('static', 'config_arrayexpression_help', '',
-            get_string('reader_json:arrayexpression_help', 'tool_dataflows', ['jsonexample' => $jsonexample, 'expression' => $expression]));
+            get_string('reader_json:arrayexpression_help', 'tool_dataflows',
+                ['jsonexample' => $jsonexample, 'expression' => $expression]));
 
         // JSON array sort by.
         $mform->addElement('text', 'config_arraysortexpression', get_string('reader_json:arraysortexpression', 'tool_dataflows'));
         $mform->addElement('static', 'config_arraysortexpression_help', '',
-            get_string('reader_json:arraysortexpression_help', 'tool_dataflows', '<code>userdetails.firstname</code>'));
+            get_string('reader_json:arraysortexpression_help', 'tool_dataflows',
+                html_writer::nonempty_tag('code', 'usersdetails.firstname')));
     }
 }
--- a/classes/local/step/reader_json.php
+++ b/classes/local/step/reader_json.php
@@ -19,6 +19,7 @@ namespace tool_dataflows\local\step;
 use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
 use tool_dataflows\local\execution\iterators\iterator;
 use tool_dataflows\local\execution\iterators\dataflow_iterator;
+
 /**
  * JSON reader step
  *
@@ -36,7 +37,7 @@ class reader_json extends reader_step {
      */
     protected static function form_define_fields(): array {
         return [
-            'json' => ['type' => PARAM_TEXT],
+            'pathtojson' => ['type' => PARAM_TEXT],
             'arrayexpression' => ['type' => PARAM_TEXT],
             'arraysortexpression' => ['type' => PARAM_TEXT],
         ];
@@ -56,22 +57,22 @@ class reader_json extends reader_step {
     /**
      * Parses json string to php array.
      *
-     * @return array
+     * @return mixed
      * @throws \moodle_exception
      */
-    protected function parse_json(): array {
+    protected function parse_json() {
         $config = $this->enginestep->stepdef->config;
-        $jsonstring = $this->get_json_string($config);
+        $jsonstring = $this->get_json_string($config->pathtojson);
 
         $decodedjson = json_decode($jsonstring);
         if (is_null($decodedjson)) {
-            throw new \moodle_exception(get_string('reader_json:failed_to_decode_json', 'tool_dataflows', $config->json));
+            throw new \moodle_exception(get_string('reader_json:failed_to_decode_json', 'tool_dataflows', $config->pathtojson));
         }
 
         $arrayexpression = $config->arrayexpression;
         $expressionlanguage = new ExpressionLanguage();
         $returnarray = $expressionlanguage->evaluate(
-            'data.'.$arrayexpression,
+            $arrayexpression != '' ? 'data.'.$arrayexpression : 'data',
             [
                 'data' => $decodedjson,
             ]
@@ -97,12 +98,11 @@ class reader_json extends reader_step {
      *
      * @return string
      */
-    protected function get_json_string($config): string {
-        $jsonstring = file_get_contents($config->json);
-
+    protected function get_json_string(string $path): string {
+        $jsonstring = file_get_contents($this->enginestep->engine->resolve_path($path));
         if ($jsonstring === false) {
             $this->enginestep->log(error_get_last()['message']);
-            throw new \moodle_exception(get_string('reader_json:failed_to_open_file', 'tool_dataflows', $config->json));
+            throw new \moodle_exception(get_string('reader_json:failed_to_open_file', 'tool_dataflows', $path));
         }
 
         return $jsonstring;
@@ -116,17 +116,17 @@ class reader_json extends reader_step {
      */
     public static function sort_by_config_value(array $array, string $sortbyexpression): array {
         $expressionlanguage = new ExpressionLanguage();
-            usort($array, function($a, $b) use ($sortbyexpression, $expressionlanguage) {
-                $a = $expressionlanguage->evaluate(
-                    'data.'.$sortbyexpression,
-                    ["data" => $a]
-                );
-                $b = $expressionlanguage->evaluate(
-                    'data.'.$sortbyexpression,
-                    ["data" => $b]
-                );
-                return strnatcasecmp($a, $b);
-            });
+        usort($array, function($a, $b) use ($sortbyexpression, $expressionlanguage) {
+            $a = $expressionlanguage->evaluate(
+                'data.'.$sortbyexpression,
+                ['data' => $a]
+            );
+            $b = $expressionlanguage->evaluate(
+                'data.'.$sortbyexpression,
+                ['data' => $b]
+            );
+            return strnatcasecmp($a, $b);
+        });
         return $array;
     }
 
@@ -138,8 +138,8 @@ class reader_json extends reader_step {
      */
     public function validate_config($config) {
         $errors = [];
-        if (empty($config->json)) {
-            $errors['config_json'] = get_string('config_field_missing', 'tool_dataflows', 'json', true);
+        if (empty($config->pathtojson)) {
+            $errors['config_pathtojson'] = get_string('config_field_missing', 'tool_dataflows', 'pathtojson', true);
         }
         return empty($errors) ? true : $errors;
     }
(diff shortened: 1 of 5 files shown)


classes/local/step/reader_json.php
```diff
@@ -0,0 +1,112 @@
...
+     * inputs for their configuration
+     *
+     * It's recommended you prefix the additional config related fields to avoid
+     * conflicts with any existing fields.
+     *
+     * @param \MoodleQuickForm &$mform
+     */
+    public function form_add_custom_inputs(\MoodleQuickForm &$mform) {
+        // JSON array source.
+        $mform->addElement('textarea', 'config_json', get_string('reader_json:json', 'tool_dataflows'));
```
brendanheywood, line 101: This looks like inline json in the config? This should be a path to where a json file can be found

when parsing it throw a couple different error if it is not found vs not readable vs not parable

lang/en/tool_dataflows.php
```diff
@@ -77,6 +77,7 @@
 $string['step_name_flow_logic_join'] = 'Flow join';
 $string['step_name_flow_transformer_filter'] = 'Filter transformer';
 $string['step_name_reader_sql'] = 'SQL reader';
+$string['step_name_reader_json'] = 'JSON reader';
```
brendanheywood, line 80: Try to keep lang strings order alphabetically by their key, so this should be up one line

classes/local/step/reader_json.php
keevan: Just noting here for consideration, that the path could potentially be something referenced by an expression set here: 
e.g. `${{ curl_step.config.output.path }}`

lang/en/tool_dataflows.php
petersistrom: Updated

classes/local/step/reader_json.php
petersistrom: Updated to use JSON file path and added error handling.

classes/local/step/reader_json.php
```diff
@@ -0,0 +1,138 @@
...
+     * conflicts with any existing fields.
+     *
+     * @param \MoodleQuickForm &$mform
+     */
+    public function form_add_custom_inputs(\MoodleQuickForm &$mform) {
+        // JSON array source.
+        $mform->addElement('text', 'config_json', get_string('reader_json:json', 'tool_dataflows'));
+        $mform->addElement('static', 'config_json_help', '', get_string('reader_json:json_help', 'tool_dataflows'));
+
+        // Array iterator value.
```
brendanheywood, line 130: Both these to settings need to be converted to accept expressions but this an be in a follow up pr if preferred

lang/en/tool_dataflows.php
```diff
@@ -219,6 +225,14 @@
 $string['reader_sql:countervalue'] = 'Counter Value';
 $string['reader_sql:countervalue_help'] = 'Current value from the counter field';
 
+// Reader JSON.
+$string['reader_json:arraykey'] = 'Array Key';
+$string['reader_json:arraykey_help'] = 'Nested array to extract from JSON. For example, <code>user.tags</code> will parse the tags array. If empty the entire array will be parsed.';
```
brendanheywood, line 230: Make sure the help text for both of these has a tangible example with a slightly nested data structure and the expression needed to grab the data out.

lang/en/tool_dataflows.php
```diff
@@ -219,6 +225,14 @@
...
+$string['reader_json:arraysort'] = 'Sort by';
+$string['reader_json:arraysort_help'] = 'Value to sort array by';
```
brendanheywood, line 232: Make sure this is an expression and the expression should be relative to each item in the array specified by the first expression and the docs have an example using the same data structure examples in the json setting

lang/en/tool_dataflows.php
```diff
@@ -219,6 +226,14 @@
 $string['reader_sql:countervalue'] = 'Counter Value';
 $string['reader_sql:countervalue_help'] = 'Current value from the counter field';
 
+// Reader JSON.
+$string['reader_json:arraykey'] = 'Array Key';
+$string['reader_json:arraykey_help'] = 'Nested array to extract from JSON. For example, <code>data.list.users</code> will return the users array from the following JSON: <code>{"data": {"list": {"users": [{ "id": "1", "username": "User1", "name": "Name1"},{ "id": "2","username": "User2","name": "Name2"} ] } } ,"modified": 1654058940, "errors": []}</code>. If empty it is assumed the starting point of the JSON file is an array.';
```
brendanheywood, line 231: Can you move the json example out of the string and into code as this will not be translated, and can you pretty print format it in code and pass it into the string

classes/local/step/reader_json.php
```diff
@@ -0,0 +1,183 @@
...
+    protected function parse_json(): array {
+        $config = $this->enginestep->stepdef->config;
+        $jsonstring = $this->get_json_string($config);
+
+        $decodedjson = json_decode($jsonstring);
+        if (is_null($decodedjson)) {
+            throw new \moodle_exception(get_string('reader_json:failed_to_decode_json', 'tool_dataflows', $config->json));
+        }
+
+        $arraykey = $config->arraykey;
```
brendanheywood, line 77: Since this is no longer an arraykey can you rename this and related vars to something better like maybe 'arrayexpression'

petersistrom: Updated var names and lang strings

classes/local/step/reader_json.php
```diff
@@ -0,0 +1,184 @@
+<?php
```
keevan, line 1:
```
/admin/tool/dataflows/classes/local/step/reader_json.php
----------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 3 WARNINGS AFFECTING 3 LINES
----------------------------------------------------------------------------------------------------------
  87 | WARNING | Line exceeds 132 characters; contains 133 characters
 178 | WARNING | Line exceeds 132 characters; contains 151 characters
 182 | WARNING | Line exceeds 132 characters; contains 146 characters
----------------------------------------------------------------------------------------------------------
```

classes/local/step/reader_json.php
```diff
@@ -0,0 +1,184 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_dataflows\local\step;
+
+use core_admin\local\settings\filesize;
```
keevan, line 19: This was declared but is not used anywhere, so it seems you can safely remove this.

classes/local/step/reader_json.php
```diff
@@ -0,0 +1,184 @@
...
+use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
+use tool_dataflows\local\execution\engine_step;
+use tool_dataflows\local\execution\flow_engine_step;
```
keevan, line 22: This was declared but is not used anywhere, so it seems you can safely remove this.

classes/local/step/reader_json.php
```diff
@@ -0,0 +1,184 @@
...
+use tool_dataflows\local\execution\iterators\iterator;
+use tool_dataflows\local\execution\iterators\dataflow_iterator;
+use tool_dataflows\local\formats\encoders\json;
```
keevan, line 25: This was declared but is not used anywhere, so it seems you can safely remove this.

classes/local/step/reader_json.php
```diff
@@ -0,0 +1,184 @@
...
+use tool_dataflows\parser;
```
keevan, line 26: This was declared but is not used anywhere, so it seems you can safely remove this.

classes/local/step/reader_json.php
```diff
@@ -0,0 +1,184 @@
...
+     */
+    public function get_iterator(): iterator {
+        $jsonarray = $this->parse_json();
+        return new dataflow_iterator($this->enginestep, new \ArrayIterator($jsonarray));
+    }
+
+    /**
+     * Parses json string to php array.
+     *
+     * @return string
```
keevan, line 65: The return type seems to be wrong here.

classes/local/step/reader_json.php
```diff
@@ -0,0 +1,184 @@
...
+
+        return $jsonstring;
+    }
+
+    /**
+     * Sort array by config value.
+     *
+     * @param array $array
+     * @param string $sortbyexpression
+     * @param engine_step $enginestep
```
keevan, line 114: Extra unused `@param` here.

classes/local/step/reader_json.php
```diff
@@ -0,0 +1,184 @@
...
+                'list' => ['users' => [
+                            [ "id" => "1",  "userdetails" => ["firstname" => "Bob", "lastname" => "Smith", "name" => "Name1"]],
+                            [ "id" => "2",  "userdetails" => ["firstname" => "John", "lastname" => "Doe", "name" => "Name2"]],
+                            [ "id" => "3",  "userdetails" => ["firstname" => "Foo", "lastname" => "Bar", "name" => "Name3"]]
+                        ],
+                    ]
+                ],
+                'modified' => [1654058940],
+                'errors' => [],
+            ];
```
keevan, line 175: I think the indentation can be improved. I think having the array printed as the following will look fine:
```
        $arrayexample = (object) [
            'data' => (object) [
                'list' => [
                    'users' => [
                        [ "id" => "1", "userdetails" => ["firstname" => "Bob", "lastname" => "Smith", "name" => "Name1"]],
                        [ "id" => "2", "userdetails" => ["firstname" => "John", "lastname" => "Doe", "name" => "Name2"]],
                        [ "id" => "3", "userdetails" => ["firstname" => "Foo", "lastname" => "Bar", "name" => "Name3"]]
                    ],
                ]
            ],
            'modified' => [1654058940],
            'errors' => [],
        ];

```

keevan (commented):
Thanks for making the changes. Can you please add a supporting phpunit test and add a workflow that demonstrates this works from the UI as well?

classes/local/step/reader_json.php
keevan: We should aim to leave the code in a better state than what it was prior to our changes.

petersistrom:
Test dataflow and JSON array
[jsonarray.txt](https://github.com/catalyst/moodle-tool_dataflows/files/9051585/jsonarray.txt)
[JSON_Reader_test_20220706_0428.txt](https://github.com/catalyst/moodle-tool_dataflows/files/9051588/JSON_Reader_test_20220706_0428.txt)

classes/local/step/reader_json.php
```diff
@@ -0,0 +1,185 @@
...
+class reader_json extends reader_step {
+
+    /**
+     * Return the definition of the fields available in this form.
+     *
+     * @return array
+     */
+    protected static function form_define_fields(): array {
+        return [
+            'json' => ['type' => PARAM_TEXT],
```
keevan, line 39: I wonder instead of reading this, questioning what the JSON key is used for / as, and looking up the lang string to reach some sort of clarity.. could you rename this field to `pathtojson`, `input`, `filepath` or similar to make things easier overall, as json is a bit vague.

classes/local/step/reader_json.php
```diff
@@ -0,0 +1,185 @@
...
+    protected function get_json_string($config): string {
+        $jsonstring = file_get_contents($config->json);
+
+        if ($jsonstring === false) {
+            $this->enginestep->log(error_get_last()['message']);
+            throw new \moodle_exception(get_string('reader_json:failed_to_open_file', 'tool_dataflows', $config->json));
+        }
+
+        return $jsonstring;
+    }
```
keevan, line 109: Instead of passing in the config,  could we pass in the value of the field used as is? I believe the type would be string
e.g. `function get_json_string(string $path): string {`

keevan (commented):
Just did another pass of the code. Looking at the tests and testing it in the UI now.

tests/tool_dataflows_json_reader_test.php
```diff
@@ -0,0 +1,165 @@
...
+
+    /**
+     * Test validate_config().
+     *
+     * @covers \tool_dataflows\local\step\reader_json::validate_config
+     * @throws \coding_exception
+     */
+    public function test_validate_config() {
+        // Test valid configuration.
+        $config = (object)['json' => '{key:{value:Array[]}}'];
```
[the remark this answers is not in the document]
keevan: Also it seems the json key (currently) is only used as a file path, so having the value of it be anything other than that might seem strange as a test value. 

Can you either update to update the validation to ensure it is only sourced from a valid location? If that isn't as quick to add then perhaps it could just hold a more sane value for the mean time.

tests/tool_dataflows_json_reader_test.php
```diff
@@ -0,0 +1,165 @@
...
+            'modified' => [1654058940],
+            'errors' => [],
+        ]);
+
+        // Test unsorted array.
+        $expecteduserarray = json_decode('[
+            { "id": "2",  "userdetails": {"firstname": "John", "lastname": "Doe", "name": "Name2"}},
+            { "id": "1",  "userdetails": {"firstname": "Bob", "lastname": "Smith", "name": "Name1"}},
+            { "id": "3",  "userdetails": {"firstname": "Foo", "lastname": "Bar", "name": "Name3"}}
+        ]');
```
keevan, line 102: I think you can simplify the setup a bit. Create a $users array in php. Link it to the `data.list` key in $json, and set it as the $expecteduserarray. Since json_decode would result in the same array.

tests/tool_dataflows_json_reader_test.php
```diff
@@ -0,0 +1,165 @@
...
+        ob_start();
+        $engine = new engine($dataflow);
+        $engine->execute();
+        ob_get_clean();
+
+        $resultdata = json_decode(file_get_contents($tmpoutput2));
+        $this->assertEquals($expecteduserarray, $resultdata);
+
+        $this->assertEquals(engine::STATUS_FINALISED, $engine->status);
+    }
```
keevan, line 164: Could you add another test to see if can loop over an array of object keys? I'm curious if it is supported out of the box, and whether or not sorting will work out of the box as well.

Example data could be something like:
```json
{
  "firstname": "Bob",
  "lastname": "Smith",
  "name": "Name1"
}
```

I'm expecting the iterator to return a value like `Bob`, `Smith`, `Name` .. but it might not be that simple.

classes/local/step/reader_json.php
```diff
@@ -0,0 +1,185 @@
...
+                            [ "id" => "2",  "userdetails" => ["firstname" => "John", "lastname" => "Doe", "name" => "Name2"]],
+                            [ "id" => "3",  "userdetails" => ["firstname" => "Foo", "lastname" => "Bar", "name" => "Name3"]]
+                        ],
+                    ]
+                ],
+                'modified' => [1654058940],
+                'errors' => [],
+            ];
+        $jsonexample = json_encode($arrayexample, JSON_PRETTY_PRINT);
+        $mform->addElement('text', 'config_arrayexpression', get_string('reader_json:arrayexpression', 'tool_dataflows'));
```
keevan, line 176: Functional testing seems mostly okay. I've added my concerns in prior comments.

image

Could you provide a smaller example, and have the example be pretty printed so that it's not all on one line. This is a bit hard to read for a field talking about nesting and structure. I think you might be missing a `<code>` tag to wrap the example in.

keevan (commented):
Aside from these final comments, it's working great and I'm happy to get this merged after you've fixed up those small bits.

classes/local/step/reader_json.php
```diff
@@ -0,0 +1,196 @@
...
+                'list' => [
+                    'users' => [
+                        [ "id" => "1",  "userdetails" => ["firstname" => "Bob", "lastname" => "Smith", "name" => "Name1"]],
+                    ],
+                ]
+            ],
+            'modified' => [1654058940],
+            'errors' => [],
+        ];
+        $jsonexample = html_writer::empty_tag('br').html_writer::nonempty_tag('code', json_encode($arrayexample, JSON_PRETTY_PRINT));
```
keevan, line 182: Could you please split the $jsonexample setting into multiple lines, e.g. one for `br` one for `code`.

```

FILE: /admin/tool/dataflows/classes/local/step/reader_json.php
----------------------------------------------------------------------------------------------------------
----------------------------------------------------------------------------------------------------------
 182 | WARNING | Line exceeds 132 characters; contains 133 characters
----------------------------------------------------------------------------------------------------------
```

classes/local/step/reader_json.php
```diff
@@ -179,7 +179,8 @@ public function form_add_custom_inputs(\MoodleQuickForm &$mform) {
             'modified' => [1654058940],
             'errors' => [],
         ];
-        $jsonexample = html_writer::empty_tag('br').html_writer::nonempty_tag('code', json_encode($arrayexample, JSON_PRETTY_PRINT));
+        $jsonexample = html_writer::empty_tag('br').
+            html_writer::nonempty_tag('pre', json_encode($arrayexample, JSON_PRETTY_PRINT));
```
keevan, line 183: I would have personally preferred something like:
```suggestion
        $jsonexample = html_writer::empty_tag('br');
        $jsonexample .= html_writer::nonempty_tag('pre', json_encode($arrayexample, JSON_PRETTY_PRINT));
```
But this is fine for now.

keevan (approved):

PR #249, opened 2022-06-30
fix: add validation for step name and alias fields

### LINKED ISSUE #247: Bug when creating a step with the same name as an existing step
This seems to silently fail and it looks like it worked but you don't get a step

It should be a validation error and say 'bzzt duplicate step'

keevan:
Resolves #247


brendanheywood:
lgtm, auto merged

Issue #199, opened 2022-06-10
Have help text which shows all the available variables usable in expressions


brendanheywood:
This list might end up large but we'll worry about that later. Probably needs to be a method which takes a step as context and then it just makes a list of all the global, flow, and step vars which can be referenced.

Some vars we won't know in advance but we want to, so we probably also need a method that each step can override which provides metadata around the vars that it would populate and a lang string describing what they are. Base step impl would add the keys for the state changes, and then the flow class might extend that with some more etc

Show this as help text somehow next to any input which can accept expressions. Might have a helper custom field input type for text and for textareas which manages it

PR #252, opened 2022-06-30
Issue #158: Add a scratch directory for dataflows.

### LINKED ISSUE #158: Have a temporary scratch directory / sandpit for every instance of a running workflow
There is a couple drivers for this:
1) make it much easier to configure steps using a relative path name so we don't have to worry about stream prefixes like file:// or path prefixes like /var/www/sitedata/
2) from a security perspective I think it is good to have most steps potentially locked down so that can't read or write from a completely arbitrary internal path. Need to carefully walk the balance here between security and usefulness

So proposing:
* [ ] every instance starts by getting an instance id
* [ ] for most runs the instance will run from end to end on a single web front end or cron node so we can internally make the directory using make_request_directory
* [ ] later if we implement instance hibernation or multi processes workflows then we will need to move this to a shared directory
* [ ] most steps that do things like copy a file, or rename a file or something with files will be able to accept either a relative path, or an absolute path, or a stream url, depending on what type of step it is
* [ ] we can probably make or repurposed a form element for a 'path' which can only handle paths, and a helper which translates the relative paths in the step config to the real path inside our scratch directory
* [ ] steps will then just be configured with 'input.csv' instead of something much bigger
* [ ] we could have global admin settings, which can be locked, which is a whitelist of paths that the steps can read and write to outside its own scratch dir

jaypha:
Resolves #158

### DIFF
--- a/classes/helper.php
+++ b/classes/helper.php
@@ -1,0 +1,110 @@
+<?php
+// This file is part of Moodle - https://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_dataflows;
+
+/**
+ * Support functions for dataflows.
+ *
+ * @package   tool_dataflows
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class helper {
+
+    /**
+     * Get the scheme for a path string. will default to 'file' if none present.
+     *
+     * @param string $path
+     * @return string
+     */
+    public static function path_get_scheme(string $path): string {
+        $splitpath = explode('://', $path, 2);
+        if (count($splitpath) !== 2) {
+            return 'file';
+        }
+        return $splitpath[0];
+    }
+
+    /**
+     * Does the path string include a scheme?
+     *
+     * @param string $path
+     * @return bool
+     */
+    public static function path_has_scheme(string $path): bool {
+        $splitpath = explode('://', $path, 2);
+        return count($splitpath) === 2;
+    }
+
+    /**
+     * Is the path a relative one?
+     * Note: Windows paths are not yet supported.
+     *
+     * @param string $path
+     * @return bool
+     */
+    public static function path_is_relative(string $path): bool {
+        if (self::path_has_scheme($path)) {
+            return false;
+        }
+
+        // Unix absolute path.
+        return substr($path, 0, 1) !== '/';
+
+        // TODO: Windows support.
+    }
+
+    /**
+     * Makes a full path from a relative one using the given base dir.
+     *
+     * @param  string $path
+     * @param  string $scratchdir
+     * @return string
+     */
+    public static function path_get_absolute(string $path, string $scratchdir): string {
+        if (self::path_is_relative($path)) {
+            return $scratchdir . '/' . $path;
+        }
+        return $path;
+    }
+
+    /**
+     * Validate a path.
+     *
+     * @param  string $path
+     * @return \lang_string|true True if the path is valid. A lang_string otherwise.
+     */
+    public static function path_validate(string $path) {
+        if (self::path_is_relative($path)) {
+            return true;
+        }
+
+        if (self::path_get_scheme($path) !== 'file') {
+            return true;
+        }
+
+        $permitteddirs = explode(PHP_EOL, trim(get_config('tool_dataflows', 'permitted_dirs')));
+        foreach ($permitteddirs as $dir) {
+            if (substr($path, 0, strlen($dir)) === $dir) {
+                return true;
+            }
+        }
+
+        return get_string('path_invalid', 'tool_dataflows', $path, true);
+    }
+}
--- a/classes/local/execution/engine.php
+++ b/classes/local/execution/engine.php
@@ -19,6 +19,7 @@ namespace tool_dataflows\local\execution;
 use Symfony\Component\Yaml\Yaml;
 use tool_dataflows\dataflow;
 use tool_dataflows\exportable;
+use tool_dataflows\helper;
 use tool_dataflows\local\step\flow_cap;
 use tool_dataflows\run;
 
@@ -106,6 +107,9 @@ class engine {
     /** @var bool True if executing via automation. */
     protected $automated = true;
 
+    /** @var string Scratch directory for temporary files. */
+    protected $scratchdir = null;
+
     /**
      * Constructs the engine.
      *
@@ -172,6 +176,8 @@ class engine {
             // Add sinks to the execution queue.
             $this->queue = $this->sinks;
             $this->set_status(self::STATUS_INITIALISED);
+
+            $this->scratchdir = make_request_directory();
         } catch (\Throwable $thrown) {
             $this->abort($thrown);
         }
@@ -339,6 +345,7 @@ class engine {
             case 'isdryrun':
             case 'run':
             case 'status':
+            case 'scratchdir':
                 return $this->$parameter;
             case 'name':
                 return $this->dataflow->name;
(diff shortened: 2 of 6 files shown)


classes/lib.php
```diff
@@ -0,0 +1,115 @@
+<?php
```
keevan, line 1: I think this file (since everything here is static) probably belongs in locallib.php OR under the classes/helper.php (also classname is helper.php). This observation is based on examples I can see in Moodle core, e.g. badges/classes/helper.php, mod/lti/classes/helper.php.

classes/lib.php
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /**
+     * Makes a full path from a relative one using th given base dir.
+     *
+     * @param string $path
+     * @param string $scratchdir
+     * @return string
+     */
+    public static function path_make_absolute(string $path, string $scratchdir): string {
```
keevan, line 86: Minor name change suggestion(s):
- get_absolute_path (for recall-ability)
- path_get_absolute (for consistency)

classes/lib.php
```diff
@@ -0,0 +1,115 @@
...
+        if (self::path_is_relative($path)) {
+            return $scratchdir . '/' . $path;
+        } else {
+            return $path;
+        }
```
keevan, line 91: Suggestion is to remove this `else` nesting as it is redundant
```suggestion

        return $path;
```

classes/lib.php
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /**
+     * Validate a path string.
```
keevan, line 95: PHPDocs summary should generally be followed by a blank line, before the other phpdoc tags

classes/lib.php
```diff
@@ -0,0 +1,115 @@
...
+     * @param string $path
+     */
+    public static function path_validate(string $path) {
+        global $CFG;
+
+        if (self::path_is_relative($path)) {
+            return true;
+        }
```
keevan, line 103: Please space out each if individual block for legibility.

classes/lib.php
```diff
@@ -0,0 +1,115 @@
...
+        if (self::path_get_scheme($path) !== 'file') {
+            return true;
+        }
```
keevan, line 106: Please space out each if individual block for legibility.

classes/local/execution/engine.php
```diff
@@ -172,6 +175,8 @@ public function initialise() {
             // Add sinks to the execution queue.
             $this->queue = $this->sinks;
             $this->set_status(self::STATUS_INITIALISED);
+
+            $this->scratchdir = make_request_directory();
```
keevan, line 179: I believe the instance id is available now `tool_dataflows_runs.id`, so perhaps it's a good idea to start adding support for a shared directory (run folder?) based on this instance id of the dataflow. I think initially, it would just use that by default, and any path will be relative under it. e.g. the dataflow cannot reference a file from a different run folder as the supporting checks would be assisting with.

Covers points 1 + 3 here: https://github.com/catalyst/moodle-tool_dataflows/issues/158

keevan (commented):
I've done an initial code review. Please let me know if you have any questions. Thanks Jason.

classes/lib.php
keevan: I'd probably resolve this after resolving the comments under the file as it might just, _dissapear_ since the file no longer exists (maybe GitHub is better now)

classes/lib.php
```diff
@@ -0,0 +1,115 @@
...
+
+        // Unix absolute path.
+        $first = substr($path, 0, 1);
+        if ($first == '/' || $first == '\\') {
+            return false;
+        }
+
+        // Windows absolute path.
+        if (preg_match('/^[a-zA-Z]:\\\/', $path) === 1) {
```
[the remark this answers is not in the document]
brendanheywood: You can lean on pathinfo()['dirname'] == ''  to do all the cross OS logic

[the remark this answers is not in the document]
jaypha: https://www.php.net/manual/en/wrappers.file.php

classes/lib.php
jaypha: locallib.php is deprecated, so I'm changing it to helper.php.

classes/local/execution/engine.php
brendanheywood: I'm on the fence around this. make_request_directory() is what we want from a performance and scaling point of view and I think is good for now.

make_request_directory isn't good if we want to pause a workflow and resume it which might end up on another node. If and when we do that then I think when we hibernate and thaw a workflow we can move it from the request dir to datadir as needed

classes/helper.php
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    /**
+     * Validate a path string.
+     *
+     * @param  string $path
+     * @return \lang_string|true True if the path is valid. A lang_string otherwise.
+     */
+    public static function path_validate(string $path) {
+
```
keevan, line 93: Please remove this empty line

classes/helper.php
```diff
@@ -0,0 +1,111 @@
...
+     */
+    public static function path_get_absolute(string $path, string $scratchdir): string {
+        if (self::path_is_relative($path)) {
+            return $scratchdir . '/' . $path;
+        }
+        return $path;
+    }
+
+    /**
+     * Validate a path string.
```
keevan, line 87:
```suggestion
     * Validate a path.
```

settings.php
```diff
@@ -32,14 +32,24 @@
...
-    $settings->add(new admin_setting_configcheckbox('tool_dataflows/enabled',
-        get_string('enabled', 'tool_dataflows'),
-        get_string('enabled_help', 'tool_dataflows'), '0'));
-
     $dataflowsettings = new admin_externalpage('tool_dataflows_overview',
         get_string('pluginmanage', 'tool_dataflows'),
         new moodle_url('/admin/tool/dataflows/index.php'));
 
+    if ($ADMIN->fulltree) {
+        $settings->add(new admin_setting_configcheckbox('tool_dataflows/enabled',
```
keevan, line 40: Perhaps move `'tool_dataflows/enabled',` down to the next line too, similar to how it's been done for the following block.

classes/local/step/writer_stream.php
```diff
@@ -100,7 +101,7 @@ public function get_iterator(): iterator {
 
             public function __construct(flow_engine_step $step, string $streamname, string $format, iterator $input) {
                 $this->streamname = $streamname;
-                $this->handle = fopen($streamname, 'a');
+                $this->handle = fopen(helper::path_get_absolute($streamname, $step->engine->scratchdir), 'a');
```
keevan, line 104: I feel like file operations to and from the "scratch dir" should be done in a nicer API

The train of thought here is:
- other steps might want to read / write to and from the scratch directory
- by default files should be opening from and to the scratch directory. 
- files in the scratch directory (since this is created at the engine level) should be namespaced under the step in some way, for example, using the step's alias. This means under the same scratch directory you can have things in the `curl/` folder, the `writer` folder, etc. If it needs to reference another step it would use the relative path (e.g. `../curl`). This also means a step can by default have files in its scope without worrying about affecting other files (e.g. out.csv as the stream output for multiple writer steps, wouldn't affect other out.csv files).

It's a bit long so please ping if you have any questions. I'd also like to hear your thoughts on this @brendanheywood

brendanheywood: 1) yes almost all connector steps, and reader and writer steps, will interact via the scratch dir so the cleaner the api for this the better
2) yes agree, most steps will only use the scratch dir, and I would even go one step further and say that certain steps with ONLY be allowed to use the scratch dir and not arbitrary paths. For now just
3) I'm not a fan of namespacing files inside the scratch dir. I have considered this but in practice I think it will just complicate things. If you want it then you can always just specify in a step that it should write to 'curl/download.txt' and it will auto make the directory for you. If you have a series of connector steps it is likely that they will all do stuff with the same files and the namespace is just cruft

brendanheywood: Re a default writer like out.csv I'm ok with there not being a default and you have to specify it. OR maybe a better way is that you can have a default but it is a default file name not a default base directory, so a curl step called 'call-api' would default write to a file called [scratchdir]/call-api

keevan: With regards to 3, I think we can leave it open for now and see how it goes. It might work fine without issues so there's no need to do anything special there. 

Regarding the extra comment, I do prefer specifying it since it gives it a bit more meaning than using the generic (if any) value of probably the alias. The alternative suggested might be a bit too magic for me, and it might be difficult to determine (without simply storing the value on save) and list the options as per #199.

keevan: @jaypha perhaps it can be a method accessible from the engine step which accepts a single string for the path. This means if we do decide to namespace it later it is still possible since the context of the caller is correct. 

If you think that's unlikely then it can be added to the engine class, though it _#feelsbad_ calling all the way up from the step type every time you want to do something related.

classes/local/step/writer_stream.php
```diff
@@ -100,7 +101,7 @@ public function get_iterator(): iterator {
 
             public function __construct(flow_engine_step $step, string $streamname, string $format, iterator $input) {
                 $this->streamname = $streamname;
-                $this->handle = fopen($streamname, 'a');
+                $this->handle = $step->engine->open_stream($streamname, 'a');
```
keevan, line 104: Hi @jaypha , whilst I think this is an improvement, I don't think it's as flexible for all the possible types of steps that might be used. For example, I don't think steps that want to use a file they know exists in the scratch directory should be forced to open a file stream to do so. (i.e. `file_get_contents`).

Perhaps the only change that needs to happen is something similar to this:
```
                $this->streamname = $step->engine->resolve_path($streamname);
                $this->handle = fopen($this->streamname, 'a');
                ...
```
And everywhere else in that class uses `$this->streamname` to reference the file. This should make it easier for step maintainers to use the underlying php functions as expected. What do you think?

PR #259, opened 2022-07-04
feat: show a list of available, known fields usable in expressions

### LINKED ISSUE #199: Have help text which shows all the available variables usable in expressions

keevan:
Here's a preview:

image

Note that it doesn't have run-time variables available, in particular, it doesn't list flow variables within the same scope as the flow, but that's okay, it's still reference-able currently.

TIP: that you should also see the value (if set) of the corresponding key when hovering over the particular entry (e.g. with the mouse over the curl url)

image

Resolves #199

### DIFF
--- a/classes/form/step_form.php
+++ b/classes/form/step_form.php
@@ -39,6 +39,8 @@ class step_form extends \core\form\persistent {
      * Define the form.
      */
     public function definition() {
+        global $OUTPUT;
+
         $mform = $this->_form;
         $dataflowid = $this->_customdata['dataflowid'];
         $type = $this->_customdata['type'];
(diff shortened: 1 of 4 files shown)


classes/form/step_form.php
```diff
@@ -93,12 +95,37 @@ public function definition() {
         );
         $select->setMultiple(true);
 
+        // List all the available fields available for configuration, in dot syntax.
+        $variables = $persistent->get_variables();
```
brendanheywood, line 99: we will at some point want to show this in other places outside a form, so refactor at some point

PR #269, opened 2022-07-06
Issue #141: pretty print JSON

### LINKED ISSUE #141: json writing should have option to pretty print
Even if pretty print is off, still add a linebreak after each record

jaypha:
resolves #141

### DIFF
--- a/classes/local/formats/encoder_base.php
+++ b/classes/local/formats/encoder_base.php
@@ -27,6 +27,7 @@ namespace tool_dataflows\local\formats;
 abstract class encoder_base {
     /** @var bool Has data been added yet? */
     protected $sheetdatadded = false;
+    protected $prettyprint = false;
 
     /**
      * Return the start of the output.
@@ -52,5 +53,14 @@ abstract class encoder_base {
      */
     public function close_output(): string {
         return '';
+    }
+
+    /**
+     * Set to use/not use pretty print for suppported formats.
+     *
+     * @param bool $prettyprint
+     */
+    public function set_prettyprint(bool $prettyprint) {
+        $this->prettyprint = $prettyprint;
     }
 }
--- a/classes/local/formats/encoders/json.php
+++ b/classes/local/formats/encoders/json.php
@@ -29,12 +29,23 @@ use tool_dataflows\local\formats\encoder_base;
 class json extends encoder_base {
 
     /**
+     * @var string Indent space used for pretty printing. As far as can be determined, 4 spaces for
+     * indenting is hardwired in json_encode() for pretty printing.
+     */
+    protected const INDENT = '    ';
+
+    /**
      * Return the start of the output.
      *
      * @return string
      */
     public function start_output(): string {
-        return '[';
+        $output = '[' . PHP_EOL;
+
+        if ($this->prettyprint) {
+            $output .= self::INDENT;
+        }
+        return $output;
     }
 
     /**
@@ -46,11 +57,17 @@ class json extends encoder_base {
     public function encode_record($record, int $rownum): string {
         $output = '';
         if ($this->sheetdatadded) {
-            $output .= ',';
+            $output .= ',' . PHP_EOL;
         }
 
-        $output .= json_encode($record) . PHP_EOL;
+        // Add encoded record to ouput.
+        $flags = $this->prettyprint ? JSON_PRETTY_PRINT : 0;
+        $output .= json_encode($record, $flags);
 
+        // Add indenting inbetween records as json_encode() will only pretty print the record itself.
+        if ($this->prettyprint) {
+            $output = str_replace(PHP_EOL, PHP_EOL . self::INDENT, $output);
+        }
         $this->sheetdatadded = true;
         return $output;
     }
@@ -61,6 +78,6 @@ class json extends encoder_base {
      * @return string
      */
     public function close_output(): string {
-        return ']' . PHP_EOL;
+        return  PHP_EOL . ']' . PHP_EOL;
     }
 }
--- a/classes/local/step/writer_stream.php
+++ b/classes/local/step/writer_stream.php
@@ -48,6 +48,7 @@ class writer_stream extends writer_step {
         return [
             'streamname' => ['type' => PARAM_TEXT],
             'format' => ['type' => PARAM_TEXT],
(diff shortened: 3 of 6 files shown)


classes/local/formats/encoder_base.php
```diff
@@ -53,4 +54,13 @@ abstract public function encode_record($record, int $rownum): string;
     public function close_output(): string {
         return '';
     }
+
+    /**
+     * Set to use/not use pretty print for suppported formats.
+     *
+     * @param bool $prettyprint
+     */
+    public function prettyprint(bool $prettyprint) {
```
keevan, line 63: Could you please change this function name to describe a behaviour instead of a property. Maybe `use_pretty_print()`?

classes/local/formats/encoders/json.php
```diff
@@ -46,11 +54,15 @@ public function start_output(): string {
     public function encode_record($record, int $rownum): string {
         $output = '';
         if ($this->sheetdatadded) {
-            $output .= ',';
+            $output .= ',' . PHP_EOL;
         }
 
-        $output .= json_encode($record) . PHP_EOL;
+        $flags = $this->prettyprint ? JSON_PRETTY_PRINT : 0;
```
keevan, line 60: Please add a comment above this block, e.g. "Print out the record"

classes/local/formats/encoders/json.php
```diff
@@ -46,11 +54,15 @@ public function start_output(): string {
...
+        $output .= json_encode($record, $flags);
 
+        if ($this->prettyprint) {
+            $output = str_replace(PHP_EOL, PHP_EOL . self::INDENT, $output);
```
keevan, line 64: Question: although this would probably pretty print as expected, do you know if there would be any value adding the first level of indentation? To me, it feels like wasted compute.

classes/local/step/connector_debug_filedump.php
```diff
@@ -0,0 +1,90 @@
+<?php
```
keevan, line 1: It seems that you've added this extra step for debugging. I think what you've done can be useful, but I think this should be split out in a different issue + PR because this issue doesn't require this step.

classes/local/step/writer_stream.php
```diff
@@ -91,23 +92,26 @@ public function get_iterator(): iterator {
...
                 $this->handle = fopen($this->streamname, 'a');
                 if ($this->handle === false) {
                     $step->log(error_get_last()['message']);
                     throw new \moodle_exception(get_string('writer_stream:failed_to_open_stream', 'tool_dataflows', $streamname));
                 }
-                $classname = writer_stream::resolve_encoder($format);
+
+                $classname = writer_stream::resolve_encoder($config->format);
                 $this->writer = new $classname();
+                $this->writer->prettyprint((bool) ($config->prettyprint ?? false));
```
keevan, line 114: When storing and receiving the prettyprint config value, does it come back out as a bool? And if not, wouldn't just one of the "cast" OR "null coalescing" be enough to provide to the function. I would probably just remove the bool casting.

lang/en/tool_dataflows.php
```diff
@@ -231,6 +232,8 @@
 $string['writer_stream:streamname_help'] = 'For example, setting <code>file:///path/to/file.txt</code> as the stream name will write the contents to this target.';
 $string['writer_stream:format'] = 'Format';
 $string['writer_stream:format_help'] = 'The output format of the contents written.';
+$string['writer_stream:prettyprint'] = 'Pretty print';
+$string['writer_stream:prettyprint_help'] = 'Format the output to be human readable.';
```
keevan, line 236: Please bump the version due to the changes here.

classes/local/formats/encoders/json.php
keevan: Okay so reading further comments in the issue - e.g. https://github.com/catalyst/moodle-tool_dataflows/issues/141#issuecomment-1174587022 - it seems this is to be expected which is fine. 

Are you able to **add a supporting test** please, which confirms the output comes out as expected? (e.g. it could  check the output with something like `json_encode($data, JSON_PRETTY_PRINT);`

classes/local/formats/encoder_base.php
jaypha: I'll use 'set_prettyprint'.

classes/local/step/connector_debug_filedump.php
jaypha: I put it in this PR #272

lang/en/tool_dataflows.php
```diff
@@ -97,6 +97,7 @@
 $string['step_name_trigger_cron'] = 'Scheduled task trigger';
 $string['step_name_writer_debugging'] = 'Debugging writer';
 $string['step_name_writer_stream'] = 'Stream writer';
+$string['step_name_connector_debug_filedump'] = 'File dump debug connector';
```
keevan, line 100: Seems like this string was left from the split in the other PR.

https://github.com/catalyst/moodle-tool_dataflows/pull/272/files#diff-56c8d9ed220eec5c0f99664e70d45607a79ae6e0a37d70c1f53d6bf01f7cffbb

Since that PR already has a string set aside for it's name, I think this should be removed to avoid any conflict once both PRs are merged in.

keevan (commented):
Aside from a very small lang string artifact, the changes look good.

Happy to merge this in after that has been removed.

Thanks @jaypha

classes/local/formats/encoders/json.php
```diff
@@ -28,13 +28,21 @@
  */
 class json extends encoder_base {
 
+    /** @var string indent space used for pretty printing. TODO try to find out if there is a config setting for the size of the indent. */
```
keevan, line 31: Did you find out if there was a configuration setting for this? And if not, maybe the comment should be adjjusted to show findings. (e.g. the default for json_encode with pretty print is 4 spaces so ... etc).

classes/local/step/writer_stream.php
jaypha: done

lang/en/tool_dataflows.php
jaypha: Done.

classes/local/formats/encoders/json.php
jaypha: I changed the comment.

PR #270, opened 2022-07-06
#261 s3 copy improvements

### LINKED ISSUE #261: s3 copy step prod polish
* [ ] Convert from yaml text area to proper form elements for from and to
* [ ] use the scratch directory for all local paths
* [ ] validate that at least one of to or from is an s3 path

keevan:
- feat: convert config into form inputs and refactored usage
- feat: automatically set 'required' for relevant fields in the step type

Resolves #261

### DIFF
--- a/classes/local/step/base_step.php
+++ b/classes/local/step/base_step.php
@@ -235,7 +235,7 @@ abstract class base_step {
      *
      * @param \MoodleQuickForm &$mform
      */
-    public function form_setup(&$mform) {
+    public function form_setup(\MoodleQuickForm &$mform) {
         $this->form_add_custom_inputs($mform);
         $this->form_set_input_types($mform);
         $this->form_set_input_rules($mform);
--- a/classes/local/step/base_step.php
+++ b/classes/local/step/base_step.php
@@ -231,6 +231,17 @@ abstract class base_step {
     }
 
     /**
+     * Sets up the form fields and inputs.
+     *
+     * @param \MoodleQuickForm &$mform
+     */
+    public function form_setup(&$mform) {
+        $this->form_add_custom_inputs($mform);
+        $this->form_set_input_types($mform);
+        $this->form_set_input_rules($mform);
+    }
+
+    /**
      * Allows each step type to determine a list of optional/required form
      * inputs for their configuration
      *
@@ -262,6 +273,20 @@ abstract class base_step {
         foreach ($fields as $fieldname => $config) {
             if (isset($config['type'])) {
                 $mform->setType("config_$fieldname", $config['type']);
+            }
+        }
+    }
+
+    /**
+     * Sets rules for each input field defined if supported
+     *
+     * @param \MoodleQuickForm &$mform
+     */
+    public function form_set_input_rules(\MoodleQuickForm &$mform) {
+        $fields = static::form_define_fields();
+        foreach ($fields as $fieldname => $config) {
+            if (!empty($config['required'])) {
+                $mform->addRule("config_$fieldname", get_string('required'), 'required', null, 'client');
             }
         }
     }
--- a/classes/local/step/connector_s3.php
+++ b/classes/local/step/connector_s3.php
@@ -16,6 +16,8 @@
 
 namespace tool_dataflows\local\step;
 
+use tool_dataflows\helper;
+
 /**
  * S3 connector step type
  *
@@ -230,12 +232,30 @@ class connector_s3 extends connector_step {
         // Check source/target has an s3:// path set once both source and targets have been set.
         if (!empty($config->source) && !empty($config->target)) {
             // Check if the source or target is an expression, and evaluate it if required.
-            $hass3path = $this->has_s3_path($config->source) || $this->has_s3_path($config->target);
+            $sourceins3 = $this->has_s3_path($config->source);
+            $targetins3 = $this->has_s3_path($config->target);
+            $hass3path = $sourceins3 || $targetins3;
         }
         if (isset($hass3path) && $hass3path === false) {
             $errormsg = get_string('connector_s3:missing_s3_source_or_target', 'tool_dataflows', null, true);
             $errors['config_source'] = $errors['config_source'] ?? $errormsg;
             $errors['config_target'] = $errors['config_target'] ?? $errormsg;
+        }
+
+        // Confirm the non s3 source is safe to use.
+        if (empty($sourceins3) && !isset($errors['config_source'])) {
+            $error = helper::path_validate($config->source);
+            if ($error !== true) {
+                $errors['config_source'] = $error;
+            }
+        }
+
+        // Confirm the non s3 target is safe to use.
+        if (empty($targetins3) && !isset($errors['config_target'])) {
+            $error = helper::path_validate($config->target);
+            if ($error !== true) {
+                $errors['config_target'] = $error;
+            }
         }
 
         return empty($errors) ? true : $errors;
--- a/classes/local/step/connector_s3.php
+++ b/classes/local/step/connector_s3.php
@@ -15,13 +15,13 @@
 // along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
 
 namespace tool_dataflows\local\step;
-use tool_dataflows\local\execution\engine_step;
 
 /**
  * S3 connector step type
  *
  * @package    tool_dataflows
  * @author     Peter Burnettt <peterburnett@catalyst-au.net>
+ * @author     Kevin Pham <kevinpham@catalyst-au.net>
  * @copyright  Catalyst IT, 2022
  * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
@@ -29,6 +29,9 @@ class connector_s3 extends connector_step {
 
     /** @var bool whether or not this step type (potentially) contains a side effect or not */
     protected $hassideeffect = true;
+
+    /** @var string the prefix identifier for an s3 path, e.g. s3://path/to/file. */
+    const S3_PREFIX = 's3://';
 
     /**
      * Return the definition of the fields available in this form.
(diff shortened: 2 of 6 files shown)


classes/local/step/connector_s3.php
```diff
@@ -87,10 +107,27 @@ public function execute(): bool {
...
+        // Check source path.
+        $sourceins3 = $this->has_s3_path($config->source);
+        $source = $this->resolve_path($config->source, $sourceins3);
 
-            @$stream = fopen($config->source, 'r');
+        // Resolve target path.
+        $targetins3 = $this->has_s3_path($config->target);
+        $target = $this->resolve_path($config->target, $targetins3);
+
+        // Fix target (check if it is a directory, and use the source's basename).
```
Peterburnett, line 118: Not a critique, this is a great QoL

classes/local/step/connector_s3.php
```diff
@@ -87,10 +107,27 @@ public function execute(): bool {
...
+        if (substr($target, -1) === '/') {
+            $target .= basename($source);
+        }
+
+        // Do not execute s3 operations during a dry run.
+        if ($this->enginestep->engine->isdryrun) {
```
Peterburnett, line 124: I thought we were doing this at the engine level, and not running things that had side effects? I might be ootl here

classes/local/step/connector_s3.php
```diff
@@ -100,27 +137,117 @@ public function execute(): bool {
...
+            'bucket',
+            'region',
+            'key',
+            'secret',
+            'source',
+            'target'
+        ] as $field) {
+            if (empty($config->$field)) {
+                $errors["config_$field"] = get_string('config_field_missing', 'tool_dataflows', "$field", true);
+            }
```
Peterburnett, line 221: Considering we are using Mforms here, why dont we just add support for addRule('required')? unless thats already supported with the required flag in the form definition array at the top and this is just backup?

Peterburnett: Because if so, we could just make the required rule a serverside validation instead of clientside

classes/local/step/connector_s3.php
```diff
@@ -87,10 +107,27 @@ public function execute(): bool {
...
-        // Copy TO remote.
-        if (!$config->sourceremote) {
+        // Check source path.
+        $sourceins3 = $this->has_s3_path($config->source);
+        $source = $this->resolve_path($config->source, $sourceins3);
 
-            @$stream = fopen($config->source, 'r');
+        // Resolve target path.
+        $targetins3 = $this->has_s3_path($config->target);
+        $target = $this->resolve_path($config->target, $targetins3);
```
Peterburnett, line 116: Here I think we need to do some additional cleaning on the local path with PARAM_PATH which will catch any naughty LFI shenanigans out of the scratch directory. I didnt bother last time because it wasnt sandboxed, but now I do :)

lang/en/tool_dataflows.php
```diff
@@ -266,3 +266,16 @@
...
+// S3 File Copy.
+$string['connector_s3:bucket'] = 'Bucket';
+$string['connector_s3:region'] = 'Region';
+$string['connector_s3:key'] = 'Key';
+$string['connector_s3:secret'] = 'Secret';
+$string['connector_s3:source'] = 'Source / From';
+$string['connector_s3:source_help'] = 'Path to the source file. This can be a local (e.g. output.csv) or s3 path (e.g. s3://path/to/file).';
+$string['connector_s3:target'] = 'Target / To';
+$string['connector_s3:target_help'] = 'Path to the target file. Similar to the source, it can be a local or s3 path. If a directory is used, the source filename is applied.';
+$string['connector_s3:missing_s3_source_or_target'] = 'At least one source or target path must reference a location in s3';
```
Peterburnett, line 279: Capitalisation on S3

classes/local/step/connector_s3.php
keevan: A few versions and a refactor brought me to that 2 liner haha.

classes/local/step/connector_s3.php
keevan: I believe each step had control of how they handled dry runs. If you say we'd do this at the engine level then it sounds like it would just fail/stop at the side effect step. At least this way it's given a chance to do something and not effect the outside world.

I believe the writer step/stream has this check as well, and only writes to things in its scratch dir (not considered a side effect) but for everything else, it would be skipped.

classes/local/step/connector_s3.php
keevan: First comment response: yeah it's handled in the base form as part of these changes. This is actually for dataflow validation, which doesn't pass through into the moodle forms land (e.g. validating stored values).

It might be a good QoL to have this in some form, as I've seen it duplicated in other places. But I'm not aware of a clean way to apply both moodle form and other validation to both entrypoints.

classes/local/step/connector_s3.php
keevan: You are correct, after reviewing a similar PR, resolve path alone will not prevent me from accessing any other local files. I'll make some changes (going to include a helper function), but the core issue may or may not be unrelated. PARAM_PATH might be useful in adding to that helper function which I'll reference in a bit.

lang/en/tool_dataflows.php
keevan: Ah right, in the strings themselves. Will do. Thanks for the catch!

lang/en/tool_dataflows.php
```diff
@@ -266,3 +266,16 @@
...
+// S3 File Copy.
+$string['connector_s3:bucket'] = 'Bucket';
+$string['connector_s3:region'] = 'Region';
+$string['connector_s3:key'] = 'Key';
+$string['connector_s3:secret'] = 'Secret';
+$string['connector_s3:source'] = 'Source / From';
+$string['connector_s3:source_help'] = 'Path to the source file. This can be a local (e.g. output.csv) or S3 path (e.g. s3://path/to/file).';
+$string['connector_s3:target'] = 'Target / To';
+$string['connector_s3:target_help'] = 'Path to the target file. Similar to the source, it can be a local or S3 path. If a directory is used, the source filename is applied.';
+$string['connector_s3:missing_s3_source_or_target'] = 'At least one source or target path must reference a location in S3';
```
jaypha, line 279: Full stop required.

classes/local/step/connector_s3.php
```diff
@@ -100,27 +139,135 @@ public function execute(): bool {
...
+        $errors = [];
+
+        // Check mandatory fields.
+        foreach ([
+            'bucket',
+            'region',
+            'key',
+            'secret',
+            'source',
+            'target'
```
jaypha, line 219: comma needed

classes/local/step/connector_s3.php
```diff
@@ -100,27 +139,135 @@ public function execute(): bool {
...
+        // GET - Handle s3 path to local source file.
+        if ($sourceins3 && !$targetins3) {
             // Copy FROM remote.
             try {
                 $s3client->getObject([
                     'Bucket' => $config->bucket,
-                    'Key' => $config->source,
-                    'SaveAs' => $config->target
+                    'Key' => $source,
+                    'SaveAs' => $target
```
jaypha, line 158: comma?

classes/local/step/base_step.php
```diff
@@ -213,6 +213,17 @@ public function validate_config($config) {
         return true;
     }
 
+    /**
+     * Sets up the form fields and inputs.
+     *
+     * @param \MoodleQuickForm &$mform
+     */
+    public function form_setup(&$mform) {
```
jaypha, line 221: Should there be a type hint here?

classes/local/step/connector_s3.php
```diff
@@ -100,27 +139,135 @@ public function execute(): bool {
...
+     * @param   bool $ins3 whether or not the path is a location in s3.
+     * @return  string resolved path
+     */
+    public function resolve_path(string $path, bool $ins3): string {
+        if (!$ins3) {
+            // Resolve local path (must be in scratch dir).
+            return $this->enginestep->engine->resolve_path($path);
+        }
+
+        if ($ins3) {
```
jaypha, line 198: This is redundant.

classes/local/step/connector_s3.php
```diff
@@ -100,27 +139,135 @@ public function execute(): bool {
...
+        if ($sourceins3 && $targetins3) {
+            try {
+                $s3client->copyObject([
+                    'Bucket'     => $config->bucket,
+                    'CopySource' => "{$config->bucket}/{$source}",
+                    'Key'        => $target,
+                ]);
+            } catch (\Aws\S3\Exception\S3Exception $e) {
+                $this->enginestep->log(get_string('s3_copy_failed', 'tool_dataflows'));
+                return false;
```
jaypha, line 177: By returning false, the step will be cancelled rather than aborted.
I just want to check that this is the intended behaviour.

classes/local/step/connector_s3.php
```diff
@@ -87,10 +109,27 @@ public function execute(): bool {
...
+        }
+
+        // Do not execute s3 operations during a dry run.
+        if ($this->enginestep->engine->isdryrun) {
+            return true;
+        }
+
+        // PUT - Handle local source file to s3 path.
+        if (!$sourceins3 && $targetins3) {
+            @$stream = fopen($source, 'r');
```
jaypha, line 132: This stream is not closed on a successful run.

classes/local/step/connector_s3.php
jaypha: Adding it to the gives the user a visual cue. But we also need to be able to validate independently of the Mform. This unfortunately results in a double check.
I'm OK with what @keevan has done here.

classes/local/step/connector_s3.php
keevan: Yes I believe so, this was copied over from the initial handling of the s3 step, so I've the exception handling the same.

Thanks for clarifying the differences between cancelled & aborted. Ideally all steps should cancel by default and abort (the whole dataflow) if the user wants it to instead of cancelling. (e.g. if this step fails for whatever reason, abort the dataflow).

classes/local/step/connector_s3.php
keevan: I've updated this and tested an equivalent flow in a sandbox as shown below to ensure it works as expected even with a return statement in commit [`b84b572ac93329245ee139bf8deb5fba31eb97ba`](https://github.com/catalyst/moodle-tool_dataflows/pull/270/commits/b84b572ac93329245ee139bf8deb5fba31eb97ba)

With exception, does process finally block:

image

No exception, does process finally block:

image

(Ignore the weird X, the code runs fine)

classes/local/step/connector_s3.php
keevan: Good pickup. I've updated this based on method name and special-ness. I've flipped the order and moved the s3 handling into a conditional block placed earlier.

classes/local/step/base_step.php
keevan: I've added this, thanks.

classes/local/step/connector_s3.php
keevan: Added.

classes/local/step/connector_s3.php
keevan: Added.

lang/en/tool_dataflows.php
keevan: Added

Issue #42, opened 2022-04-12
Copy from from a to b connector, can handle any php stream


brendanheywood:
ie we can copy a file from s3, then connect it to a csv or json parsing source step

At the end we can have a connector which streams into a file, then when it is finished we can sign it and then copy it to s3

PR #272, opened 2022-07-07
Issue #271: Add connector_debug_filedump

### LINKED ISSUE #271: Add 'dump' step to help with testing workflows.
Add a step to dump the contents of a file to the output.

jaypha:
Resolves #271

### DIFF
--- a/classes/import_form.php
+++ b/classes/import_form.php
@@ -60,7 +60,7 @@ class import_form extends \moodleform {
     public function validation($data, $files) {
         global $USER;
 
-        $validationerrors = array();
+        $validationerrors = [];
 
         // Get the file from the filestystem. $files will always be empty.
         $fs = get_file_storage();
--- a/classes/local/scheduler.php
+++ b/classes/local/scheduler.php
@@ -77,6 +77,12 @@ class scheduler {
         if (is_null($reftime)) {
             $reftime = time();
         }
-        return $DB->get_records_select_menu(self::TABLE, 'nextruntime <= :time', ['time' => $reftime], '', 'dataflowid, nextruntime');
+        return $DB->get_records_select_menu(
+            self::TABLE,
+            'nextruntime <= :time',
+            ['time' => $reftime],
+            '',
+            'dataflowid, nextruntime'
+        );
     }
 }
--- a/classes/local/step/base_step.php
+++ b/classes/local/step/base_step.php
@@ -61,8 +61,12 @@ abstract class base_step {
     protected $hassideeffect = false;
 
     /**
+     * Constructs the step type
+     *
+     * If used during execution this will link the execution engine which requires a stepdef to be defined.
+     *
      * @param step|null $stepdef A stepdef of this type. Pass null if step type is to be used without a context.
-     * @param engine|null $engine An engine that is to execute the step of this type. stepdef must be defined. Leave as null if not in an execution.
+     * @param engine|null $engine An engine that executes this step type, null if not in an execution.
      */
     public function __construct(?step $stepdef = null, ?engine $engine = null) {
         $this->stepdef = $stepdef;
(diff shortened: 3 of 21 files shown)


classes/local/step/connector_debug_filedump.php
```diff
@@ -0,0 +1,95 @@
...
+     *
+     * @return bool Returns true if successful, false otherwise.
+     */
+    public function execute(): bool {
+        $config = $this->enginestep->stepdef->config;
+
+        $streamname = $this->enginestep->engine->resolve_path($config->streamname);
+        $content = file_get_contents($streamname);
+
+        mtrace($content);
```
keevan, line 44: Could we use the enginestep's log function here instead of doing an mtrace? 

(Or is the purpose here specifically to only ever mtrace the file contents?)

keevan (commented):
I think broadly speaking the functionality is fine. Just a small point regarding mtrace vs log and what the intentions are behind that as it might be more useful to have it in the logs if used.

Also I'm unsure about the name of the step. When I think outputting file contents to the screen, I think plaintext files generally - there is some degree of assumption here for that purpose - and that it behaves equivalently to the `cat` cli command (https://explainshell.com/explain/1/cat)

My preference would be to rename it to connector_file_cat, or connector_cat, but I understand this might confuse those who aren't across these terminologies.

jaypha:
> Also I'm unsure about the name of the step - because _dump_ isn't clear to me. When I think outputting file contents to the screen, I think plaintext files generally - there is some degree of assumption here for that purpose - and that it behaves equivalently to the `cat` cli command (https://explainshell.com/explain/1/cat)
> 
> My preference would be to rename it to connector_file_cat, or connector_cat, but I understand this might confuse those who aren't across these terminologies.

It is unlikely that anyone who is not *nix programmer would understand what 'cat' means. The term would need to be meaningful to tech savvy and non-savvy alike.

How about 'connector_debug_file_display'? It makes it clear that it is a debug tool, is about files, and displays them.

classes/local/step/connector_debug_filedump.php
jaypha: Logging already records a lot of information. This is intended to be a simple visual aid tool.

If you want this to be more sophisticated, I could try to implement #42, and use php://output for the output stream.

What do you think?

keevan: Hmm, I actually don't want it to store the file contents in the logs (e.g. DB) later on so it's probably best to avoid this in the first place. I'm okay with just leaving it as an mtrace for now. 

This may be improved upon and might become redundant later on once we implement better dataflow debugging, but I'm okay with this landing for now as you've found it to be helpful.

tests/tool_dataflows_connector_filedump_test.php
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * Test for connector_debug_file_display step.
+ *
+ * @package   tool_dataflows
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class tool_dataflows_connector_debug_file_display_test extends \advanced_testcase {
```
keevan, line 30: One small thing:
```
FILE: admin/tool/dataflows/tests/tool_dataflows_connector_filedump_test.php
---------------------------------------------------------------------------------------------------------------------------------------------------------------
---------------------------------------------------------------------------------------------------------------------------------------------------------------
 30 | WARNING | [x] PHPUnit testcase name "tool_dataflows_connector_debug_file_display_test" does not match file name "tool_dataflows_connector_filedump_test"
---------------------------------------------------------------------------------------------------------------------------------------------------------------
---------------------------------------------------------------------------------------------------------------------------------------------------------------
```

PR #274, opened 2022-07-08
267 handle secrets better

### LINKED ISSUE #267: Redact secrets from displaying on the dataflow details page (config), state dumps (DB) and logs
Secrets could be things configured as password / passwordunmask fields or simply set as such in the `form_define_fields` array e.g. `secret => true`.

This might be a bandaid to what really should be a some level of secrets management (a manager class), that evalulates the secrets at run time only when needed. If the secret is an expression, it should only display the expression itself and not the value under it for debugging purposes.

keevan:
feat: add secret service to ensure secrets are managed and contained

Table:

image

Export:

image

Logging:

image

Runs:

image

Resolves #267

### DIFF
--- a/classes/local/service/secret_service.php
+++ b/classes/local/service/secret_service.php
@@ -1,0 +1,50 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


keevan:
Here is an example dataflow you can import and check:

[#261_20220708_0635.yml.txt](https://github.com/catalyst/moodle-tool_dataflows/files/9069575/261_20220708_0635.yml.txt)

keevan:
Probably worth mentioning that this does NOT currently redact any secrets that might appear in the log output as a result of a step e.g. "echoing" it out. 

That might need to be covered here e.g. via the log class or separately when the situation arises which requires this to be done. Currently the run output only dumps the configuration at the end of the run, and doesn't actively dump anything containing secrets during apart from the aforementioned.

PR #284, opened 2022-07-12
Issue #169: Cron based workflows are spawned as ad hoc tasks.

### LINKED ISSUE #169: Improved cron trigger to allow parallel workflows
Builds on https://github.com/catalyst/moodle-tool_dataflows/issues/94

* [x] if there is more than one workflow to be run in a given minute, spawn adhoc tasks for all the others except one
* [x] run one of them right now so we don't add more latency
* [x] introduce locking for each workflow so only once instance can run at a given point

jaypha:
resolves #169

### DIFF
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -95,6 +95,7 @@ class dataflow extends persistent {
         return [
             'name' => ['type' => PARAM_TEXT],
             'enabled' => ['type' => PARAM_BOOL, 'default' => false],
+            'concurrencyenabled' => ['type' => PARAM_BOOL, 'default' => false],
             'config' => ['type' => PARAM_TEXT, 'default' => ''],
             'timecreated' => ['type' => PARAM_INT, 'default' => 0],
             'userid' => ['type' => PARAM_INT, 'default' => 0],
@@ -348,7 +349,8 @@ class dataflow extends persistent {
 
         // Check if each step is valid based on its own definition of valid (e.g. which could be based on configuration).
         foreach ($steps as $step) {
-            if (isset($step->steptype) && $step->steptype->get_group() == 'triggers') {
+            $steptype = $step->steptype;
+            if (isset($steptype) && $steptype->get_group() == 'triggers') {
                 ++$numtriggers;
             }
             $stepvalidation = $step->validate_step();
@@ -393,6 +395,35 @@ class dataflow extends persistent {
         }
 
         return empty($errors) ? true : $errors;
+    }
+
+    /**
+     * Returns whether concurrency is possible by the step configurations.
+     *
+     * @return array|true True if concurrency is supported, or an array of strings giving reasons why it isn't.
+     */
+    public function is_concurrency_supported() {
+        $steps = $this->steps;
+        $reasons = [];
+        foreach ($steps as $step) {
+            $steptype = $step->steptype;
+            if (isset($steptype)) {
+                $supported = $steptype->is_concurrency_supported();
+                if ($supported !== true) {
+                    $reasons[$step->name] = $supported;
+                }
+            }
+        }
+        return count($reasons) ? $reasons : true;
+    }
+
+    /**
+     * Returns whether concurrent running is enabled and supported.
+     *
+     * @return bool True is concurrency is both supported and enabled by setting, False otherwise.
+     */
+    public function is_concurrency_enabled(): bool {
+        return ($this->concurrencyenabled && ($this->is_concurrency_supported() === true));
     }
 
     /**
--- a/classes/form/dataflow_form.php
+++ b/classes/form/dataflow_form.php
@@ -49,6 +49,21 @@ class dataflow_form extends \core\form\persistent {
         $mform->addElement('advcheckbox', 'enabled', get_string('dataflow_enabled', 'tool_dataflows'));
         $mform->setDefault('enabled', 0);
 
+        // Enable/Disable concurrency.
+        $mform->addElement('advcheckbox', 'concurrencyenabled', get_string('concurrency_enabled', 'tool_dataflows'));
+        $mform->setDefault('concurrencyenabled', 0);
+
+        if ($this->_customdata['concurrencyallowed'] !== true) {
+            $reasons = '';
+            foreach ($this->_customdata['concurrencyallowed'] as $name => $reason) {
+                $reasons .= \html_writer::tag('li', "$name: $reason");
+            }
+            $reasons = \html_writer::tag('ul', $reasons);
+
+            $mform->addElement('static', 'concurrencyenableddisabled', '',
+                    get_string('concurrency_enabled_disabled_desc', 'tool_dataflows') . $reasons);
+        }
+
         // Configuration - YAML format.
         $mform->addElement(
             'textarea',
--- a/classes/local/execution/engine.php
+++ b/classes/local/execution/engine.php
@@ -111,6 +111,15 @@ class engine {
     /** @var string Scratch directory for temporary files. */
     protected $scratchdir = null;
 
+    /** @var \core\lock\lock|false Lock for the dataflow. Sometimes, only one dataflow of each def should be running at a time. */
+    protected $lock = false;
+
+    /** @var \core\lock\lock_factory Factory to produce locks.  */
+    protected static $lockfactory = null;
+
+    /** @var bool Has this engine been blocked by a lock. */
+    protected $blocked = false;
+
     /**
      * Constructs the engine.
      *
@@ -168,10 +177,95 @@ class engine {
     }
 
     /**
-     * Initialises the engine
+     * Destructor function.
+     * Releases any locks that are still held.
+     */
+    public function __destruct() {
+        $this->release_lock();
+    }
+
+    /**
+     * Tries to obtain a lock for this dataflow
+     *
+     * @param int $timeout
+     * @return \core\lock\lock|false.
+     */
+    public function get_lock($timeout = 0) {
+        global $DB;
+
+        $lockfactory = \core\lock\lock_config::get_lock_factory('tool_dataflows_engine');
+        $this->lock = $lockfactory->get_lock('tool_dataflows_engine_' . $this->dataflow->id, $timeout);
+        if ($this->lock) {
+            $DB->insert_record(
+                'tool_dataflows_lock_metadata',
+                [
+                    'dataflowid' => $this->dataflow->id,
+                    'timestamp' => time(),
+                    'processid' => getmypid(),
+                ]
+            );
+        }
+        return $this->lock;
+    }
+
+    /**
+     * Release any locks that have been previously obtained.
+     *
+     * @throws \dml_exception
+     */
+    public function release_lock() {
+        global $DB;
+
+        if ($this->lock) {
+            $this->lock->release();
+            $this->lock = false;
+            $DB->delete_records('tool_dataflows_lock_metadata', ['dataflowid' => $this->dataflow->id]);
+        }
+    }
+
+    /**
+     * Gets the metadata associated with the dataflow lock.
+     *
+     * @param int $dataflowid
+     * @return false|mixed|\stdClass
+     * @throws \dml_exception
+     */
+    public static function get_lock_metadata(int $dataflowid) {
+        global $DB;
+
+        return $DB->get_record('tool_dataflows_lock_metadata', ['dataflowid' => $dataflowid]);
+    }
+
+    /**
+     * Returns if the execution was unable to obtain a lock.
+     *
+     * @return object|false Lock metadata, for false is not blocked.
+     */
+    public function is_blocked() {
+        if ($this->blocked) {
+            return self::get_lock_metadata($this->dataflow->id);
+        } else {
+            return false;
+        }
+    }
+
+    /**
+     * Initialises the dataflow.
      */
     public function initialise() {
+        $this->status_check(self::STATUS_NEW);
+
+        if (!$this->dataflow->is_concurrency_enabled()) {
+            if (!$this->get_lock()) {
+                $metadata = self::get_lock_metadata($this->dataflow->id);
+                $this->log('Execution blocked by a lock. Time: ' . userdate($metadata->timestamp) .
+                        ", Process ID: {$metadata->processid}");
+                $this->blocked = true;
+                return;
+            }
+        }
         try {
+            $this->blocked = false;
             $this->status_check(self::STATUS_NEW);
 
             foreach ($this->enginesteps as $enginestep) {
@@ -261,6 +355,10 @@ class engine {
      */
     public function execute() {
         $this->initialise();
+        if ($this->is_blocked() || $this->status == self::STATUS_ABORTED) {
+            return;
+        }
+
         $this->status_check(self::STATUS_INITIALISED);
 
         // Initalise a new run (only for non-dry runs). This should only be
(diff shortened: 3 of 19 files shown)


classes/task/process_dataflow_ad_hoc.php
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Ad hoc task for running a dataflow.
+ *
+ * @package   tool_dataflows
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class process_dataflow_ad_hoc extends \core\task\adhoc_task {
+    public function execute() {
```
keevan, line 31: Missing phpdocs and a blank line here as well.

classes/local/execution/engine.php
```diff
@@ -110,6 +110,11 @@ class engine {
     /** @var string Scratch directory for temporary files. */
     protected $scratchdir = null;
 
+
+    protected $lock = null;
+
+    protected static $lockfactory = null;
+
```
keevan, line 117: Could you please add phpdoc comments here.

classes/local/execution/engine.php
```diff
@@ -165,21 +170,36 @@ public function __construct(dataflow $dataflow, bool $isdryrun = false, $automat
         $this->create_flow_caps();
     }
 
+    public function __destruct() {
```
keevan, line 173: Missing phpdocs here

classes/local/execution/engine.php
```diff
@@ -165,21 +170,36 @@ public function __construct(dataflow $dataflow, bool $isdryrun = false, $automat
...
+        if (!is_null($this->lock)) {
+            $this->lock->release();
+        }
+    }
+
     public function initialise() {
```
keevan, line 179: Missing phpdocs here

tests/local/execution/tool_dataflows_engine_test.php
```diff
@@ -140,17 +140,20 @@ function($s) {
...
+        $engine->finalise();
+        ob_get_clean();
     }
 
     /**
      * Create a dataflow for testing.
      *
      * @return array
      */
+
```
keevan, line 156: Unnecessary empty line here.

tests/local/execution/tool_dataflows_engine_test.php
```diff
@@ -140,17 +140,20 @@ function($s) {
         // Test initialisation.
         ob_start();
         $engine->initialise();
-        ob_get_clean();
         foreach ($engine->enginesteps as $id => $enginestep) {
             $this->assertEquals(engine::STATUS_INITIALISED, $enginestep->status);
         }
+        $engine->set_status(engine::STATUS_FINISHED);
+        $engine->finalise();
+        ob_get_clean();
```
keevan, line 148: Was there meant to be an assertion here that's missing?

classes/task/process_dataflow_ad_hoc.php
```diff
@@ -0,0 +1,46 @@
...
+        $rec = $this->get_custom_data();
```
keevan, line 32: What is $rec? What's it meant to hold? 

Could you please rename this, and all occurrences into something more meaningful?

classes/local/execution/engine.php
```diff
@@ -165,21 +170,36 @@ public function __construct(dataflow $dataflow, bool $isdryrun = false, $automat
...
-            $this->status_check(self::STATUS_NEW);
+        $timeout = 0;
+        if (is_null(self::$lockfactory)) {
+            self::$lockfactory = \core\lock\lock_config::get_lock_factory('tool_dataflows_engine');
+        }
 
-            foreach ($this->enginesteps as $enginestep) {
-                $enginestep->initialise();
-            }
+        if ($this->lock = self::$lockfactory->get_lock('tool_dataflows_engine_' . $this->dataflow->id, $timeout)) {
```
keevan, line 185: Similar to objectfs, I think this should be moved out to its own method, and perhaps you might not need the lockfactory check (as objectfs didn't).

The call could then be simplified to something like:
```
if ($this->get_lock()) { 
    ...
}
```

classes/local/execution/engine.php
```diff
@@ -300,6 +320,10 @@ public function finalise() {
             if (isset($this->run)) {
                 $this->run->finalise($this->status, $this->export());
             }
+            if (!is_null($this->lock)) {
+                $this->lock->release();
+                $this->lock = null;
```
keevan, line 325: It seems the `$this->lock = null;` setting might be redundant?

classes/local/execution/engine.php
```diff
@@ -323,6 +347,10 @@ public function abort(?\Throwable $reason = null) {
         }
         $this->set_status(self::STATUS_ABORTED);
         $this->log('Aborted: ' . $message);
+        if (!is_null($this->lock)) {
+            $this->lock->release();
+            $this->lock = null;
```
keevan, line 352: It seems the `$this->lock = null;` setting might be redundant?

classes/task/process_dataflows.php
```diff
@@ -42,20 +42,32 @@ public function get_name() {
...
+     * Run the dataflows.
      */
     public function execute() {
-        $dataflowids = scheduler::get_due_dataflows();
-        foreach ($dataflowids as $id => $nextruntime) {
+        $dataflowrecs = scheduler::get_due_dataflows();
+
+        $firstrec = array_shift($dataflowrecs);
+
+        if (!is_null($firstrec)) {
```
keevan, line 52: This can be an `isset` check instead, as it performs the same thing as is a language construct which is slightly faster.

classes/task/process_dataflows.php
```diff
@@ -42,20 +42,32 @@ public function get_name() {
...
     /**
-     * Processes dataflows.
+     * Run the dataflows.
      */
     public function execute() {
-        $dataflowids = scheduler::get_due_dataflows();
-        foreach ($dataflowids as $id => $nextruntime) {
+        $dataflowrecs = scheduler::get_due_dataflows();
+
+        $firstrec = array_shift($dataflowrecs);
```
keevan, line 50: Similar to the comment about $rec / $recs, can this be renamed to $dataflows (as that is what the expected return value from the function call is)

classes/local/execution/engine.php
```diff
@@ -300,6 +320,10 @@ public function finalise() {
             if (isset($this->run)) {
                 $this->run->finalise($this->status, $this->export());
             }
+            if (!is_null($this->lock)) {
```
keevan, line 323: This can be an `isset` check instead of an `!is_null`

classes/local/execution/engine.php
```diff
@@ -323,6 +347,10 @@ public function abort(?\Throwable $reason = null) {
         }
         $this->set_status(self::STATUS_ABORTED);
         $this->log('Aborted: ' . $message);
+        if (!is_null($this->lock)) {
```
keevan, line 350: This can be an `isset` check instead of an `!is_null`

classes/local/execution/engine.php
jaypha: This is to avoid it being released in the __destruct() function.

keevan: Do you think it woul dbe better to add the relevant checks then in the destruct function, instead of needing to call `= null` everywhere else?

keevan (commented):
It also seems like there is no test with regards to the spawning and handling of those adhoc tasks, in particular with the locking I saw during WIP. Was this something you still wanted to add at some stage? Or was the locking within the same process you've mentioned the other day causing a problem with testing - so it was excluded?

tests/local/execution/tool_dataflows_engine_test.php
jaypha: These extra statements are to ensure that the locks get released, and are not part of the test.

jaypha:
> It also seems like there is no test with regards to the spawning and handling of those adhoc tasks

I'm not sure how you would go about testing spawning tasks, but I'll see what I can come up with.

classes/local/step/reader_sql.php
```diff
@@ -53,6 +53,21 @@ public static function form_define_fields(): array {
         ];
     }
 
+    /**
+     * Can dataflows with this step be executed in parallel.
+     *
+     * @return bool
+     */
+    public function is_concurrency_allowed(): bool {
```
brendanheywood, line 61: Thinking out loud, if a step cannot allow concurrency then it would be useful to say why it can't be and show that on the form where the tick box is shown. And the reasons would be dynamic like in this case.

So perhaps this method should return either null or a lang string similar to form validation

classes/form/dataflow_form.php
```diff
@@ -49,6 +49,10 @@ public function definition() {
         $mform->addElement('advcheckbox', 'enabled', get_string('dataflow_enabled', 'tool_dataflows'));
         $mform->setDefault('enabled', 0);
 
+        // Enable/Disable control.
+        $mform->addElement('advcheckbox', 'concurrencyallowed', get_string('concurrency_allowed', 'tool_dataflows'));
+        $mform->setDefault('concurrencyallowed', 0);
+
```
brendanheywood, line 55: If concurrency isn't allowed because of step types when we should show this here under the checkbox and make it clear that the checkbox won't take effect.

classes/local/execution/engine.php
```diff
@@ -63,6 +63,8 @@ class engine {
     const STATUS_ABORTED = 8;
     /** @var int Step/dataflow has completely finished. */
     const STATUS_FINALISED = 9;
+    /** @var int Dataflow cannot run because of locking. */
+    const STATUS_LOCKED = 10;
```
brendanheywood, line 67: I would be very carefully having a state of locked, or rather avoid thinking conceptually of being locked as a state a workflow can be in. If a workflow is running in process A then its state is STATUS_PROCESSING or some other state. It is never actually in a state of being locked, it is merely a second process B who can't touch it and it is only locked from their point of view. So we need to be careful that process B can never persist the state and end up writing that its state is locked as that is impossible.

So maybe this should be it's own independent var, or maybe we have some protecting in the persistent to make sure this can never happen

classes/local/step/trigger_cron.php
```diff
@@ -237,14 +237,17 @@ public function on_delete() {
     }
 
     /**
-     * Hook function that gets called when an engine step has been finalised.
+     * Hook function that gets called when an engine step has been initialised.
      *
      * @throws \dml_exception
      */
-    public function on_finalise() {
+    public function on_initialise() {
```
brendanheywood, line 244: Was this change strictly needed for this patch?

I'm thinking if you have a cron task which is supposed to run every hour on the hour, like an import and it is being synchronized with some external process running on the same schedule, then generally we want the next run time to be the next clean hour boundary. If we calculate this in advance at the start of the run rather than the end then we change this behavior.

So if each flow consistently takes 70 minutes then the flow executes every two hours and skips the ones where the hour started and the previous flow was still running. If you do it at the start then I think you'll end up with runs starting at very inconsistent times when they overlap.

brendanheywood: Actually I think the more nuanced behavior is to leverage the value of whether concurrency is allowed for this run, and if it is then calculate it at the start of the run, and if it is not then calculate it at the end of the run.

classes/dataflow.php
```diff
@@ -386,15 +389,32 @@ public function validate_steps() {
...
             }
+            $concurrencyallowed &= $steptype->is_concurrency_allowed();
         }
 
         if ($numtriggers > 1) {
             $errors['toomanytriggers'] = get_string('toomanytriggers', 'tool_dataflows');
         }
 
+        if (!$concurrencyallowed && $this->concurrencyallowed === '1') {
+            $errors['concurrency_not_allowed'] = get_string('concurrency_not_allowed', 'tool_dataflows');
```
brendanheywood, line 400: I think instead of treating this like an error, treat it more like the form field is not honored. Because you could set this, then go change the flow like touch the sql reader, then come back here and now it is invalid

classes/local/execution/engine.php
```diff
@@ -168,9 +177,38 @@ public function __construct(dataflow $dataflow, bool $isdryrun = false, $automat
...
+
+    /**
+     * Tries to obtain a lock for this dataflow
+     *
+     * @returns \core\lock\lock|false.
+     */
+    public function get_lock($timeout = 0) {
+        $lockfactory = \core\lock\lock_config::get_lock_factory('tool_dataflows_engine');
+        $this->lock = $lockfactory->get_lock('tool_dataflows_engine_' . $this->dataflow->id, $timeout);
+        return $this->lock;
```
brendanheywood, line 197: One nice thing is when you lock a workflow we stash some metadata somewhere around who and what is running it and then we can display that when it is locked eg 'This workflow is already being run by cron started at 12:45 pid 12345' or 'This workflow is already being run by Brendan Heywood started at 12:45'

jaypha: Can you tell me where the metadata is stashed?

brendanheywood: This is up to you. I think the place which makes the most sense is a new column in tool_dataflows_runs but double check that is valid for when a run fails and we re-run it

keevan:
Some phpdoc issues found:
```
admin/tool/dataflows/classes/local/execution/engine.php
    Line 187: Phpdocs for function engine::get_lock has incomplete parameters list
    Line 190: Invalid phpdocs tag @returns used
```

classes/local/step/trigger_cron.php
```diff
@@ -236,15 +236,34 @@ public function on_delete() {
...
+    }
+
+    /**
+     * Reschedules the dataflow.
+     */
+    protected function reschedule() {
         if (!$this->enginestep->engine->isdryrun) {
             $dataflowid = $this->stepdef->dataflowid;
 
+            // Reschedule on initialisation, so that it can run on the next cron schedule, even if this
```
brendanheywood, line 264: This comment should probably be moved up to on_initialise and a similar comment for on_finalise

lang/en/tool_dataflows.php
```diff
@@ -207,6 +208,10 @@
 $string['path_invalid'] = 'Path "{$a}" is not permitted.';
 $string['path_not_absolute'] = 'Path "{$a}" is not an absolute path.';
 $string['unknown_placeholder'] = 'Placeholder "{$a}" is not recognised.';
+$string['concurrency_not_allowed'] = 'Dataflow includes steps with disallow concurrency. Dataflow concurrency setting should be off.';
```
brendanheywood, line 211: steps *which* disallow

jaypha: The string is not actually used, so I removed it.

classes/local/step/trigger_cron.php
jaypha: done

classes/local/execution/engine.php
jaypha: Even though I put a release in the destruct function as a precaution, I still needed to add it elsewhere because Moodle checks the locks before the object gets destroyed.

classes/task/process_dataflows.php
jaypha: done

classes/local/execution/engine.php
jaypha: done

PR #291, opened 2022-07-13
fix: restrict generation of the dot graph to when site has set pathtodot

### LINKED ISSUE #290: PHPUnit tests not running to completion due to missing dot executable
Tasks:
- [x] <s>Skip the tests relating to dot usage if the dot cli is not found</s> Skipping this one because handling in core is the same
- [x] See if there's a simple way to add dot to the CLI runner (GH actions)

Background:

https://github.com/catalyst/moodle-tool_dataflows/runs/7313125448?check_suite_focus=true#step:3:1975
```
......................FFfailed to execute cmd: "dot -Tsvg". stderr: `sh: 1: dot: not found
```

In particular I believe the absence of the dot executable is causing the plugin specific tests to silently fail, and so other potential issues aren't discovered / noticed.

```
Moodle 4.1dev (Build: 20220707), 1802cd526054cc52c3089c10b0dadfcdd5b6891e
Php: 7.4.28, pgsql: 14.0 (Debian 14.0-1.pgdg110+1), OS: Linux 5.15.0-40-generic x86_64
PHPUnit 9.5.13 by Sebastian Bergmann and contributors.

..........................FFF......................               51 / 51 (100%)

Time: 00:07.717, Memory: 328.00 MB

There were 4 failures:

1) tool_dataflows\tool_dataflows_curl_connector_test::test_execute
Failed asserting that two strings are equal.
[output omitted]

2) tool_dataflows\tool_dataflows_secret_service_test::test_secret_redaction_in_logs_and_during_run
[output omitted]

3) tool_dataflows\tool_dataflows_secret_service_test::test_secret_redaction_for_exports
Failed asserting that 'OHNOYOUWERENTSUPPOSEDTOSEEME' is not equal to 'OHNOYOUWERENTSUPPOSEDTOSEEME'.

4) tool_dataflows\tool_dataflows_secret_service_test::test_secret_redaction_for_table_display
[output omitted]

keevan:
Initially added a way to skip dot graph generation if the path is not set, but there wasn't a quick way to include the necessary config values, which would have resulted in a fixed added option in reusable workflows which isn't great. 

For now, adding the dot CLI will cause the unit tests to fail successfully

Resolves #290

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,4 +12,6 @@ jobs:
     with:
       disable_behat: true
       disable_grunt: true
-      extra_plugin_runners: 'moodle-plugin-ci add-plugin catalyst/moodle-local_aws --branch ci-for-tool_dataflows'
+      extra_plugin_runners: |
+          moodle-plugin-ci add-plugin catalyst/moodle-local_aws --branch ci-for-tool_dataflows;
+          sudo apt-get install -y graphviz;


jaypha:
This does not pick up all the unit test failures. Bu it does address the dot problem, so I'm OK to merge.

PR #294, opened 2022-07-13
Issue #289: Fix failing unit tests.

### LINKED ISSUE #289: Unit tests failing on local machine, despite passing on Github CI.
This might be the cause.
https://github.com/catalyst/moodle-tool_dataflows/runs/7313125448?check_suite_focus=true#step:3:1975

jaypha:
resolves #289 

Removed '\n' from expected value in curl_connector_test.
Swapped to values so the expected value is passed first.

Changed step::get_config() to get steptype from get_steptype. This fixed
weird error where isset($this->type) was returning false even though it has a value.

### DIFF
--- a/classes/step.php
+++ b/classes/step.php
@@ -171,9 +171,10 @@ class step extends persistent {
      * expressions are evaluated at this point in time.
      *
      * @param   bool $expressions whether or not to parse expressions when returning the config
+     * @param   bool $redacted Will redact secret values if true
      * @return  \stdClass configuration object
      */
-    protected function get_config($expressions = true, $redacted = false): \stdClass {
+    protected function get_config($expressions = true, bool $redacted = false): \stdClass {
         $rawconfig = $this->raw_get('config');
         $yaml = $rawconfig;
         if (gettype($rawconfig) === 'string') {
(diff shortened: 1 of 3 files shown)


classes/step.php
```diff
@@ -184,10 +185,11 @@ protected function get_config($expressions = true, $redacted = false): \stdClass
             return new \stdClass();
         }
 
+        $steptype = $this->get_steptype();
+
         // Ensure the secret service gets involved to redact any information required.
-        if ($redacted && isset($this->type)) {
+        if ($redacted && isset($steptype)) {
```
keevan, line 191: I think this is one of the weird quirks of magic methods. Doing what you've done is fine, and doing this is also fine and will work. 
```
        $steptype = $this->type;
        ...
        if ($redacted && isset($steptype)) {
            ...        
        }
```
It's only an issue if you use a magic value directly in a conditional (at least from observation). It's a bit annoying and catch you off guard admittedly.

classes/step.php
```diff
@@ -171,9 +171,10 @@ public function set_dataflow(dataflow $dataflow) {
      * expressions are evaluated at this point in time.
      *
      * @param   bool $expressions whether or not to parse expressions when returning the config
+     * @param   bool $redacted Will redact secret values if true
      * @return  \stdClass configuration object
      */
-    protected function get_config($expressions = true, $redacted = false): \stdClass {
+    protected function get_config($expressions = true, bool $redacted = false): \stdClass {
```
keevan, line 177: Thanks for adding the typehints here :heart:

PR #295, opened 2022-07-13
279 fix incorrect return type

### LINKED ISSUE #279: Return type of scheduler::get_scheduled_times is incorrect
https://github.com/catalyst/moodle-tool_dataflows/blob/MOODLE_35_STABLE/classes/local/scheduler.php#L32-L42

Resulting in some false positives in code that uses it:

https://github.com/catalyst/moodle-tool_dataflows/blob/MOODLE_35_STABLE/classes/local/step/trigger_cron.php#L76

image

keevan:
- fix: return type of scheduler::get_scheduled_times
- docs: add missing param / typehints for phpdocs

Resolves #279

### DIFF
--- a/classes/local/step/trigger_cron.php
+++ b/classes/local/step/trigger_cron.php
@@ -18,8 +18,6 @@ namespace tool_dataflows\local\step;
 
 use tool_dataflows\step;
 use tool_dataflows\local\scheduler;
-use tool_dataflows\local\execution\engine_step;
-
 
 /**
  * CRON trigger class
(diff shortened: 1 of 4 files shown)


classes/local/step/trigger_cron.php
```diff
@@ -45,9 +48,10 @@ protected static function form_define_fields(): array {
     /**
      * Get the default data.
      *
-     * @return stdClass
+     * @param \stdClass $data from the persistent form class
+     * @return \stdClass
      */
-    public function form_get_default_data(&$data) {
+    public function form_get_default_data(\stdClass &$data): \stdClass {
```
petersistrom, line 54: Query: A return type is defined but the function doesn't have a return statement.

keevan: Yes good catch. This was setting the values directly because it was passed in as a reference, but I think this is hard to follow so I've updated the code to always pass it by value.

PR #299, opened 2022-07-14
fix: improved s3 handling of paths with more tests

### LINKED ISSUE #298: S3 copy issue with s3 paths starting with "S" or "3"
This is caused by incorrect trimming / truncating of the s3 prefix.

For example: a target of "s3://something/output.csv" will currently be translated to "omething/output.csv" instead of the expected "something/output.csv" path.

keevan:
- fixed s3 prefix removal handling
- fix directory nesting possibility for local files (e.g. path/to/file)
- add tests to cover the fixed issues
- some minor phpdoc cleanup, refactoring

Resolves #298

### DIFF
--- a/classes/local/execution/engine.php
+++ b/classes/local/execution/engine.php
@@ -141,6 +141,7 @@ class engine {
         foreach ($dataflow->steps as $stepdef) {
             $classname = $stepdef->type;
             $steptype = new $classname($stepdef, $this);
+            $stepdef->steptype = $steptype;
             $this->enginesteps[$stepdef->id] = $steptype->get_engine_step();
         }
 
@@ -496,10 +497,12 @@ class engine {
      * @return false|resource
      */
     public function resolve_path(string $pathname) {
+        global $CFG;
+
         $fullpath = helper::path_get_absolute($pathname, $this->scratchdir);
         $dir = dirname($fullpath);
         if (!file_exists($dir)) {
-            mkdir($dir);
+            mkdir($dir, $CFG->directorypermissions, true);
         }
         return $fullpath;
     }
--- a/classes/local/step/base_step.php
+++ b/classes/local/step/base_step.php
@@ -33,7 +33,8 @@ abstract class base_step {
 
     /** @var engine_step Engine step made by this class for an execution. */
     protected $enginestep = null;
-    /** @var step The step definition use to create the engien step.  */
+
+    /** @var step The step definition use to create the engine step.  */
     protected $stepdef = null;
 
     /**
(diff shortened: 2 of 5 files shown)


classes/local/execution/engine.php
```diff
@@ -496,10 +497,12 @@ protected function status_check(int $expected) {
      * @return false|resource
      */
     public function resolve_path(string $pathname) {
+        global $CFG;
+
         $fullpath = helper::path_get_absolute($pathname, $this->scratchdir);
         $dir = dirname($fullpath);
         if (!file_exists($dir)) {
-            mkdir($dir);
+            mkdir($dir, $CFG->directorypermissions, true);
```
keevan, line 505: This allows nested directory / file creation.

classes/local/execution/engine.php
```diff
@@ -141,6 +141,7 @@ public function __construct(dataflow $dataflow, bool $isdryrun = false, $automat
         foreach ($dataflow->steps as $stepdef) {
             $classname = $stepdef->type;
             $steptype = new $classname($stepdef, $this);
+            $stepdef->steptype = $steptype;
```
keevan, line 144: This ensures the step type can be referenced from the step def (persistent) instead of always returning a new unlinked step type (no engine)

classes/local/step/connector_s3.php
```diff
@@ -124,6 +124,7 @@ public function execute(): bool {
 
         // Do not execute s3 operations during a dry run.
         if ($this->enginestep->engine->isdryrun) {
+            $this->enginestep->log("Skipping copy to '{$target}' as this is a dry run.");
```
keevan, line 127: Just making it a bit clearer, from the run output. There wasn't a clear way to see if something was working / failing / or in this case, skipping.

classes/local/step/connector_s3.php
```diff
@@ -193,7 +194,8 @@ public function execute(): bool {
     public function resolve_path(string $path, bool $ins3): string {
         // S3 Path: when the path is in s3, trim the prefix.
         if ($ins3) {
-            return ltrim($path, self::S3_PREFIX);
+            // The path returned is the substring starting from where the prefix ends.
+            return substr($path, strlen(self::S3_PREFIX));
```
keevan, line 198: So with `ltrim`, the individual characters of the prefix were being removed from the left side of the path. The correct function to use was probably `substr`. The comment explain how it should work.

classes/step.php
```diff
@@ -144,14 +144,34 @@ public function get_dataflow(): dataflow {
...
     /**
      * Returns the step type associated with this step
+     *
+     * @return base_step|null a step type derived from the base step type, or null if the type configured was invalid.
      */
     public function get_steptype() {
+        // This would generally be initiliased by the engine.
+        if (property_exists($this, 'steptype') && isset($this->steptype)) {
+            return $this->steptype;
+        }
```
keevan, line 154: Supporting the changes made in the engine earlier, this ensures if a value (step type) was set, it would be returned instead of initializing a new one every time.

classes/local/step/base_step.php
```diff
@@ -70,12 +71,23 @@ abstract class base_step {
...
+        if (isset($engine)) {
+            $this->set_engine($engine, $stepdef);
+        }
+    }
+
+    /**
+     * Links the engine to this step type, which requires a step def to be provided as well.
+     *
+     * @param   engine $engine
+     * @param   step $stepdef
```
petersistrom, line 84: Seems to be missing a @throws tag in the phpdocs

Issue #260, opened 2022-07-04
Refactor to allow vars to be nested eg curl response


brendanheywood:
Lets pick a step such as the curl connector. After it runs it will dump a whole bunch of vars available for reading, but rather than a giant list of keys it makes sense to nest some of these together, which is also how they are created in code anyway.

So we want something vaguely like this as potential vars we could read afterwards:

```
curl.response.header.server <--
curl.response.header.vary <-- these keys are arbitrary based on what comes back
curl.response.header...etc 
curl.response.body
curl.response.size
curl.response.timing.connect
curl.response.timing.wait
```

From an api point of view, I could see this step either calling set_var with curl.response.header and a hash, and then calling is with curl.response.body with a string, etc OR it could build the whole response in one thing and call set_var once and they are exactly equivalent

See also #245

PR #300, opened 2022-07-14
webservice step 13

### LINKED ISSUE #13: Processor call internal web service for each row
Each argument to the webservice would be an expression

* [x] config for which web service to call
* [x] config for which user to pseudo as for the call, defaults to the admin / cli user
* [x] expression for the data pushed into the call
* [x] config for what to do if it fails, either record it and move on or abort the entire flow?

marcghaly:
Hi @keevan,

I am aware there is still an issue with the state but have not been able to troubleshoot it, maybe another pair of eyes can be helpful.

Resolves #13

Regards,

Marc

### DIFF
--- a/classes/local/step/flow_web_service.php
+++ b/classes/local/step/flow_web_service.php
@@ -1,0 +1,221 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_dataflows\local\step;
+
+use external_api;
+use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
+use Symfony\Component\Yaml\Yaml;
+use tool_dataflows\parser;
+
+/**
+ * Flow web service
+ *
+ * @package    tool_dataflows
+ * @author     Ghaly Marc-Alexandre <marc-alexandreghaly@catalyst-ca.net>
+ * @copyright  Catalyst IT, 2022
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class flow_web_service extends flow_logic_step {
+
+    /** @var bool whether or not this step type (potentially) has side effects, this will vary depending on Web Service */
+    protected $hassideeffect = true;
+
+    /**
+     * Return the definition of the fields available in this form.
+     *
+     * @return array
+     */
+    public static function form_define_fields(): array {
+        return [
+            'webservice' => ['type' => PARAM_RAW_TRIMMED, 'required' => true],
+            'user' => ['type' => PARAM_TEXT, 'required' => true],
+            'parameters' => ['type' => PARAM_RAW],
+            'failure' => ['type' => PARAM_TEXT, 'required' => true],
+            'path' => ['type' => PARAM_PATH],
+        ];
+    }
+
+    /**
+     * Allows each step type to determine a list of optional/required form
+     * inputs for their configuration
+     *
+     * It's recommended you prefix the additional config related fields to avoid
+     * conflicts with any existing fields.
+     *
+     * @param \MoodleQuickForm $mform
+     */
+    public function form_add_custom_inputs(\MoodleQuickForm &$mform) {
+        $options = [
+            'multiple' => false,
+            'noselectionstring' => get_string('flow_web_service:selectuser', 'tool_dataflows'),
+            'placeholder' => get_string('flow_web_service:selectuser', 'tool_dataflows'),
+        ];
+
+        $jsonexample = [
+            'users' => [
+                [
+                    'username' => 'john1234',
+                    'password' => 'Pwdtest123*&12',
+                    'firstname' => 'john',
+                    'lastname' => 'doe',
+                    'email' => 'john@doe.ca',
+                ],
+            ],
+        ];
+        $class = ['class' => 'badge badge-dark rounded-0'];
+        $outputexample['somekey'] = \html_writer::nonempty_tag('span', 'steps.' . ($this->stepdef->alias ?? 'alias') . '.somekey',
+            $class);
+        $outputexample['id'] = \html_writer::nonempty_tag('span', 'steps.' . ($this->stepdef->alias ?? 'alias') . '.id', $class);
+        $outputexample['username'] = \html_writer::nonempty_tag('span', 'steps.' . ($this->stepdef->alias ?? 'alias') . '.username',
+            $class);
+        $examples['yaml'] = \html_writer::nonempty_tag('pre', Yaml::dump($jsonexample, 3));
+
+        $mform->addElement('text', 'config_webservice', get_string('flow_web_service:webservice', 'tool_dataflows'));
+        $mform->addElement('static', 'config_webservice_help', '',
+            get_string('flow_web_service:webservice_help', 'tool_dataflows'));
+        $mform->addElement('text', 'config_user', get_string('flow_web_service:user', 'tool_dataflows'));
+        $mform->addHelpButton('config_user', 'flow_web_service:user', 'tool_dataflows');
+        $mform->addElement('textarea', 'config_parameters', get_string('flow_web_service:parameters', 'tool_dataflows'),
+            ['cols' => 50, 'rows' => 7]);
+        $mform->addHelpButton('config_parameters', 'flow_web_service:parameters', 'tool_dataflows');
+        $mform->addElement('static', 'parameters_help', '',
+            get_string('flow_web_service:field_parameters_help', 'tool_dataflows', $examples));
+        $mform->addElement('select', 'config_failure', get_string('flow_web_service:failure', 'tool_dataflows'),
+            [
+                'abortstep' => get_string('flow_web_service:abortstep', 'tool_dataflows'),
+                'abortflow' => get_string('flow_web_service:abortflow', 'tool_dataflows'),
+            ]
+        );
+        $mform->addHelpButton('config_failure', 'flow_web_service:failure', 'tool_dataflows');
+        $mform->addElement('text' , 'config_path', get_string('flow_web_service:path', 'tool_dataflows'));
+        $mform->hideIf('config_path', 'config_failure', 'neq', 'record');
+        $mform->disabledIf('config_path', 'config_failure', 'neq', 'record');
+    }
+
+    /**
+     * Validate the configuration settings.
+     *
+     * @param object $config
+     * @return true|\lang_string[] true if valid, an array of errors otherwise
+     */
+    public function validate_config($config) {
+        $errors = [];
+        if (empty($config->webservice)) {
+            $errors['config_webservice'] = get_string('config_field_missing', 'tool_dataflows', 'webservice', true);
+        }
+        if (empty($config->user)) {
+            $errors['config_user'] = get_string('config_field_missing', 'tool_dataflows', 'user', true);
+        }
+        if ($config->user) {
+            // Gets user count and checks whether deleted and confirmed.
+            $user = get_users(false, '', true, null, 'firstname ASC', '', '', '', '', '*', 'username = :username',
+                ['username' => $config->user]);
+            if (!$user) {
+                $errors['config_user'] = get_string('config_user_invalid', 'tool_dataflows', $config->user, true);
+            }
+        }
+        if ($config->failure === 'record' && empty($config->path)) {
+            $errors['config_path'] = get_string('config_field_missing', 'tool_dataflows', 'path', true);
+        }
+        return empty($errors) ? true : $errors;
+    }
+
+    /**
+     * Executes the step
+     *
+     * Performs web service call.
+     *
+     * @param object|mixed $input
+     * @return bool Returns true if successful, false otherwise.
+     */
+    public function execute($input) {
+        global $DB, $SESSION, $USER, $OUTPUT;
+        // Store the previous user and session, setting it back once the step is finished.
+        $previoususer = $USER;
+        $session = $SESSION;
+        $currentoutput = $OUTPUT;
+        $outertransaction = $DB->is_transaction_started();
+
+        $config = $this->enginestep->stepdef->config;
+        $isdryrun = $this->enginestep->engine->isdryrun;
+        $userid = $DB->get_field('user', 'id', ['username' => $config->user]);
+        $user = \core_user::get_user($userid);
+        $user->ignoresesskey = true;
+        \core\session\manager::init_empty_session();
+        \core\session\manager::set_user($user);
+        set_login_session_preferences();
+
+        // Fake it till you make it - set the the lastaccess in advance to avoid
+        // this value being updated in the database via user_accesstime_log() as
+        // we are not actually logging in and accessing the site as this user.
+        $USER->lastaccess = time();
+        $params = json_decode($config->parameters, true);
+
+        if (is_null($params)) {
+            $params = Yaml::parse($config->parameters);
+        }
+
+        // Updates parameters passed to WS.
+        $parser = new parser();
+        // Check for any data in input to replace.
+        $replacement = $parser->evaluate_recursive($params, ['data' => $input]);
+        // Ensure structures are arrays for webservice.
+        $replacement = json_decode(json_encode($replacement), true);
+        $params = array_replace_recursive($params, $replacement);
+        // If password is need will throw exception.
+        array_walk_recursive($params, function (&$item) {
+            $item = strtolower($item);
+        });
+        $failure = !isset($config->failure) ? 'abortflow' : $config->failure;
+        $path = $config->path ?? null;
+
+        if ($path) {
+            if ($path[0] === '/') {
+                $path = ltrim($path, '/');
+            }
+        }
+
+        if (external_api::external_function_info($config->webservice)->type === 'read') {
+            $this->hassideeffect = false;
+        };
+
+        if (!$isdryrun) {
+            $response = external_api::call_external_function($config->webservice, $params);
+            // Restore the previous user to avoid any side-effects occuring in later steps / code.
+            // Avoid moodle state errors because of webservice call - we are still in body.
+            \core\session\manager::set_user($previoususer);
+            $SESSION = $session;
+            $OUTPUT = $currentoutput;
+            if ($response['error']) {
+                if (!$outertransaction && $DB->is_transaction_started()) {
+                    $DB->force_transaction_rollback();
+                }
+                if ($failure === 'abortflow') {
+                    throw new \moodle_exception($response['exception']->debuginfo ?? $response['exception']->message);
+                }
+                if ($failure === 'abortstep') {
+                    $this->enginestep->log($response['exception']->debuginfo ?? $response['exception']->message);
+                    return false;
+                }
+            }
+            // Success - store any desired value for subsequent steps.
+            // Output will then evaluate result.
+            $this->set_variables('result', $response['data']);
+        }
+        return true;
+    }
+}
--- a/lib.php
+++ b/lib.php
@@ -49,18 +49,19 @@ function tool_dataflows_after_config() {
 function tool_dataflows_step_types() {
     return [
         new step\connector_curl,
-        new step\trigger_cron,
-        new step\reader_sql,
-        new step\reader_json,
-        new step\writer_debugging,
-        new step\flow_logic_join,
-        new step\flow_logic_case,
-        new step\writer_stream,
         new step\connector_debugging,
         new step\connector_debug_file_display,
-        new step\connector_s3,
         new step\connector_sns_notify,
+        new step\connector_s3,
         new step\connector_wait,
+        new step\flow_logic_case,
+        new step\flow_logic_join,
+        new step\flow_web_service,
+        new step\reader_json,
+        new step\reader_sql,
+        new step\trigger_cron,
+        new step\writer_debugging,
+        new step\writer_stream,
     ];
 }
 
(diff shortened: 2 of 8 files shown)


marcghaly:
Managed to trouble shoot state issue,

Regards,

Marc

classes/local/step/flow_web_service.php
```diff
@@ -0,0 +1,281 @@
...
+/**
+ * Flow web service
+ *
+ * @package    tool_dataflows
+ * @author     Ghaly Marc-Alexandre <marc-alexandreghaly@catalyst-ca.net>
+ * @copyright  Catalyst IT, 2022
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class flow_web_service extends flow_logic_step {
+    /**
```
keevan, line 32: Can you please add a blank line above this function's phpdoc

classes/local/step/flow_web_service.php
```diff
@@ -0,0 +1,281 @@
...
+                0 => [
+                    'username' => 'john1234',
+                    'password' => 'Pwdtest123*&12',
+                    'firstname' => 'john',
+                    'lastname' => 'doe',
+                    'email' => 'john@doe.ca',
+                ]
+            ]
+        ];
+        $ex['yaml'] = '<pre>users:
```
keevan, line 80: I'm not sure what this variable means just from looking at it.

May you consider changing this to something that is more meaningful / self explanatory and searchable?

classes/local/step/flow_web_service.php
```diff
@@ -0,0 +1,281 @@
...
+        0:
+          username: john1234
+          password: Pwdtest123*&12
+          firstname: john
+          lastname: doe
+          email: john@doe.ca
+        </pre>';
+        $ex['json'] = \html_writer::nonempty_tag('pre', json_encode($jsonexample, JSON_PRETTY_PRINT));
```
keevan, line 88: You don't need to recreate the example in various different formats. I suggest keeping it in a PHP array, and using the methods to encode it to YAML and JSON as required. Possibly just showing the preferred format as an example, but listing all the supported formats.

classes/local/step/flow_web_service.php
```diff
@@ -0,0 +1,281 @@
...
+            ]
+        ];
+        $ex['yaml'] = '<pre>users:
+        0:
+          username: john1234
+          password: Pwdtest123*&12
+          firstname: john
+          lastname: doe
+          email: john@doe.ca
+        </pre>';
```
keevan, line 87: Can you please use the `html_writer::tag` method here instead? 

https://docs.moodle.org/dev/html_writer#References

classes/local/step/flow_web_service.php
```diff
@@ -0,0 +1,281 @@
...
+     * @return array
+     */
+    protected static function form_define_fields(): array {
+        return [
+            'webservice' => ['type' => PARAM_RAW_TRIMMED],
+            'user' => ['type' => PARAM_TEXT],
+            'parameters' => ['type' => PARAM_RAW],
+            'datastore' => ['type' => PARAM_RAW_TRIMMED],
+            'failure' => ['type' => PARAM_TEXT],
+            'path' => ['type' => PARAM_PATH],
```
keevan, line 44: The fields now support a new type (bool) called 'required', so adding this should automatically apply a form rule to the relevant field. This should currently be used in addition to the validate_config method, because of the different _forms_ of validation. 

And would have otherwise been asked to be included in the custom form inputs below if not already.

Ref:
https://docs.moodle.org/dev/lib/formslib.php_Form_Definition#addRule

classes/local/step/flow_web_service.php
```diff
@@ -0,0 +1,281 @@
...
+                    $res = $expressionlanguage->evaluate($value, ['data' => $input]);
+                    $arr[$key] = strtolower($res);
+                }
+                if ($key == $lastkey) {
+                    $return = $arr;
+                }
+            }
+        }
+        return $return;
+    }
```
keevan, line 162: If my understanding is correct, this does the same thing the parser would, but this supports an array coming in, and a processed array coming back out (and some hard rules for the data field)

I think this should be moved into the parser class, and it should use the evaluation function of the parser class instead of the raw expression language class for the following reasons:
- Keeps usage of expression variables the same throughout the dataflows plugin
- It will actually support setting webservice fields to something that can be referenced in an expression. I believe currently you would be bound to whatever $data you pass through.
- Keeps how it's used centralized in one place, which helps with maintainability.

This will mean you'll need to update usage to pass in the dataflow variables as well as the $input fields under e.g. a data key.

And this will also mean the configuration you've set currently should probably be `${{ data.FIRST_NAME }}` instead of `data.FIRST_NAME`

classes/local/step/flow_web_service.php
```diff
@@ -0,0 +1,281 @@
...
+                if ($failure === 'abortstep') {
+                    $this->enginestep->log($response['exception']->debuginfo);
+                    return false;
+                }
+                if ($failure === 'record') {
+                    $destination = $this->enginestep->engine->resolve_path($path);
+                    file_put_contents($destination, $response['exception']);
+                    return false;
+                }
+            } else {
```
keevan, line 257: Unnecessary else / nested block. 

Reason: The if guard used above for `$response['error']` is expected to throw / return / halt execution. So anything after that should be expected to be working and normal. 

https://github.com/jupeter/clean-code-php#avoid-nesting-too-deeply-and-return-early-part-1

classes/local/step/flow_web_service.php
```diff
@@ -0,0 +1,281 @@
...
+                // Success - store any desired value for subsequent steps.
+                if (!empty($config->datastore)) {
+                    $data = explode(',', $config->datastore);
```
keevan, line 260: This variable name is a bit generic. Including usage of $data and $val below it's a bit hard to know what's going on.

May you consider changing this to something that is more meaningful / self explanatory and searchable?

classes/local/step/flow_web_service.php
```diff
@@ -0,0 +1,281 @@
...
+                    foreach ($data as $val) {
+                        $value = $this->array_search_key($response['data'], $val);
+                        $value = $value[$val];
+                        $val = $this->stepdef->alias . '_' . $val;
+                        $input->{$val} = $value;
+                    }
```
keevan, line 266: With regards to the usage here, I'm somewhat on the fence about how it's being used currently. Is it not possible to reference what you want directly? And how you want it passed on directly as well? I think this can be more generic, and could be applied across steps as needed.

It's somewhat related to this: https://github.com/catalyst/moodle-tool_dataflows/issues/260 but the idea I had in mind at least for this step initially is something like, the configuration can be in yaml format
```
userid: newuser.id
username: newuser.username
```

This is based on the assumption the result of the web service call being nested in a field called newuser, the userid and username fields would then be merged in with the $input value so the data would be "transformed" by this web service step. 

Perhaps that is one way to do it? I'm wary of value conflicts and setting aliases so this seems like a way to resolve that issue.

keevan (commented):
You will need to bump the version.php because:
- lang strings have been updated

I've done a quick code pass, but will test functionality soon.

keevan:
Could you also please add some unit tests

keevan:
Could you also add a "Fixes #13" or "Resolves #13" to the PR description. This links the PR to the issue and automatically closes it after this PR merges.

It also shows as a linked pull request from the project board. Example below:

image

classes/local/step/flow_web_service.php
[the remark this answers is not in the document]
marcghaly: Added.

tests/tool_dataflows_web_service_flow_test.php
```diff
@@ -0,0 +1,218 @@
...
+/**
+ * Unit test for the web service flow step.
+ *
+ * @package   tool_dataflows
+ * @author    Ghaly Marc-Alexandre <marc-alexandreghaly@catalyst-ca.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class tool_dataflows_web_service_flow_test extends \advanced_testcase {
+    /**
```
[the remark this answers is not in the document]
marcghaly: Added.

classes/local/step/flow_web_service.php
```diff
@@ -0,0 +1,281 @@
...
+        $jsonexample = [
+            'users' => [
+                0 => [
+                    'username' => 'john1234',
+                    'password' => 'Pwdtest123*&12',
+                    'firstname' => 'john',
+                    'lastname' => 'doe',
+                    'email' => 'john@doe.ca',
+                ]
+            ]
```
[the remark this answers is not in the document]
marcghaly: Added.

tests/tool_dataflows_web_service_flow_test.php
```diff
@@ -0,0 +1,218 @@
...
+        $dataflow->enabled = true;
+        $dataflow->save();
+
+        $curlstep = new step();
+        // Curl step property setup.
+        $curlstep->config = Yaml::dump([
+            'curl' => $this->getExternalTestFileUrl('/h5puuid.json'),
+            'destination' => 'test.html',
+            'headers' => '',
+            'method' => 'get'
```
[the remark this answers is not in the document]
marcghaly: Added.

tests/tool_dataflows_web_service_flow_test.php
```diff
@@ -0,0 +1,218 @@
...
+        $jsonexample = [
+            'users' => [
+                0 => [
+                    'username' => 'john1234',
+                    'password' => 'Pwdtest123*&12',
+                    'firstname' => 'john',
+                    'lastname' => 'doe',
+                    'email' => 'john@doe.ca',
+                ]
+            ]
```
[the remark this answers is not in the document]
marcghaly: Added.

tests/tool_dataflows_web_service_flow_test.php
```diff
@@ -0,0 +1,218 @@
...
+        $newdataflow->name = 'flow-webservice-step';
+        $newdataflow->enabled = true;
+        $newdataflow->save();
+
+        // Re-add curl step.
+        $curlstep->config = Yaml::dump([
+            'curl' => $this->getExternalTestFileUrl('/h5pcontenttypes.json'),
+            'destination' => 'test1.html',
+            'headers' => '',
+            'method' => 'get'
```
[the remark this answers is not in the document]
marcghaly: Added.

tests/tool_dataflows_web_service_flow_test.php
```diff
@@ -0,0 +1,218 @@
...
+        $jsonexample = [
+            'users' => [
+                0 => [
+                    'username' => 'john1234567',
+                    'password' => 'Pwdtest123*&12',
+                    'firstname' => 'data.owner',
+                    'lastname' => 'doe',
+                    'email' => 'john@dodoe.ca',
+                ]
+            ]
```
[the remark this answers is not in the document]
marcghaly: Added

classes/local/step/flow_web_service.php
```diff
@@ -0,0 +1,281 @@
...
+        $mform->addElement('autocomplete', 'config_user',
+            get_string('flow_web_service:user', 'tool_dataflows'), $visibleusers, $options);
+        $mform->addElement('textarea', 'config_parameters', get_string('flow_web_service:parameters', 'tool_dataflows'),
+            ['cols' => 50, 'rows' => 7]);
+        $mform->addHelpButton('config_parameters', 'flow_web_service:parameters', 'tool_dataflows');
+        $mform->addElement('static', 'parameters_help', '',
+            get_string('flow_web_service:field_parameters_help', 'tool_dataflows', $ex));
+        $mform->addElement('text', 'config_datastore', get_string('flow_web_service:datastore', 'tool_dataflows'));
+        $mform->addHelpButton('config_datastore', 'flow_web_service:datastore', 'tool_dataflows');
+        $mform->addElement('static', 'datastore_example', '', get_string('flow_web_service:datastore_example', 'tool_dataflows', $this->stepdef->alias));
```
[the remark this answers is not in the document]
marcghaly: fixed

classes/local/step/flow_web_service.php
marcghaly: Changed $examples

classes/local/step/flow_web_service.php
marcghaly: Fixed.

classes/local/step/flow_web_service.php
marcghaly: Fixed.

classes/local/step/flow_web_service.php
marcghaly: Fixed.

classes/local/step/flow_web_service.php
marcghaly: Fixed. Thank you

classes/local/step/flow_web_service.php
marcghaly: Fixed.

classes/local/step/flow_web_service.php
marcghaly: This issue I ran through here is that Web services such as `core_user_create_users` and `core_course_get_courses_by_field` have responses that have `id` keys, by adding the alias it solves the conflicting values overriding each other. Values are being added to the input so they are reused by subsequent steps.

classes/local/step/flow_web_service.php
marcghaly: I have made changes pertaining to the parse in a separate commit, I m opening another PR that should be merged before with this, if all other changed are ok I ll drop the other commit.

marcghaly:
Hi @keevan,

I have used the outputs and made a few changes to parser / base_step, let me know if the use is incorrect,
Made appropriate changes to unit tests as well.

Regards,

Marc

marcghaly:
Hi @keevan, I have rebased over #309  and squashed those commits, if all is good to you I will rebase this branch over main branch once #309 is merged.

Regards,

Marc

brendanheywood:
@marcghaly I tried to resolve the merge conflicts in the github ui and it made a bit of a mess, can you please rebase this and push a clean patch up

Once more thing for in future (no need to retro do it here), can you make branch names which are human readable, eg if the issue was 13-webservice-step this makes it much easier when jumping around a lot of branches and cleaning them up later

classes/local/step/flow_web_service.php
```diff
@@ -0,0 +1,229 @@
...
+     * Allows each step type to determine a list of optional/required form
+     * inputs for their configuration
+     *
+     * It's recommended you prefix the additional config related fields to avoid
+     * conflicts with any existing fields.
+     *
+     * @param \MoodleQuickForm &$mform
+     */
+    public function form_add_custom_inputs(\MoodleQuickForm &$mform) {
+        $users = get_users(true, '', true, null, 'firstname ASC', '', '', '', 25, 'id, firstname, lastname, email');
```
brendanheywood, line 60: I don't think this can work, most sites will have many thousands of users.

So can you make it so that if it is left blank then it looks up the primary user. If it is specified then it looks up that user and I think it can just be a plain text field an accept the username of the user. This should validate when saving. Make the help text and placeholder text clear what is going on

classes/local/step/flow_web_service.php
```diff
@@ -0,0 +1,229 @@
...
+            // Success - store any desired value for subsequent steps.
+            // Output will then evaluate result.
+            $this->set_variables('result', $response['data']);
+
+            // Restore the previous user to avoid any side-effects occuring in later steps / code.
+            \core\session\manager::set_user($previoususer);
+            $SESSION = $session;
+            $OUTPUT = $currentoutput;
+        } else {
+            // Restore the previous user to avoid any side-effects occuring in later steps / code.
```
brendanheywood, line 222: This chunk is repeated 3 times, can you reorder things and refactor this into one?

classes/local/step/flow_web_service.php
```diff
@@ -0,0 +1,229 @@
...
+                if ($failure === 'abortflow') {
+                    throw new \moodle_exception($response['exception']->debuginfo ?? $response['exception']->message);
+                }
+                if ($failure === 'abortstep') {
+                    $this->enginestep->log($response['exception']->debuginfo ?? $response['exception']->message);
+                    return false;
+                }
+                if ($failure === 'record') {
+                    $destination = $this->enginestep->engine->resolve_path($path);
+                    file_put_contents($destination, $response['exception'] ?? $response['exception']->message);
```
brendanheywood, line 209: If there are multiple errors for different rows this will overwrite it each time, if we want this to be in a file then we should append it. 

But I'm not even sure putting it into a file is the most useful though as it isn't easily accessible for a site admin to view and do anything with. I think for MVP lets just log it into the flow logs, we can come back to this in another issue when with some real world use cases

classes/local/step/flow_web_service.php
```diff
@@ -0,0 +1,229 @@
...
+
+            if ($response['error']) {
+                // Restore the previous user to avoid any side-effects occuring in later steps / code.
+                // Avoid moodle state errors because of webservice call - we are still in body.
+                \core\session\manager::set_user($previoususer);
+                $SESSION = $session;
+                $OUTPUT = $currentoutput;
+
+                // Throw an exception to be propagated for proper error capture.
+                if (!$outertransaction && $DB->is_transaction_started()) {
```
brendanheywood, line 197: Can you explain this bit in more detail around the transactions in the comment?

classes/local/step/flow_web_service.php
marcghaly: Done.

classes/local/step/flow_web_service.php
marcghaly: Fixed

classes/local/step/flow_web_service.php
marcghaly: This is just a layer of safety in case the Webservice needs to be writting and fails - just ensure it does not leave the db in an uncomplete state

marcghaly:
Hi @brendanheywood, I rebased this branch over latest main one, there one comment I am not sure to have understood concerning file put contents, do you want this to be dropped or just leave it as is and fix it later ?

Regards,

Marc

brendanheywood:
Right now is crunch time, so if BCPS don't explicitly need a feature then lets err on the side of dropping the feature and splitting it into a new issue for later. So can you please do that here, just remove the option, make an issue, and you can stash the wip in another branch to come back to

classes/local/step/flow_web_service.php
```diff
@@ -0,0 +1,220 @@
...
+        $errors = [];
+        if (empty($config->webservice)) {
+            $errors['config_webservice'] = get_string('config_field_missing', 'tool_dataflows', 'webservice', true);
+        }
+        if (empty($config->user)) {
+            $errors['config_user'] = get_string('config_field_missing', 'tool_dataflows', 'user', true);
+        }
+        if ($config->user) {
+            global $DB;
+            $userid = $DB->get_field('user', 'id', ['username' => $config->user]);
```
brendanheywood, line 123: Can you also check that they are not suspended and not deleted. Even better use get_users() as it does some of this for you

classes/local/step/flow_web_service.php
brendanheywood: I meant in the comment in code. But the bit I specifically don't understand is the handling for nested transactions

lib.php
```diff
@@ -61,6 +61,7 @@ function tool_dataflows_step_types() {
         new step\connector_s3,
         new step\connector_sns_notify,
         new step\connector_wait,
+        new step\flow_web_service,
```
brendanheywood, line 64: can you please keep this list sorted by alpha (and fix any others which are wrong)

classes/local/step/flow_web_service.php
marcghaly: Done.

lib.php
marcghaly: Fixed.

marcghaly:
Hi @brendanheywood, I have made requested changes and opened another issue for logging the output to a file #368. Concerning the comment around transaction, it was a forgotten one, I have removed it as it is not throwing at this place.

Regards,

Marc

keevan (changes requested):
Some small things:
```
-------------
PHPCS issues
-------------
classes/local/step/flow_web_service.php:122:146: warning - Line exceeds 132 characters; contains 146 characters

--------------------------
VERSION NEEDS TO BE BUMPED
--------------------------
- lang strings have been updated
-------------
PHPDOC issues
-------------
admin/tool/dataflows/classes/local/step/flow_web_service.php
    Line 56: Phpdocs for function flow_web_service::form_add_custom_inputs has incomplete parameters list
```

Unit test is also failing, most likely because the user field has changed from using the `id` of the user and now expects the `username` instead. I also encountered this issue and had to change the previously provided dataflow.

The help text for the parameters field also needs to be updated, in particular it is now expecting `${{ data.something }}` instead of simply `data.something`, which allows for both forms if required. See screenshot below:

image



Once the above things are fixed up I'm happy to merge this for now and catch any issues at a later stage, as functionally it seems to be working.

classes/local/step/flow_web_service.php
```diff
@@ -0,0 +1,213 @@
...
+ * @package    tool_dataflows
+ * @author     Ghaly Marc-Alexandre <marc-alexandreghaly@catalyst-ca.net>
+ * @copyright  Catalyst IT, 2022
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class flow_web_service extends flow_logic_step {
+
+    /**
+     * Return the definition of the fields available in this form.
+     *
```
brendanheywood, line 36: Also this step should declare that it has side effects. This is not as simple as just saying true, it should really leverage the web service definition to say whether or not it is a read ws or not.

eg core_auth_confirm_user is a write ws, but core_auth_is_minor is a read one:

https://github.com/moodle/moodle/blob/master/lib/db/services.php#L49-L57

If the absence of config this step should by default be considered a write step which has side effects.

marcghaly: Added that here https://github.com/catalyst/moodle-tool_dataflows/commit/66e6f6e453afaf7b5288bd4fc27388310659d48a#diff-aa35067838cac274f171b1a761e9e37cc4652fbd570662aa5e9c3dd7bdab2aa1R192, but unsure if that should be always false when dry run and if it should be in the has_side_effect method instead ?

marcghaly:
Hi @brendanheywood and @keevan, I have fixed issues above and rebased against main branch.

Regards,

Marc

keevan (approved):

PR #302, opened 2022-07-16
Method array_evaluate_recursive added to parser

marcghaly:
This is linked to #13 and should be merged before


keevan:
@marcghaly Just wanting to let you know I've moved a big block which was handling some aspect of the recursion out to the parser class, whilst adding the changes for #260 which will take over the changes here, so this PR isn't required anymore.

PR #305, opened 2022-07-18
260 add more flexible vars

### LINKED ISSUE #260: Refactor to allow vars to be nested eg curl response

keevan:
Resolves #260 

tl;dr:
- steps expose their variables
- you declare what you want from the step
- you can use that in steps that come after

What this doesn't do is modify the input as it flows through. The concern was with a step doing too many things, and if used in another step later on such as a json writer step, you don't necessarily want the input (json / php object flowing through) to be polluted with all these additional things you've picked up along the way.

I think there should be a some _transformer_ step which uses these output variables and will explicitly update the input as it flows through, when required.

### DIFF
--- a/classes/parser.php
+++ b/classes/parser.php
@@ -62,7 +62,7 @@ class parser {
             }
 
             // If the field is a nested object, recurse and continue evalulating.
-            if (is_object($string)) {
+            if (is_object($string) || is_array($string)) {
                 $string = $this->evaluate_recursive($string, $variables);
                 continue;
             }
@@ -82,6 +82,9 @@ class parser {
                 if ($hasexpression) {
                     $resolved = $this->evaluate($string, $variables);
                     if ($resolved !== $string) {
+                        if (is_array($yaml)) {
+                            $yaml = (object) $yaml;
+                        }
                         $yaml->{$key} = $resolved;
                         $resolvedvalues->{$key} = $resolved;
                         $max++;
--- a/classes/parser.php
+++ b/classes/parser.php
@@ -32,7 +32,74 @@ class parser {
 
     public function __construct() {
         $expressionlanguage = new ExpressionLanguage();
+
+        // Register the fromJSON helper function to convert a json string into a php object.
+        $expressionlanguage->register('fromJSON', function ($str) {
+            return sprintf('(is_string(%1$s) ? json_decode(%1$s) : %1$s)', $str);
+        }, function ($arguments, $str) {
+            if (!is_string($str)) {
+                return $str;
+            }
+            return json_decode($str);
+        });
         $this->expressionlanguage = $expressionlanguage;
+    }
+
+    /**
+     * Loops through the properties of the input yaml and returns a parsed object back
+     *
+     * @param   \stdClass $yaml
+     * @param   array $variables
+     * @return  \stdClass $yaml
+     */
+    public function evaluate_recursive($yaml, $variables) {
+        $resolvedvalues = new \stdClass;
+        $unresolvedvalues = new \stdClass;
+        foreach ($yaml as $key => &$string) {
+            // If the field was NULL, then it should be skipped.
+            if (is_null($string)) {
+                continue;
+            }
+
+            // If the field is a nested object, recurse and continue evalulating.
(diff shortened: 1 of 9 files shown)


classes/parser.php
```diff
@@ -32,9 +32,76 @@ class parser {
...
+
+            [$hasexpression] = $this->has_expression($string);
+            if (!$hasexpression) {
+                $resolvedvalues->{$key} = $string;
+            } else {
+                $unresolvedvalues->{$key} = $string;
+            }
+        }
+        // NOTE: Assumption that doing this does not corrupt the state in any way.
+        // $variables['steps']->{$this->alias}->config = $resolvedvalues;
```
keevan, line 78: NTS: Bit of a remnant from previous handling, since this was pulled directly out of step get_config handling. 

Will need to look into the unit test failing.

classes/parser.php
```diff
@@ -104,7 +171,14 @@ private function internal_evaluator(string $string, array $variables, ?callable
                     }
                     // Set the evalulated expression to the one that passed through the expression language.
                     if (isset($result)) {
-                        $evaluatedexpression = str_replace($match['expressionwrapper'], $result, $evaluatedexpression);
+                        if (is_string($result)) {
+                            // Normally, the result returned from the expression is a non-scalar value. Replace the occurance of
+                            // it in the original provided expression with the result.
+                            $evaluatedexpression = str_replace($match['expressionwrapper'], $result, $evaluatedexpression);
+                        } else if (count($matches) === 1) {
```
keevan, line 180: NTS: probably throw an exception if it doesn't match any of these cases. 

Reasons: 
- 1st case is when it's part of a string, which is fine, just replace the expression bit. 
- 2nd case is when there was only one expression which is more or less safe to assume the whole field should be replaced
- The alternative, is when there's a string expression, but the result is not a string which might mean it's an object/array, and it wouldn't really make sense to have it be something like `array + string`, so I think it's an invalid expression and case we don't want to support for now.

tests/tool_dataflows_curl_connector_test.php
```diff
@@ -71,10 +78,10 @@ public function test_execute() {
         $engine = new engine($dataflow, false, false);
         $engine->execute();
         ob_get_clean();
-        $variables = $engine->get_variables()['steps']->connector->config;
+        $variables = $engine->get_variables()['steps']->connector->outputs;
```
keevan, line 81: Noting that step values will be available for expression handling, in the outputs configuration. So you can map these values to any custom key you want. This also means the results from the step, e.g. a curl call's response, won't pollute the configuration object and "accidentally" store its values on top. It should always be set under this outputs object/array unless the intention is to update the "configuration" (like sql reader)

tests/tool_dataflows_variables_test.php
```diff
@@ -275,4 +276,59 @@ public function test_variables_set_at_different_scopes() {
...
+
+        // Tests get method.
+        $stepdef->config = Yaml::dump([
+            'curl' => $testgeturl,
+            'destination' => '',
+            'headers' => '',
+            'method' => 'get',
+            // Sets expressed values typically for values in scope with this step, so they can be accessed from other steps.
+            'outputs' => [
+                'customOutputKey' => '${{ fromJSON(response.result).contentTypes[0].id }}'
```
keevan, line 317: Note that fromJSON was added to support expressing a value that might be set as a string from a returning JSON API call which the curl step handles. Without it, it will just be a string that might need to be read from a JSON reader step before being useful.

PR #306, opened 2022-07-19
Issue #143: Store config version history

### LINKED ISSUE #143: Store a version history of workflows
The idea is that for any log of an instance we can see the configuration for that workflow at the point in time it was run.

Thinking the easiest way is when we run a workflow we check it's configuration, export it to yaml, and stash that in a versions table along with a hash of that configuration. We can then store the hash of the config in the data flow table, and when anything changes we unset the hash so we know it needs to be resaved. This means we only save versions of workflows that actually get run.

Also when we eventually clean up the logs of run instances we can delete any version which are no longer referenced.

petersistrom:
Resolves #143

### DIFF
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -97,6 +97,7 @@ class dataflow extends persistent {
             'userid' => ['type' => PARAM_INT, 'default' => 0],
             'timemodified' => ['type' => PARAM_INT, 'default' => 0],
             'usermodified' => ['type' => PARAM_INT, 'default' => 0],
+            'confighash' => ['type' => PARAM_TEXT, 'default' => ''],
         ];
     }
 
@@ -663,5 +664,23 @@ class dataflow extends persistent {
 
         // Updates the stored config.
         $this->config = Yaml::dump((array) $config, 2, 4, Yaml::DUMP_MULTI_LINE_LITERAL_BLOCK);
+    }
+
+    /**
+     * Saves the current config to the tool_dataflows_versions table
+     *
+     */
+    public function save_config_version() {
+        global $DB;
+
+        $config = $this->config;
+        $configyaml = $this->config = Yaml::dump((array) $config, 2, 4, Yaml::DUMP_MULTI_LINE_LITERAL_BLOCK);
+        $newconfighash = sha1($configyaml);
+
+        // Updates the stored config.
+        $DB->insert_record('tool_dataflows_versions',
+            (object) ['dataflowid' => $this->id, 'confighash' => $newconfighash, 'configyaml' => $configyaml]);
+
+        $DB->update_record('tool_dataflows', (object) ['id' => $this->id, 'confighash' => $newconfighash]);
     }
 }
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -665,9 +665,9 @@ class dataflow extends persistent {
         // Updates the stored config.
         $this->config = Yaml::dump((array) $config, 2, 4, Yaml::DUMP_MULTI_LINE_LITERAL_BLOCK);
     }
+
     /**
      * Saves the current config to the tool_dataflows_versions table
-     *
      */
     public function save_config_version() {
         global $DB;
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -665,22 +665,24 @@ class dataflow extends persistent {
         // Updates the stored config.
         $this->config = Yaml::dump((array) $config, 2, 4, Yaml::DUMP_MULTI_LINE_LITERAL_BLOCK);
     }
-
     /**
      * Saves the current config to the tool_dataflows_versions table
      *
      */
     public function save_config_version() {
         global $DB;
+        if (empty($this->confighash)) {
+            $config = $this->export();
+            $configyaml = Yaml::dump((array) $config, 2, 4, Yaml::DUMP_MULTI_LINE_LITERAL_BLOCK);
+            $newconfighash = sha1($configyaml);
 
-        $config = $this->config;
-        $configyaml = $this->config = Yaml::dump((array) $config, 2, 4, Yaml::DUMP_MULTI_LINE_LITERAL_BLOCK);
-        $newconfighash = sha1($configyaml);
-
-        // Updates the stored config.
-        $DB->insert_record('tool_dataflows_versions',
-            (object) ['dataflowid' => $this->id, 'confighash' => $newconfighash, 'configyaml' => $configyaml]);
-
-        $DB->update_record('tool_dataflows', (object) ['id' => $this->id, 'confighash' => $newconfighash]);
+            // Inserts the stored config if it is a new config.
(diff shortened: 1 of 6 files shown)


classes/dataflow.php
```diff
@@ -664,4 +665,22 @@ public function set_var($name, $value) {
...
+    public function save_config_version() {
+        global $DB;
+
+        $config = $this->config;
+        $configyaml = $this->config = Yaml::dump((array) $config, 2, 4, Yaml::DUMP_MULTI_LINE_LITERAL_BLOCK);
+        $newconfighash = sha1($configyaml);
+
+        // Updates the stored config.
+        $DB->insert_record('tool_dataflows_versions',
+            (object) ['dataflowid' => $this->id, 'confighash' => $newconfighash, 'configyaml' => $configyaml]);
```
keevan, line 682: I know the issue mentions the configuration, but doesn't configuration also encompass name, alias, dependencies, step type, etc?

classes/dataflow.php
```diff
@@ -664,4 +665,22 @@ public function set_var($name, $value) {
...
+     */
+    public function save_config_version() {
+        global $DB;
+
+        $config = $this->config;
+        $configyaml = $this->config = Yaml::dump((array) $config, 2, 4, Yaml::DUMP_MULTI_LINE_LITERAL_BLOCK);
+        $newconfighash = sha1($configyaml);
+
+        // Updates the stored config.
+        $DB->insert_record('tool_dataflows_versions',
```
brendanheywood, line 681: Should also be an upsert? ie you could change the config back to the same as a previous state so it will have the same hash as before

classes/dataflow.php
brendanheywood: I'd expect a full export and hash that then throw away the export, it should call $this->export()

classes/local/execution/engine.php
```diff
@@ -298,6 +298,10 @@ public function finalise() {
 
             // Stores a dump of the current engine state as the finalstate of the run.
             if (isset($this->run)) {
+                // If confighash is empty save config version.
+                if ($this->dataflow->confighash === null || $this->dataflow->confighash === "") {
```
brendanheywood, line 302: Might be better to move all logic around when to skip a save into save_config_version

classes/dataflow.php
petersistrom: In that case should it just be a check if the record with that hash exists, if not then insert? As there is nothing to update if the hash already exists.

classes/dataflow.php
petersistrom: Updated to $this->export()

classes/local/execution/engine.php
petersistrom: Moved logic to save_config_version

classes/dataflow.php
```diff
@@ -664,4 +665,26 @@ public function set_var($name, $value) {
...
         $this->config = Yaml::dump((array) $config, 2, 4, Yaml::DUMP_MULTI_LINE_LITERAL_BLOCK);
     }
+
+    /**
+     * Saves the current config to the tool_dataflows_versions table
+     *
+     */
+    public function save_config_version() {
+        if ($this->confighash === null || $this->confighash === "") {
+            global $DB;
```
keevan, line 675: Global vars should be defined at the top of the function regardless of when they are used.

classes/dataflow.php
```diff
@@ -664,4 +665,26 @@ public function set_var($name, $value) {
         // Updates the stored config.
         $this->config = Yaml::dump((array) $config, 2, 4, Yaml::DUMP_MULTI_LINE_LITERAL_BLOCK);
     }
+
+    /**
+     * Saves the current config to the tool_dataflows_versions table
+     *
```
keevan, line 671: Unnecessary whitespace here.

classes/dataflow.php
```diff
@@ -664,4 +665,26 @@ public function set_var($name, $value) {
...
+     */
+    public function save_config_version() {
+        if ($this->confighash === null || $this->confighash === "") {
```
keevan, line 674:
```suggestion
        if ($this->confighash === null || $this->confighash === '') {
```

keevan: The only reason this would not work is when it's equal to zero, which is unlikely so an empty check seems more fitting here.

classes/dataflow.php
```diff
@@ -664,4 +665,26 @@ public function set_var($name, $value) {
...
+            global $DB;
+
+            $config = $this->export();
+            $configyaml = Yaml::dump((array)$config, 2, 4, Yaml::DUMP_MULTI_LINE_LITERAL_BLOCK);
```
keevan, line 678: Can you please ensure there is a space before and after the `(cast)` as shown in the example referenced below:

```suggestion
            $configyaml = Yaml::dump((array) $config, 2, 4, Yaml::DUMP_MULTI_LINE_LITERAL_BLOCK);
```

https://www.php-fig.org/psr/psr-12/#61-unary-operators

classes/dataflow.php
```diff
@@ -664,4 +665,26 @@ public function set_var($name, $value) {
...
+            $newconfighash = sha1($configyaml);
+
+            // Inserts the stored config if it is a new config.
+            if (!$DB->record_exists('tool_dataflows_versions', array('dataflowid' => $this->id, 'confighash' => $newconfighash))) {
```
keevan, line 682: Could you please use square brackets shorthand for arrays instead of the "array" language construct.

classes/dataflow.php
```diff
@@ -664,4 +665,26 @@ public function set_var($name, $value) {
...
+                $DB->insert_record('tool_dataflows_versions',
+                    (object)['dataflowid' => $this->id, 'confighash' => $newconfighash, 'configyaml' => $configyaml]);
```
keevan, line 684: Can you please ensure there is a space before and after the `(cast)` as shown in the example referenced below:

https://www.php-fig.org/psr/psr-12/#61-unary-operators

classes/dataflow.php
```diff
@@ -664,4 +665,26 @@ public function set_var($name, $value) {
...
+            }
+            // Set the config hash in dataflows table.
+            $DB->update_record('tool_dataflows', (object)['id' => $this->id, 'confighash' => $newconfighash]);
```
keevan, line 687: Can you please ensure there is a space before and after the `(cast)` as shown in the example referenced below:

https://www.php-fig.org/psr/psr-12/#61-unary-operators

classes/dataflow.php
```diff
@@ -664,4 +665,24 @@ public function set_var($name, $value) {
         // Updates the stored config.
         $this->config = Yaml::dump((array) $config, 2, 4, Yaml::DUMP_MULTI_LINE_LITERAL_BLOCK);
     }
+    /**
```
keevan, line 668: `Expected 1 blank line after function; 0 found`

keevan (commented):
You'll need to bump the version.php and add the bits to the upgrade.php script, but otherwise it seems fine.

keevan (approved):

PR #318, opened 2022-07-20
304 add support for step dep ordering

### LINKED ISSUE #119: Have a 'case' flow step which diverted records to multiple streams
Very similar to the connector step, but acting on a record from a stream and pushing data to exactly one of many outputs.

For instance a simple example could be an input stream of random numbers and it splits it into two streams of odds and evens.

Config is a list of expressions evaluated in order, the number of outputs must match the number of expressions.

It could also have an optional 'else' for a last stream, but if not then anything not matching gets eaten

Because records are pulled through the flow one of the outputs might request a record, but after looking at the input record it will give the record to a different output. So it may need to buffer some small number of records for each output until they are requested. See also #120

### LINKED ISSUE #304: Each step must declare it's outputs as named order outputs (and inputs)
* [x] When a step declares outputs it also gets to declare a name for each of them and an explicit order
* [x] eg the case step would dynamically declare N outputs, and in a certain order, and each output has a name
* [x] By default most steps would have a single output and call that output just the bland step name
* [x] when wiring up dependencies between steps you are not merely choosing a step, you are choosing an output. So by default for most step it would appear unchanged as the new output names will be the same as the previous step names
* [x] but for the case step you would see N options in the 'depends'

So you might see something like this in the UI

```
Trigger step
Reader step
My case step, option A
My case step, option B
My case step, option C
Email step
```

under the hood each output will be a named and a fully qualified 'output' will be the combination of the step and its named output. So multiple steps can safely declare the same output names and not clash'

* [x] when rendering the graph try to tell dot to align the arrows in order with the first at the top

keevan:
Basic support

Example dataflow: 
[even_and_odds_20220725_0518.yml.txt](https://github.com/catalyst/moodle-tool_dataflows/files/9178224/even_and_odds_20220725_0518.yml.txt)

Dataflow details view, with step dependency and the output position shown: 

image

image

image

DB example of it linking to the case step (position is based on option chosen):

image

Known issues:
- [x] No export/import support for it yet.
- [x] Dependency in step form not showing custom label (e.g. provided in case step)
- [x] Custom labels for output connections not displaying in dot graph yet
- [ ] No validation currently to ensure it doesn't break during runtime. 
- [ ] No validation for a case step expression set, but no link for it assigned yet.
- [x] Help text for cases can be improved, but not added yet.
- [x] Cannot see the numbered option as a selected option for it when you're updating a step 
- [x] Multiple flow groups might not work - it is assumed there is only one flow group. The logic for flow caps have been updated so this will likely cause issues if there are more than one flow group available.

Resolves #304 
Resolves #119

### DIFF
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -546,7 +546,11 @@ class dataflow extends persistent {
             ]);
             if (isset($dependency->position)) {
                 $outputlabel = $srcstep->steptype->get_output_label($dependency->position);
-                $connectionstyles = ['label' => $outputlabel];
+                $connectionstyles = [
+                    'label' => $outputlabel,
+                    'fontsize'  => '10',
+                    'fontname'  => 'San Serif',
+                ];
             }
 
             // Final styles.
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -284,7 +284,8 @@ class dataflow extends persistent {
                        sd.stepid AS dest
                   FROM {tool_dataflows_step_depends} sd
              LEFT JOIN {tool_dataflows_steps} step ON step.id = sd.stepid
-                 WHERE step.dataflowid = :dataflowid";
+                 WHERE step.dataflowid = :dataflowid
+              ORDER BY sd.position ASC";
 
         // Note that this works currently because all dependencies are set for each step.
         $edges = $DB->get_records_sql($sql, ['dataflowid' => $this->id]);
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -496,6 +496,8 @@ class dataflow extends persistent {
      * @return     string dotscript
      */
     public function get_dotscript(): string {
+        global $DB;
+
         // Fetch the dot script node from each step to construct them.
         $steps = $this->steps;
         $nodes = [];
@@ -535,7 +537,20 @@ class dataflow extends persistent {
                     $localstyles['color'] = '#008196';
                 }
             }
-            $finalstyles = array_merge($baseconnectionstyles, $localstyles);
+
+            // Output connection labels, if applicable.
+            $connectionstyles = [];
+            $dependency = $DB->get_record('tool_dataflows_step_depends', [
+                'dependson' => $srcid,
+                'stepid' => $destid,
+            ]);
+            if (isset($dependency->position)) {
+                $outputlabel = $srcstep->steptype->get_output_label($dependency->position);
+                $connectionstyles = ['label' => $outputlabel];
+            }
+
+            // Final styles.
+            $finalstyles = array_merge($baseconnectionstyles, $localstyles, $connectionstyles);
 
             $styles = '';
             foreach ($finalstyles as $key => $value) {
--- a/classes/form/step_form.php
+++ b/classes/form/step_form.php
@@ -92,20 +92,53 @@ class step_form extends \core\form\persistent {
 
                 // Check if the current step is a dependant, and if so, INCLUDE the option (and ensure it is selected).
                 $currentstepid = $persistent->id;
+                $selectedposition = null;
                 if (!empty($currentstepid) && isset($dependants[$currentstepid])) {
-                    $availablepositions[] = $dependants[$currentstepid]->position;
+                    $selectedposition = $dependants[$currentstepid]->position;
+                    $availablepositions[] = $selectedposition;
                     sort($availablepositions);
                 }
 
+                // New case position should always show the next available slot for a case step.
+                // Updating a position should always show up to the current and any empty slots below.
+                $maxposition = null;
+                if ($selectedposition) {
+                    $maxposition = $selectedposition;
+                } else {
+                    $maxposition = max(array_column($dependants, 'position')) + 1;
+                }
+
+                // Start plucking positions from the top until there is a gap or the current step position is hit.
+                // This ensures you won't see positions you shouldn't be able to assign, and keeps things in order.
+                // TODO: Also consider when the connecting step has outputs defined already, stop at that maximum instead.
+                $reversed = array_reverse($availablepositions);
+                $current = reset($reversed) + 1;
+                $position = reset($reversed);
+                while (
+                    // No gap.
+                    $current - 1 === $position
+                    // Max not reached.
+                    && ($maxposition === null || $maxposition !== $position)
+                    // Always keep at least one option.
+                    && $position !== 1
+                ) {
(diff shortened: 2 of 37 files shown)


classes/local/execution/iterators/dataflow_iterator.php
```diff
@@ -119,26 +119,41 @@ public function on_next() {
...
 
         // Do not call this for the initial pull (of data) for a reader.
-        if (!empty($this->pulled) || $this->steptype->get_group() !== 'readers') {
-            $this->input->next();
+        if ($this->should_pull_next()) {
+            if ($this->input instanceof dataflow_iterator) {
+                $this->input->next($this);
+            } else {
+                $this->input->next();
+            }
```
keevan, line 149: Ugliest bit in this whole PR I think.

If you have any suggestions happy to consider them. Explanation of the differences. The one consuming `$this` is a dataflow iterator, whereas the one that doesn't is a raw iterator e.g. from DB get recordset or some raw php iterator which cannot accept any parameters.

classes/dataflow.php
```diff
@@ -534,7 +537,20 @@ public function get_dotscript(): string {
...
+
+            // Output connection labels, if applicable.
+            $connectionstyles = [];
+            $dependency = $DB->get_record('tool_dataflows_step_depends', [
+                'dependson' => $srcid,
+                'stepid' => $destid,
+            ]);
+            if (isset($dependency->position)) {
+                $outputlabel = $srcstep->steptype->get_output_label($dependency->position);
+                $connectionstyles = ['label' => $outputlabel];
```
brendanheywood, line 549: Can you make the labels a san serif and a tad smaller

image

classes/local/step/flow_logic_case.php
```diff
@@ -0,0 +1,213 @@
...
+     *
+     * @param \MoodleQuickForm $mform
+     */
+    public function form_add_custom_inputs(\MoodleQuickForm &$mform) {
+        [, $maxoutputs] = $this->outputflows;
+        $mform->addElement(
+            'textarea',
+            'config_cases',
+            get_string('flow_logic_case:cases', 'tool_dataflows'),
+            ['cols' => 50, 'rows' => $maxoutputs, 'placeholder' => "label: <expression>\nsome other label: <expression>"]
```
brendanheywood, line 112: maxoutputs it too big for normal user. In most cases it will be 3-5 so can you make it 5, but then if there is N expressions saved then make it N+2 so always room for more, up to 20

classes/form/step_form.php
```diff
@@ -121,6 +114,92 @@ public function definition() {
         $this->add_action_buttons();
     }
 
+    /**
+     * Prepares and returns an array of dependson options
+     *
+     * @return  array of options this step could depend on
+     */
+    private function get_dependson_options(): array {
```
brendanheywood, line 122: I've getting inconsistent forms, ie all in the same flow depending on what I'm editing I see:

image

image

image

At this stage I'm expecting these to always be the same, except for the case when editing the case itself so it should not be able to depend on any of its own outputs

classes/local/step/flow_logic_case.php
```diff
@@ -0,0 +1,213 @@
...
+ *
+ * @package    tool_dataflows
+ * @author     Kevin Pham <kevinpham@catalyst-au.net>
+ * @copyright  Catalyst IT, 2022
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class flow_logic_case extends flow_logic_step {
+
+    /**
+     * For a join step, it should have 2 or more inputs and for now, up to 20
```
brendanheywood, line 44: join step comment? same below

classes/local/step/flow_logic_case.php
```diff
@@ -0,0 +1,213 @@
...
+     */
+    protected $inputflows = [1, 1];
+
+    /**
+     * For a join step, there should be exactly one output. This is because
+     * without at least one output, there is no need to perform a join.
+     *
+     * @var int[] number of output flows (min, max)
+     */
+    protected $outputflows = [2, 20];
```
brendanheywood, line 57: So if I create a case with 3 outputs and only link up 2 then I'd expect this to be invalid. I'm wondering wether the outputflows should have an upper number which is dependent on its config, ie if it is not configured then it has between 2 and 20, but then if it has been configured then its number of outputs should set to the number of expressions it has??

Or it just has its own extra validation to make sure it is consistent, not fussed either way

classes/form/step_form.php
keevan: > At this stage I'm expecting these to always be the same.

Can you please clarify what your expectations are?

> when editing the case itself so it should not be able to depend on any of its own outputs

This has come in discussion previously. Is the "checking of valid options" for the dependency listing something that should be changed in this PR? Or should it be split into a separate issue? This PR is already getting bigger than it should and it's an independent feature/fix.

classes/local/step/flow_logic_case.php
keevan: Okay. I'll add this validation / check to this PR.

classes/local/step/flow_logic_case.php
keevan: Thanks for the catch, this will be fixed.

classes/local/step/flow_logic_case.php
keevan: Now adjusted to be n+2 with 20 (the class prop) as the maximum

classes/dataflow.php
keevan: image

This is what I've adjusted it to for now.

classes/form/step_form.php
keevan: Check and filter out invalid dependency options in the step form.

keevan: I've pulled out the above into another issue. And we'll adjust as we go.

classes/local/step/flow_logic_case.php
keevan: And this is now added. It's soft validation so it won't allow the dataflow to run, won't block you from saving the case step - because this depends on updating other steps so I think this is a good place for it to be for now.

classes/local/execution/iterators/dataflow_iterator.php
brendanheywood: Can't we simply enforce everywhere that they must always be a dataflow iterator?

I thought all the the raw recordsets etc were wrapped inside a dataflow iterator??

PR #321, opened 2022-07-20
persitent variable null condition - error fix

marcghaly:
Just a fix to avoid this error when no config exists
Screenshot from 2022-07-20 11-18-58
Regards,
 
 Marc


keevan (approved):

keevan:
PR description is a bit backwards. This is to avoid an error when the dataflow doesn't exist.

PR #327, opened 2022-07-21
Issue #256: Add generic checks for dataflow runs

### LINKED ISSUE #256: Generic checks for Check API
This is a range of generic checks which will be applied for free with zero config to all dataflows

* [x] Look at all enabled workflows, if the most recent run of each workflow was aborted then the status is an error
* [x] If multiple flows have errors then mention them all
* [x] in the status details show all workflows and their status

I've split off the more specialized checks per workflow into: https://github.com/catalyst/moodle-tool_dataflows/issues/313

petersistrom:
Resolves #256

### DIFF
--- a/classes/check/dataflow_runs.php
+++ b/classes/check/dataflow_runs.php
@@ -14,6 +14,13 @@
 // You should have received a copy of the GNU General Public License
 // along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
 
+namespace tool_dataflows\check;
+
+use core\check\check;
+use core\check\result;
+use tool_dataflows\dataflow;
+use tool_dataflows\local\execution\engine;
+
 /**
  * Dataflows generic check
  *
@@ -22,13 +29,16 @@
  * @copyright   Catalyst IT
  * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
-namespace tool_dataflows\check;
-use core\check\check;
-use core\check\result;
-use tool_dataflows\dataflow;
-use tool_dataflows\local\execution\engine;
-
 class dataflow_runs extends check {
+
+    /**
+     * Get the short check name
+     *
+     * @return string
+     */
+    public function get_name(): string {
+        return get_string('pluginname', 'tool_dataflows');
+    }
 
     /**
      * Getter for a link to page with more information.
--- a/classes/check/dataflow_runs.php
+++ b/classes/check/dataflow_runs.php
@@ -1,0 +1,91 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Dataflows generic check
+ *
+ * @package     tool_dataflows
+ * @author      Peter Sistrom <petersistrom@catalyst-au.net>
+ * @copyright   Catalyst IT
+ * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+namespace tool_dataflows\check;
+use core\check\check;
+use core\check\result;
+use html_writer;
+use moodle_url;
+use tool_dataflows\dataflow;
+use tool_dataflows\local\execution\engine;
+
+class dataflow_runs extends check {
+
+    /**
+     * Getter for a link to page with more information.
+     *
+     * @return \action_link|null
+     */
+    public function get_action_link(): ?\action_link {
+        $url = new \moodle_url('/admin/dataflows/index.php');
+        return new \action_link($url, get_string('pluginname', 'tool_dataflows'));
+    }
+
+    /**
+     * Getter for the status of the check.
+     *
+     * @return result
+     */
+    public function get_result() : result {
+        global $DB;
+        $enabledldataflows = $DB->get_records('tool_dataflows', ['enabled' => 1]);
+
+        $status = result::OK;
+        $details = '';
+        $summary = '';
+        $runs = false;
+        foreach ($enabledldataflows as $flow) {
+            $dataflow = new dataflow($flow->id);
+            $dataflowlink = html_writer::link(new moodle_url('/admin/tool/dataflows/view.php',
+                ['id' => $dataflow->id]), $dataflow->name);
+
+            if (!empty($dataflow->get_runs(1))) {
+                $runs = true;
+                $lastrun = $dataflow->get_runs(1)[0];
+                $runstate = engine::STATUS_LABELS[$lastrun->status];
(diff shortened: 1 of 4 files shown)


classes/check/dataflow_runs.php
```diff
@@ -0,0 +1,91 @@
...
+        $url = new \moodle_url('/admin/dataflows/index.php');
+        return new \action_link($url, get_string('pluginname', 'tool_dataflows'));
+    }
+
+    /**
+     * Getter for the status of the check.
+     *
+     * @return result
+     */
+    public function get_result() : result {
```
keevan, line 50: > When you have a return type declaration present, there MUST be one space after the colon followed by the type declaration. The colon and declaration MUST be on the same line as the argument list closing parenthesis with no spaces between the two characters.

Could you please fix the type hint spacing by removing the space before the `:`

References:
- https://docs.moodle.org/dev/Coding_style#Functions_and_Methods
- https://www.php-fig.org/psr/psr-12/

classes/check/dataflow_runs.php
```diff
@@ -0,0 +1,91 @@
...
+            $dataflow = new dataflow($flow->id);
+            $dataflowlink = html_writer::link(new moodle_url('/admin/tool/dataflows/view.php',
+                ['id' => $dataflow->id]), $dataflow->name);
+
+            if (!empty($dataflow->get_runs(1))) {
+                $runs = true;
+                $lastrun = $dataflow->get_runs(1)[0];
+                $runstate = engine::STATUS_LABELS[$lastrun->status];
+                $runresultlink = html_writer::link(new moodle_url('/admin/tool/dataflows/view-run.php',
+                    ['id' => $lastrun->id]), 'Run ' . $lastrun->name . " - " . $runstate);
```
keevan, line 68:
```suggestion
                    ['id' => $lastrun->id]), 'Run ' . $lastrun->name . ' - ' . $runstate);
```
Swap to using single quotes, alternatively, plug all the variables into double quotes
```suggestion
                    ['id' => $lastrun->id]), "Run {$lastrun->name} - {$runstate}");
```
Though at that point, I would suggest using a lang string.

classes/check/dataflow_runs.php
```diff
@@ -0,0 +1,91 @@
...
+ * @package     tool_dataflows
+ * @author      Peter Sistrom <petersistrom@catalyst-au.net>
+ * @copyright   Catalyst IT
+ * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+namespace tool_dataflows\check;
+use core\check\check;
+use core\check\result;
+use html_writer;
+use moodle_url;
```
keevan, line 29: I don't mind these being used, but it's usually shorter to call them directly like `\html_writer` in the code, and it seems like some of them already reference it directly. Are you able to remove this and adjust any usage accordingly please?

classes/check/dataflow_runs.php
brendanheywood: +1 yes this should be a lang string

classes/check/dataflow_runs.php
```diff
@@ -0,0 +1,90 @@
...
+        $status = result::OK;
+        $details = '';
+        $summary = '';
+        $runs = false;
+        foreach ($enableddataflows as $flow) {
+            $dataflow = new dataflow($flow->id);
+            $dataflowlink = \html_writer::link(new \moodle_url('/admin/tool/dataflows/view.php',
+                ['id' => $dataflow->id]), $dataflow->name);
+
+            if (!empty($dataflow->get_runs(1))) {
```
keevan, line 61: It's unclear what `dataflow->get_runs` returns (probably a bad method name). 

Could you pull this out into a variable, e.g. `$lastrun`

classes/check/dataflow_runs.php
```diff
@@ -0,0 +1,90 @@
...
+                    ['name' => $lastrun->name, 'state' => $runstate]));
+
+                if ($runstate === 'aborted') {
+                    $summary .= $dataflowlink.': '.$runresultlink.\html_writer::empty_tag('br');
+                    $status = result::ERROR;
+                }
+                $details .= $dataflowlink.': '.$runresultlink;
+            } else {
+                $details .= $dataflowlink.': '.get_string('check:dataflows_no_runs', 'tool_dataflows');
+            }
```
keevan, line 76: I find it difficult to trace through these nested conditions.

Could we have something like:
```
for {
...
  if (!$lastrun) {
    // no last run
    continue;
  }
  ..
  if ($lastrun->status === ABORTED) {
     ... do something
  }
.. 
}
```
.. with a comment above each block what the expected behaviour is at a high level.

classes/check/dataflow_runs.php
```diff
@@ -0,0 +1,90 @@
...
+            $details .= \html_writer::empty_tag('br');
+        }
+
+        if (empty($enableddataflows)) {
+            $summary = get_string('check:dataflows_not_enabled', 'tool_dataflows');
+        } else if (!$runs) {
+            $summary = get_string('check:dataflows_no_runs', 'tool_dataflows');
+        } else if ($runs && $status == result::OK) {
+            $summary = get_string('check:dataflows_completed_successfully', 'tool_dataflows');
+        }
```
keevan, line 86: Can you please add a comment for each case so it can be quickly read at a glance without knowing the internals / groking the lang strings.

classes/check/dataflow_runs.php
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * Getter for the status of the check.
+     *
+     * @return result
+     */
+    public function get_result(): result {
+        global $DB;
+        $enableddataflows = $DB->get_records('tool_dataflows', ['enabled' => 1]);
+
+        $status = result::OK;
```
keevan, line 52:
```suggestion
        global $DB;

        $enableddataflows = $DB->get_records('tool_dataflows', ['enabled' => 1]);
        $status = result::OK;
```

classes/check/dataflow_runs.php
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * Dataflows generic check
+ *
+ * @package     tool_dataflows
+ * @author      Peter Sistrom <petersistrom@catalyst-au.net>
+ * @copyright   Catalyst IT
+ * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+namespace tool_dataflows\check;
+use core\check\check;
```
keevan, line 26: Perhaps one space before the `use` block / after the namespace definition.

classes/check/dataflow_runs.php
```diff
@@ -0,0 +1,101 @@
...
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Dataflows generic check
+ *
+ * @package     tool_dataflows
+ * @author      Peter Sistrom <petersistrom@catalyst-au.net>
+ * @copyright   Catalyst IT
+ * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
```
keevan, line 24: Please move this doc block to live just above the class declaration.

classes/check/dataflow_runs.php
```diff
@@ -0,0 +1,110 @@
...
+class dataflow_runs extends check {
+    /**
```
keevan, line 33:
```
FILE: admin/tool/dataflows/classes/check/dataflow_runs.php
-------------------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
-------------------------------------------------------------------------------------------------------
 38 | ERROR | [x] Expected 1 blank line before function; 0 found
-------------------------------------------------------------------------------------------------------
-------------------------------------------------------------------------------------------------------
```
Please add some whitespace here

PR #328, opened 2022-07-21
Issue #297: Clarify permitted dirs

### LINKED ISSUE #297: Clarify permitted directories setting and provide examples
image

- [ ] Currently difficult to know what sort of value should be entered here, so an example would be good.
- [ ] It would be good to have several formats / examples shown e.g. in the help text below so the user can determine if any of those are relevant.
- [ ] Allow generic placeholders for [dataroot] eg ```[dataroot]/import``` so you can make workflows more transferable between sites and they still work

Could potentially include an example step type that uses this, e.g. the writer_stream step via the stream name field

jaypha:
Resolves #297 

Expand description of permitted directories setting.
Add support for expanding [dataroot] placeholder.

### DIFF
--- a/classes/admin_setting_permitted_directories.php
+++ b/classes/admin_setting_permitted_directories.php
@@ -1,0 +1,72 @@
+<?php
+// This file is part of Moodle - https://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_dataflows;
+
+/**
+ * A custom setting for the permitted directories that a dataflow can interact with.
+ *
+ * @package   tool_dataflows
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class admin_setting_permitted_directories extends \admin_setting_configtextarea {
+
+    /** File scheme string. */
+    protected const FILE_SCHEME = 'file://';
+
+    /**
+     * Validate the setting data as a list of filepaths.
+     *
+     * @param string $data Data to be validated.
+     * @return bool|mixed|string true for success or string:error on failure
+     */
+    public function validate($data) {
+        global $CFG;
+
+        if (empty($data)) {
+            return true;
+        }
+
+        $lines = explode("\n", trim($data));
+
+        $errors = [];
+        foreach ($lines as $line) {
+
+            // Substitute dataroot placeholder, if found.
+            if (substr($line, 0, strlen(helper::DATAROOT_PLACEHOLDER)) === helper::DATAROOT_PLACEHOLDER) {
+                $line = $CFG->dataroot . substr($line, strlen(helper::DATAROOT_PLACEHOLDER));
+            }
+
+            // Remove file scheme, if found.
+            if (substr($line, 0, strlen(self::FILE_SCHEME)) === self::FILE_SCHEME) {
+                $line = substr($line, strlen(self::FILE_SCHEME));
+            }
+
+            // Test for path validity. Must be valid and absolute.
+            if (!helper::is_filepath($line)) {
+                $errors[] = get_string('path_invalid', 'tool_dataflows', $line);
+            } else if (helper::path_is_relative($line)) {
+                $errors[] = get_string('path_not_absolute', 'tool_dataflows', $line);
+            }
+        }
+        if (count($errors)) {
+            return implode(' ', $errors);
+        }
+        return true;
+    }
+}
--- a/classes/helper.php
+++ b/classes/helper.php
@@ -26,11 +26,17 @@ namespace tool_dataflows;
  */
 class helper {
 
-    /** The level where you switch to inline YAML */
-     const YAML_DUMP_INLINE_LEVEL = 5;
+    /** The level where you switch to inline YAML. */
+    public const YAML_DUMP_INLINE_LEVEL = 5;
 
-     /** The amount of spaces to use for indentation of nested nodes*/
-     const YAML_DUMP_INDENT_LEVEL = 2;
+    /** The amount of spaces to use for indentation of nested nodes. */
+    public const YAML_DUMP_INDENT_LEVEL = 2;
+
+    /** String used to indicate the dataroot directory. */
+    public const DATAROOT_PLACEHOLDER = '[dataroot]';
+
+    /** @var null|bool Is this Windows?  */
+    protected static $iswindows = null;
 
     /**
      * Get the scheme for a path string. will default to 'file' if none present.
@@ -104,7 +110,7 @@ class helper {
             return true;
         }
 
-        $permitteddirs = explode(PHP_EOL, trim(get_config('tool_dataflows', 'permitted_dirs')));
+        $permitteddirs = self::get_permitted_dirs();
         foreach ($permitteddirs as $dir) {
             if (substr($path, 0, strlen($dir)) === $dir) {
                 return true;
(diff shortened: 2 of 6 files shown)


jaypha:
I still don't get the middle point in the original issue.

keevan:
Some some phpdoc issues:
```
admin/tool/dataflows/classes/admin_setting_permitted_directories.php
    Line 32: Phpdocs for function admin_setting_permitted_directories::validate has incomplete parameters list
--
  admin/tool/dataflows/classes/helper.php
    Line 35: Variable helper::$iswindows is not documented
    Line 142: Phpdocs for function helper::is_filepath has incomplete parameters list
```

classes/helper.php
```diff
@@ -27,10 +27,13 @@
...
-     const YAML_DUMP_INLINE_LEVEL = 5;
+    const YAML_DUMP_INLINE_LEVEL = 5;
 
-     /** The amount of spaces to use for indentation of nested nodes*/
-     const YAML_DUMP_INDENT_LEVEL = 2;
+    /** The amount of spaces to use for indentation of nested nodes*/
+    const YAML_DUMP_INDENT_LEVEL = 2;
+
+    /** @var null|bool Is this Windows?  */
+    protected static $iswindows = null;
```
keevan, line 36: Can we please set the default to be `false` instead of `null`, just so we don't have to play around with 2 types.

classes/helper.php
```diff
@@ -27,10 +27,13 @@
 class helper {
 
     /** The level where you switch to inline YAML */
-     const YAML_DUMP_INLINE_LEVEL = 5;
+    const YAML_DUMP_INLINE_LEVEL = 5;
 
-     /** The amount of spaces to use for indentation of nested nodes*/
-     const YAML_DUMP_INDENT_LEVEL = 2;
+    /** The amount of spaces to use for indentation of nested nodes*/
+    const YAML_DUMP_INDENT_LEVEL = 2;
```
keevan, line 33: Thank you for fixing up the indentation here.

classes/admin_setting_permitted_directories.php
```diff
@@ -0,0 +1,63 @@
...
+            }
+            if (!helper::is_filepath($line)) {
+                $errors[] = get_string('path_invalid', 'tool_dataflows', $line);
+            } else if (helper::path_is_relative($line)) {
+                $errors[] = get_string('path_not_absolute', 'tool_dataflows', $line);
+            }
+        }
+        if (count($errors)) {
+            return implode(' ', $errors);
+        }
```
keevan, line 60: Can you please add a brief comment to each if block to make it easier to understand.

classes/helper.php
```diff
@@ -113,4 +116,56 @@ public static function path_validate(string $path) {
...
+     * placeholder will be substituted with the correct dir.
+     *
+     * @return array
+     */
+    public static function get_permitted_dirs(): array {
+        global $CFG;
+
+        $permitteddirs = explode(PHP_EOL, trim(get_config('tool_dataflows', 'permitted_dirs')));
+        foreach ($permitteddirs as &$dir) {
+            if (substr($dir, 0, 10) == '[dataroot]') {
```
keevan, line 131: Can you please add a comment for this if statement. I think the related text from the function description can be pulled down, and the function description can be shortened as well - but I'll leave extra bit up to you.

classes/helper.php
jaypha: That will require changing the algorithm. Null indicates that it has not yet been set, whereas false indicates that it has been set.

classes/helper.php
jaypha: NP

classes/helper.php
keevan: Okay, that is fine for now. Appreciate the explanation. :slightly_smiling_face:

classes/helper.php
```diff
@@ -113,4 +116,56 @@ public static function path_validate(string $path) {
...
+            }
+        }
+        return $permitteddirs;
+    }
+
+    /**
+     * Determines if the given path is a valid filepath.
+     * Derived from https://stackoverflow.com/a/12126772.
+     *
+     * @param $path string The path to check.
```
keevan, line 142:
```suggestion
     * @param string $path The path to check.
```
The type should come before the variable name

classes/admin_setting_permitted_directories.php
```diff
@@ -0,0 +1,63 @@
...
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class admin_setting_permitted_directories extends \admin_setting_configtextarea {
+
+    /**
+     * Validate the setting data as a list of filepaths.
+     *
+     * @param $data array Data to be validated.
```
keevan, line 32:
```suggestion
     * @param array $data Data to be validated.
```
The type should come before the variable name. Also is the data always an array? I see at line 42 that it's being trimmed which might indicate that it is in fact a string? Could you please confirm and add the appropriate type hints to the function parameters as well please.

classes/admin_setting_permitted_directories.php
jaypha: done

classes/helper.php
jaypha: done

classes/helper.php
```diff
@@ -113,4 +116,54 @@ public static function path_validate(string $path) {
...
+            $tmp = dirname(__FILE__);
+            self::$iswindows = strpos($tmp, '/', 0) === false;
+        }
+
+        if (self::$iswindows) {
+            if (strpos($path, ":") == 1 && preg_match('/[a-zA-Z]/', $path[0])) {
+                $tmp = substr($path, 2);
+                $bool = preg_match('/^[^*?"<>|:]*$/', $tmp);
+                return ($bool === 1); // So that it will return only true and false.
+            }
```
keevan, line 163: Can you please clarify what this is checking? The regex + variables used doesn't make this very clear.

Perhaps add a comment, and use more meaningful variable names. (e.g. what is $tmp, $bool?)

classes/admin_setting_permitted_directories.php
```diff
@@ -0,0 +1,69 @@
...
+        $errors = [];
+        foreach ($lines as $line) {
+
+            // Substitute dataroot placeholder, if found.
+            if (substr($line, 0, 10) == '[dataroot]') {
+                $line = $CFG->dataroot . substr($line, 10);
+            }
+
+            // Remove file scheme, if found.
+            if (substr($line, 0, 7) == 'file://') {
```
keevan, line 53:
```suggestion
            if (substr($line, 0, 7) === 'file://') {
```

classes/admin_setting_permitted_directories.php
```diff
@@ -0,0 +1,69 @@
...
+            return true;
+        }
+
+        $lines = explode("\n", trim($data));
+
+        $errors = [];
+        foreach ($lines as $line) {
+
+            // Substitute dataroot placeholder, if found.
+            if (substr($line, 0, 10) == '[dataroot]') {
```
keevan, line 48: `[dataroot]` should be moved to a constant and referenced everywhere as such, even passing it to the help text. I'll leave the constant name up to you.

keevan: Also it should use something like `strlen(self::DATAROOT_CONSTANT)` rather than the magic number 10 which will break if the keyword changes. Please apply this in the relevant spots across the PR as well.

classes/admin_setting_permitted_directories.php
jaypha: thanks

classes/helper.php

classes/admin_setting_permitted_directories.php
jaypha: done

classes/admin_setting_permitted_directories.php
jaypha: Ha! Ha! Same as what I got you for.
done

classes/helper.php
```diff
@@ -113,4 +119,58 @@ public static function path_validate(string $path) {
...
+            self::$iswindows = strpos($tmp, '/', 0) === false;
+        }
+
+        if (self::$iswindows) {
+            // Look for a drive name (e.g. C:).
+            if (strpos($path, ":") == 1 && preg_match('/[a-zA-Z]/', $path[0])) {
+                // Strip the drive name.
+                $localpath = substr($path, 2);
+                // Match against valid Windows path syntax.
+                return (preg_match('/^[^*?"<>|:]*$/', $localpath) === 1);
```
keevan, line 169: Much clearer now. Awesome stuff @jaypha

lang/en/tool_dataflows.php
```diff
@@ -35,7 +35,13 @@
...
 $string['enabled_help'] = '';
 $string['permitted_dirs'] = 'Permitted directories';
-$string['permitted_dirs_desc'] = 'List directories here to allow them to be read from/written to by dataflow steps.';
+$string['permitted_dirs_desc'] = "List directories here to allow them to be read from/written to by dataflow steps.
+    One directory per line. Each directory must be an absolute path. You can use the place holder '[dataroot]' for the
+    site's data root directory.\n
+    Examples.
+    /tmp
+    /home/me/mydata
+    [dataroot]/somedir";
```
keevan, line 44: Perhaps you should use the dataroot constant for this as well? (Passing it through `{$a}`)

classes/helper.php
```diff
@@ -113,4 +119,58 @@ public static function path_validate(string $path) {
...
+        }
+
+        if (self::$iswindows === null) {
+            $tmp = dirname(__FILE__);
+            self::$iswindows = strpos($tmp, '/', 0) === false;
+        }
+
+        if (self::$iswindows) {
+            // Look for a drive name (e.g. C:).
+            if (strpos($path, ":") == 1 && preg_match('/[a-zA-Z]/', $path[0])) {
```
keevan, line 165:
```suggestion
            if (strpos($path, ":") === 1 && preg_match('/[a-zA-Z]/', $path[0])) {
```

settings.php
```diff
@@ -14,6 +14,8 @@
 // You should have received a copy of the GNU General Public License
 // along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
 
+use \tool_dataflows\helper;
```
keevan, line 17:
```
admin/tool/dataflows/settings.php
    Line 17: File-level phpdocs block is not found
```
This should be moved below the file-level phpdoc block

PR #330, opened 2022-07-22
Issue #324: Update path lang strings

### LINKED ISSUE #324: curl scratch dir
All places which take a path should be changed to accept a stratch dir path.We have converted the curl step but:

* [x] ~~need to convert the json parsing step~~
* [ ] update help docs for these fields to talk about the scratch dir
* [ ] it may be useful to create a new custom field type which refactors a lot of the help and validation
* [ ] for the s3 paths they can be normal paths or s3 so maybe concat the two paragraphs of help
* [ ] Stream writer output path
* [ ] json reading input
* [ ] File contents display
* [ ] curl download path

petersistrom:
Resolve #324

### DIFF
--- a/lang/en/tool_dataflows.php
+++ b/lang/en/tool_dataflows.php
@@ -148,6 +148,7 @@ $string['output_connector_link_limit'] = '{$a} output connector link';
 $string['output_connector_link_limit_plural'] = '{$a} output connector links';
 $string['output_connector_link_limit_range'] = 'between {$a->min} and {$a->max} output connector links';
 $string['no_output_allowed'] = 'No outputs allowed';
+$string['path_help'] = 'Path to the file to be read or written to. This can be a simple filename (e.g. input.txt/output.txt).';
 
 $string['requires_with_or'] = 'Requires {$a->str1} or {$a->str2}.';
 $string['requires'] = 'Requires {$a->str1} and {$a->str2}.';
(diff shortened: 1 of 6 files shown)


lang/en/tool_dataflows.php
```diff
@@ -148,6 +148,7 @@
 $string['output_connector_link_limit_plural'] = '{$a} output connector links';
 $string['output_connector_link_limit_range'] = 'between {$a->min} and {$a->max} output connector links';
 $string['no_output_allowed'] = 'No outputs allowed';
+$string['path_help'] = 'Path to the file to be read or written to. This can be a simple filename (e.g. input.txt/output.txt).';
```
brendanheywood, line 151: Lets be a bit more specific that it can be a relative path, or an absolute path, or a stream url (and confirm that all 3 work in all the places we touch here), suggestion

---

File path to be read or written eg:

```
out.csv           # A path relative to the flows temp working dir
/var/data.json    # An absolute path
file:///my/in.txt # Any valid php stream url
```

lang/en/tool_dataflows.php
```diff
@@ -288,9 +287,8 @@
 $string['connector_s3:key'] = 'Key';
 $string['connector_s3:secret'] = 'Secret';
 $string['connector_s3:source'] = 'Source / From';
-$string['connector_s3:source_help'] = 'Path to the source file. This can be a local (e.g. output.csv) or S3 path (e.g. s3://path/to/file).';
+$string['connector_s3:source_help'] = 'This can also be a S3 path (e.g. s3://path/to/file).';
```
brendanheywood, line 290: In the case of the s3 connector maybe we can split up the text and the example paths, and the paths can be hard coded and passed into the string as input and for the s3 field we can just add in the extra s3 example and leave the text above it unchanged

So the s3 field would just have an extra line in the example:

```
s3://path/to/file   # Any s3 stream url
```

brendanheywood: and put the extra example first in the list of examples

brendanheywood:
This looks good but is missing something in the curl step File / Response destination field

Issue #311, opened 2022-07-19
Show the large number of variables as a nested tree widget


brendanheywood:
As the number of steps grows and the number of variables inside each type of step grows this list is going to get very out of hand. This should be a reference which you need when you want it but should take up almost no space by default. I was mulling on a few different ways to present it (and still open to other better ideas) but I think the best bang for buck is showing all the available variables as a nested tree structure:

![image](https://user-images.githubusercontent.com/187449/179701911-3c2b0596-b3dc-4c05-8857-fac964a1db60.png)

This doesn't need to be fancy and I don't even think we need js, just a bunch of nested details / summary html elements and a very light sprinkling of css

* [x] show all vars as a nested tree of nodes
* [x] collapse and show
* [x] click to select https://github.com/catalyst/moodle-tool_dataflows/issues/310

PR #332, opened 2022-07-22
Resolves issue #311 feat: collapsible variables added

marcghaly:
Hi @brendanheywood, this is a PR for issue #311, might need to displace the group position in the grid, but the feature is here

### DIFF
--- a/templates/available-fields.mustache
+++ b/templates/available-fields.mustache
@@ -30,6 +30,8 @@
 
     Example context (json):
     {
+        "groups": {
+        }
     }
 
 }}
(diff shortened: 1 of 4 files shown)


templates/available-fields.mustache
```diff
@@ -40,11 +40,24 @@
     <div class="fieldcontainer col-md-9 form-inline felement">
         <p>{{# str }} available_fields_help, tool_dataflows, <code>${<span></span>{ dataflow.name }}</code> {{/ str }}</p>
         {{#groups}}
-            <div class="fieldtitle text-capitalize">{{ name }}:</div>
-            <div class="fieldgroup">
-                {{#fields}}
-                    <span class="badge badge-dark rounded-0 mb-1 mr-1" title="{{ title }}">{{ text }}</span>
-                {{/fields}}
+            <span class="badge badge-info rounded-0 mb-1 mr-1" type="button" data-toggle="collapse" data-target="#collapsableGroup-{{name}}" aria-expanded="false" aria-controls="collapseExample">
```
brendanheywood, line 43: So functionally this is kinda half working, when I click on things then children open and close, but cosmetically I want this to appear like a tree:

image

and this is what I am seeing:

image

Without getting into the code too much, at a glance it feels wrong because there is no recursion and I'd assume we need to recurse. We have an unknown level of N deep nested tags so I would expect we'd need to have a tagtree template which calls itself for its children tags. Tags which open and close should appear the same regardless of their level and have either a -/+ or arrows that swivel to show their open state and whether or not they have children

marcghaly:
Hi @brendanheywood, I have fixed issues and this is it what it looks like at the moment, may need to move it on the webpage grid but feature is here.

Screenshot from 2022-07-26 11-34-45

Regards,

Marc

brendanheywood:
just a heads up I'm digging into this one, I'll likely tack on a couple cosmetic commits and then merge it

brendanheywood:
Preview of the style polish:

image

keevan (commented):
Some small comments. 

Tested locally, looks and feels great. 

image

Happy to merge this in after the adjustments and anything extra can be a new issue.

PR #345, opened 2022-07-26
Issue #106: Rewire flow on step deletion

### LINKED ISSUE #106: Be able to delete a step

petersistrom:
Resolves #106

### DIFF
--- a/classes/step.php
+++ b/classes/step.php
@@ -534,8 +534,29 @@ class step extends persistent {
     protected function before_delete() {
         global $DB;
 
+        // Attempt to rewire/connect steps previously linked to this step.
+        $dependents = $this->dependents();
+        $dependencies = $this->dependencies();
+
+        if (count($dependencies) === 1 && count($dependents) === 1) {
+            // 1 input/output. Get new input and output and if valid rewire flow.
+            $inputstep = new step(current($dependencies)->id);
+            $outputstep = new step(current($dependents)->id);
+
+            // Validate new output against new input step.
+            $outputvalid = $inputstep->validate_outputs($dependents);
+
+            // Validate new input against new output step.
+            $inputvalid = $outputstep->validate_inputs($dependencies);
+
+            if ($inputvalid === true && $outputvalid === true) {
+                // New flow valid, update depends on.
+                $outputstep->depends_on($dependencies);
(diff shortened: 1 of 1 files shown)


classes/step.php
```diff
@@ -534,8 +534,29 @@ public function get_export_data(): array {
...
         global $DB;
 
+        // Attempt to rewire/connect steps previously linked to this step.
+        $dependents = $this->dependents();
+        $dependencies = $this->dependencies();
+
+        if (count($dependencies) === 1 && count($dependents) === 1) {
+            // 1 input/output. Get new input and output and if valid rewire flow.
+            $inputstep = new step(current($dependencies)->id);
+            $outputstep = new step(current($dependents)->id);
```
keevan, line 544: Small suggestion to name this to before and after steps, as it is 1:1 and I think it's easier to reason about than input output steps. 

If there were more steps, I would have considered using the terms upstreams (input steps) and downstreams (output steps).

keevan (changes requested):
Changes seem okay. Are you able to please add some unit tests to ensure it is working properly?

I would like to see the following cases:
- simple chain of connector steps, removing a step in the middle will link up the connections.
- an already invalid dataflow, made valid by removing the in between step
- a valid dataflow where the reader step is removed. I'm expecting it to NOT connect the before and after steps, since the dataflow would be in an invalid state.

If there are other cases you can think of, please add a unit test for them as well.

Thanks @petersistrom

classes/step.php
```diff
@@ -534,8 +534,29 @@ public function get_export_data(): array {
...
+
+            // Validate new output against new input step.
+            $outputvalid = $inputstep->validate_outputs($dependents);
+
+            // Validate new input against new output step.
+            $inputvalid = $outputstep->validate_inputs($dependencies);
+
```
keevan, line 551: I would add a high comment here, such as "The before and after steps will be rewired together only if the resulting dataflow is valid after the changes". 

Please go with what you think would work best.

PR #348, opened 2022-07-26
Issue #45: Wait step

### LINKED ISSUE #45: Have a 'wait' connector
This is configured with an expression, probably an strtotime, and it will simply pause execution and then resume at that time. Probably needs to set the workflow into a blocked state and stop execution and queue up an adhoc task to resume it.

jaypha:
Resolves #45.
Add connector_wait class.

### DIFF
--- a/tests/tool_dataflows_connector_wait_test.php
+++ b/tests/tool_dataflows_connector_wait_test.php
@@ -1,0 +1,81 @@
+<?php
+// This file is part of Moodle - https://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


keevan (changes requested):
```
/home/runner/work/moodle-tool_dataflows/moodle-tool_dataflows/moodle/admin/tool/dataflows/classes/local/step/connector_wait.php
    Line 29: Function connector_wait::form_define_fields is not documented
```

tests/tool_dataflows_connector_wait_test.php
```diff
@@ -0,0 +1,83 @@
...
+        return $dataflow;
+    }
+
+    /**
+     * Tests connector_wait
+     *
+     * @covers \tool_dataflows\local\step\connector_wait
+     */
+    public function test_wait() {
+        $timesec = 5;
```
brendanheywood, line 72: tune to 2 to make tests faster?

keevan:
Thanks for adding the tests, there's one extra small thing, but otherwise I'm happy to merge this.
```
tests/tool_dataflows_connector_wait_test.php:57:38: error - Multi-line array contains a single value; use single-line array instead
```

PR #369, opened 2022-07-28
Issue #231: Escape inputs to dot cli

### LINKED ISSUE #231: Make sure anything calling the dot cli to generate the visual flows, is sanitised input
Currently it can be escaped and has the potential to call and execute arbitrary commands.

petersistrom:
Resolves #231

### DIFF
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -558,7 +558,8 @@ class dataflow extends persistent {
 
             $styles = '';
             foreach ($finalstyles as $key => $value) {
-                // TODO escape all attributes correctly.
+                // Escape all attributes correctly.
+                $value = addslashes($value);
                 $styles .= "$key =\"$value\", ";
             }
             trim($styles);
--- a/classes/step.php
+++ b/classes/step.php
@@ -765,6 +765,7 @@ class step extends persistent {
      * @return     string
      */
     public function get_dotscript($contentonly = false): string {
+        $this->name = addslashes($this->name);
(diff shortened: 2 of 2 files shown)


petersistrom:
Naming a step with a xss polyglot such as 
`jaVasCript:/*-/*`/*\`/*'/*"/**/(/* */oNcliCk=alert() )//%0D%0A%0d%0a//</stYle/</titLe/</teXtarEa/</scRipt/--!>\x3csVg/<sVg/oNloAd=alert()//>\x3e`
Causes a syntax error and breaks the dataflow, escaping the input using addslashes() resolves the error
image

classes/dataflow.php
```diff
@@ -558,7 +558,8 @@ public function get_dotscript(): string {
 
             $styles = '';
             foreach ($finalstyles as $key => $value) {
-                // TODO escape all attributes correctly.
+                // Escape all attributes correctly.
+                $value = addslashes($value);
```
brendanheywood, line 562: I doubt this escaping is correct, but even assuming it is I think there should be an escape_dot method so it is all in one place if we do need to improve it later

classes/step.php
```diff
@@ -765,6 +765,7 @@ public function has_side_effect(): bool {
      * @return     string
      */
     public function get_dotscript($contentonly = false): string {
+        $this->name = addslashes($this->name);
```
brendanheywood, line 768: This is changing the name in place, I think the dot escaping should only affect the dot output and the escaping shouldn't accidentally leak to somewhere else

PR #371, opened 2022-07-28
Issue #341: Include checks for graphviz dot dependency

### LINKED ISSUE #341: Declare graphviz dependency properly
Tasks:
* [x] On the main dataflows page show a warning if the dot path isn't a readable file
* [x] Show this on the admin settings page too
* [x] Whenever an image would have been shown, instead show a boilerplate fallback svg or images, or even just a plain red square or something rather than a raw error

---

Moodle: MOODLE_311_STABLE
Dataflows: MOODLE_35_STABLE 
as of today

This occurs on admin/tool/dataflows/view.php?id=1

```
Exception - failed to execute cmd: "dot -Tsvg". stderr: `sh: dot: command not found `

[More information about this error](http://docs.moodle.org/311/en/error/moodle/generalexceptionmessage)

Debug info:
Error code: generalexceptionmessage×Dismiss this notification
Stack trace:
line 105 of /admin/tool/dataflows/classes/visualiser.php: Exception thrown
line 331 of /admin/tool/dataflows/classes/visualiser.php: call to tool_dataflows\visualiser::generate()
line 68 of /admin/tool/dataflows/view.php: call to tool_dataflows\visualiser::display_dataflows_view_page()
×Dismiss this notification
```

As it turns out RTFM => graphviz must be installed.

Can it be declared in an environment.xml like this - https://github.com/maths/moodle-qtype_stack/blob/master/environment.xml ?

jaypha:
Resolves #341 

Display warnings if dot not available.
Show alternate icon to dot images.

### DIFF
--- a/classes/dataflows_table.php
+++ b/classes/dataflows_table.php
@@ -79,8 +79,16 @@ class dataflows_table extends \table_sql {
      * @return string
      */
     public function col_preview(\stdClass $record): string {
+        global $OUTPUT;
         $imgurl = new \moodle_url('/admin/tool/dataflows/visual.php', ['id' => $record->id, 'type' => 'png']);
-        $img = \html_writer::img($imgurl, "Dataflow #{$record->id} visualisation", ['height' => '30px', 'class' => 'lightbox']);
+        if (helper::is_graphviz_dot_installed()) {
+            $img = \html_writer::img($imgurl, "Dataflow #{$record->id} visualisation", ['height' => '30px', 'class' => 'lightbox']);
+        } else {
+            $img = $OUTPUT->render(
+                new \pix_icon(helper::GRAPHVIZ_ALT_ICON,
+                get_string('preview_unavailable', 'tool_dataflows'))
+            ) . get_string('preview_unavailable', 'tool_dataflows');
+        }
         $dataflowstepsurl = new \moodle_url('/admin/tool/dataflows/view.php', ['id' => $record->id]);
         return \html_writer::link($dataflowstepsurl, $img);
     }
--- a/classes/helper.php
+++ b/classes/helper.php
@@ -35,6 +35,12 @@ class helper {
     /** String used to indicate the dataroot directory. */
     public const DATAROOT_PLACEHOLDER = '[dataroot]';
 
+    /** Alternate icon to display if Graphviz is not available. */
+    public const GRAPHVIZ_ALT_ICON = 'i/incorrect';
+
+    /** Location of the dependency section of the readme. */
+    public const README_DEPENDENCY_LINK = 'https://github.com/catalyst/moodle-tool_dataflows#dependencies';
+
     /** @var null|bool Is this Windows?  */
     protected static $iswindows = null;
 
@@ -185,5 +191,20 @@ class helper {
         }
 
         return false;
+    }
+
+    /**
+     * Determines if the Graphviz dot program can be accessed.
+     *
+     * @return bool
+     */
+    public static function is_graphviz_dot_installed(): bool {
+        global $CFG;
+        if (!empty($CFG->pathtodot)) {
+            return is_executable($CFG->pathtodot);
+        }
+
+        // Don't know where dot is, so look for it.
+        return !empty(shell_exec('which dot'));
     }
 }
--- a/classes/visualiser.php
+++ b/classes/visualiser.php
@@ -73,24 +73,31 @@ class visualiser {
      * @author     Kevin Pham <kevinpham@catalyst-au.net>
      */
     public static function generate(string $dotscript, ?string $type = 'svg') {
-        global $CFG;
+        global $CFG, $OUTPUT;
+
+        if (!helper::is_graphviz_dot_installed()) {
+            return $OUTPUT->render(
+                new \pix_icon(helper::GRAPHVIZ_ALT_ICON,
+                    get_string('preview_unavailable', 'tool_dataflows'))
+            ) . get_string('preview_unavailable', 'tool_dataflows');
+        }
 
         $descriptorspec = [
-           // The stdin is a pipe that the child will read from.
-           0 => ['pipe', 'r'],
-           // The stdout is a pipe that the child will write to.
-           1 => ['pipe', 'w'],
-           // The stderr is a pipe that the child will write to.
-           2 => ['pipe', 'w'],
+            // The stdin is a pipe that the child will read from.
+            0 => ['pipe', 'r'],
+            // The stdout is a pipe that the child will write to.
+            1 => ['pipe', 'w'],
+            // The stderr is a pipe that the child will write to.
+            2 => ['pipe', 'w'],
         ];
 
         $cmd = (!empty($CFG->pathtodot) ? $CFG->pathtodot : 'dot') . ' -T' . $type;
         $process = proc_open(
-          $cmd,
-          $descriptorspec,
-          $pipes,
-          sys_get_temp_dir(),
-          ['PATH' => getenv('PATH')]
+            $cmd,
+            $descriptorspec,
+            $pipes,
+            sys_get_temp_dir(),
+            ['PATH' => getenv('PATH')]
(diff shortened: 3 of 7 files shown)


classes/visualiser.php
```diff
@@ -73,45 +73,52 @@ public static function breadcrumb_navigation(array $crumbs) {
...
-          ['PATH' => getenv('PATH')]
-        );
+        global $CFG, $OUTPUT;
+
+        if (!helper::is_graphviz_dot_installed()) {
+            return $OUTPUT->render(
+                new \pix_icon(helper::GRAPHVIZ_ALT_ICON,
+                get_string('no_dot_installed_short', 'tool_dataflows'))
+            );
+        } else {
```
keevan, line 83: This else block is unnecessary due to the early return. Also caused a bunch of changes to appear in the PR which weren't relevant.

classes/helper.php
```diff
@@ -173,4 +176,19 @@ public static function is_filepath(string $path): bool {
...
+     * @return bool
+     */
+    public static function is_graphviz_dot_installed(): bool {
+        global $CFG;
+        if (!empty($CFG->pathtodot)) {
+            return is_executable($CFG->pathtodot);
+        } else {
+            // Don't know where dot is, so look for it.
+            return !empty(shell_exec('which dot'));
+        }
```
keevan, line 192: Please remove the unnecessary else block, due to the early return in the if block.

lang/en/tool_dataflows.php
```diff
@@ -207,6 +207,8 @@
 $string['path_invalid'] = 'Path "{$a}" is not permitted.';
 $string['path_not_absolute'] = 'Path "{$a}" is not an absolute path.';
 $string['unknown_placeholder'] = 'Placeholder "{$a}" is not recognised.';
+$string['no_dot_installed'] = 'No valid path to Graphviz dot found (Have you installed it?).';
```
keevan, line 210: A link to the repository dependency section would be useful here.

e.g. describe the problem that occurred, then show an action so they can fix it.

settings.php
```diff
@@ -39,6 +39,18 @@
         new moodle_url('/admin/tool/dataflows/index.php'));
 
     if ($ADMIN->fulltree) {
+
```
keevan, line 42: Can you remove the extra whitespace here.

keevan (commented):
Are you able to also add some screenshots to the PR so I know how it looks?

Thanks Jason.

lang/en/tool_dataflows.php
```diff
@@ -207,6 +207,8 @@
...
+$string['no_dot_installed_short'] = 'Graphviz not found';
```
keevan, line 211:
```suggestion
$string['no_dot_installed_short'] = 'Graphviz not installed';
```
Or perhaps this would be clearer?
```suggestion
$string['no_dot_installed_short'] = 'Graphviz not installed - dot unavailable';
```

jaypha: It could be that the specified path is wrong.

jaypha:
The main message.
Screenshot from 2022-07-28 15-27-43
The alternate icons
Screenshot from 2022-07-28 15-28-15
The step chooser with name printed next to the icon.
Screenshot from 2022-07-28 15-28-41

lang/en/tool_dataflows.php
jaypha: done

classes/visualiser.php
jaypha: done

classes/helper.php

settings.php
jaypha: done

lang/en/tool_dataflows.php
keevan: Would you be able to make this clearer so the user can tell whether or not they need to install it, or need to fix the path. I think that would be the sweet spot.

classes/dataflows_table.php
```diff
@@ -79,8 +79,16 @@ public function make_columns(): void {
...
+        global $OUTPUT;
         $imgurl = new \moodle_url('/admin/tool/dataflows/visual.php', ['id' => $record->id, 'type' => 'png']);
-        $img = \html_writer::img($imgurl, "Dataflow #{$record->id} visualisation", ['height' => '30px', 'class' => 'lightbox']);
+        if (helper::is_graphviz_dot_installed()) {
+            $img = \html_writer::img($imgurl, "Dataflow #{$record->id} visualisation", ['height' => '30px', 'class' => 'lightbox']);
+        } else {
+            $img = $OUTPUT->render(
+                new \pix_icon(helper::GRAPHVIZ_ALT_ICON,
+                get_string('no_dot_installed_short', 'tool_dataflows'))
+            );
```
keevan, line 90: Instead of displaying just a :exclamation: , could we have supporting text? This helps with accessibility and generally is better than assuming the user will hover over the img to see the tooltip.

image

image

Perhaps something like, "Preview unavailable"

classes/visualiser.php
```diff
@@ -140,6 +147,16 @@ public static function display_dataflows_table(dataflows_table $table, \moodle_u
...
         echo $output->heading($pageheading);
 
+        if (!helper::is_graphviz_dot_installed()) {
+            \core\notification::warning(
+                get_string('no_dot_installed', 'tool_dataflows') . ' ' .
+                \html_writer::link(
+                    helper::GRAPHVIZ_DOWNLOAD_LINK,
+                    get_string('download_graphviz', 'tool_dataflows')
+                )
+            );
```
keevan, line 157: image

I am of the opinion that third party dependencies for the plugin, e.g. graphviz, should instead be referenced from the latest version of our readme, instead of directly from the error / warning.

Reasons:
- The link might become outdated or changed later on, the user will then be redirected to a broken page, even if we have properly updated it in our readme.
- The plugin readme (since it's hosted on github) should be long standing and up to date, so referencing the dependency here and where to update/download it the ideal middle layer (though I understand the simplicity of going straight to the vendor)
- There might be additional instructions for setting it up, _especially_ to work with our plugin. In this particular instance, it probably isn't too important, but I still believe it's good to have this be standard behaviour for dependencies listed.

I do think the wording can be updated, which falls under the lang string section but since it's related I'll put it here also.

Currently it is:
> No valid path to Graphviz dot found (Have you installed it?). Download Graphviz here.

The suggested version might look like so:
> A dependency "dot" could not be found. View dependencies and installation instructions **here**.

And the link would be either the [dataflows github repository](https://github.com/catalyst/moodle-tool_dataflows), or the [moodle plugins directory](http://moodle.org/plugins). Since this is unpublished, it would lean towards the former.

keevan (changes requested):
I think overall the patch looks good. There are few suggestions which should make things clearer for the user to action.

jaypha:
Screenshot from 2022-07-29 16-53-07

jaypha:
Screenshot from 2022-07-29 16-54-02

jaypha:
I decided to remove the images here altogether (when Graphviz is unavailable), because it was getting too complicated otherwise).

Screenshot from 2022-07-29 16-54-16

classes/dataflows_table.php
jaypha: New screenies follow.

classes/visualiser.php
jaypha: New screenies below.

keevan:
Thanks Jason. Those changes do seem better. I'm a bit on the fence about https://github.com/catalyst/moodle-tool_dataflows/pull/371#issuecomment-1198944324

Do you think this message as shown here: https://github.com/catalyst/moodle-tool_dataflows/pull/371#issuecomment-1198944146 should also be put on the details page as well? From a user perspective I would have liked the main "preview unavailable" shown to also mention something about a dependency being missing, even specifically "dot". There's currently no direct correlation between the two and I think there should be, for the user's awareness.

keevan (approved):

PR #375, opened 2022-07-29
fix: case matching order to evaluate cases in order

### LINKED ISSUE #347: Case matching order, iterator output checks
Currently the case step matches based on the next step that is attempting to _pull_. This actually means that regardless of order, as long as the relevant case matches, this will allow the iterator to flow to that output.

The issue is that because the flow cap pulls in a round-robin fashion, the cases matched might happen in a more unpredictable fashion, at least from a high level view of the dataflow.

Instead, it should loop through the cases in order, and stop at the first matching case OR the step that is _pulling_. In other words, if the first case always evaluates to true, and the 2nd case also always evaluates to true, then **the expected behaviour is that the data will always flow to the first case's output step.**, leaving nothing for the 2nd case.

keevan:
Previously it would push the record downstream if an out of order step's expression was truthy.

Also adjusted the help text to make this a bit clearer in how it is expected to work

Resolves #347

Example workflow:
[one_direction_20220729_0119.yml.txt](https://github.com/catalyst/moodle-tool_dataflows/files/9215239/one_direction_20220729_0119.yml.txt)
This should flow in one downstream, and store all results in one file. The other file should be empty, even though both expressions are truthy

Lang string changes:

image

No placeholder shown:

image

### DIFF
--- a/classes/local/step/flow_logic_case.php
+++ b/classes/local/step/flow_logic_case.php
@@ -108,7 +108,7 @@ class flow_logic_case extends flow_logic_step {
             'config_cases',
             get_string('flow_logic_case:cases', 'tool_dataflows'),
             [
-                'placeholder' => "label: <expression>\nsome other label: <expression>",
+                'placeholder' => "label: <expression>\nsome other label: <expression>\ndefault: 1",
                 'cols' => 50,
                 // Maxoutputs is too big for normal user. In most cases it will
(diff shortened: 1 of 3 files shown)


keevan:
Please note that the other case of "no matches" has been split into another issue - https://github.com/catalyst/moodle-tool_dataflows/issues/376 to keep things moving along.

PR #379, opened 2022-07-29
Issue #231: Escape strings for dot cli

### LINKED ISSUE #231: Make sure anything calling the dot cli to generate the visual flows, is sanitised input

petersistrom:
Resolves #231

### DIFF
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -559,12 +559,12 @@ class dataflow extends persistent {
             $styles = '';
             foreach ($finalstyles as $key => $value) {
                 // Escape all attributes correctly.
-                $value = addslashes($value);
+                $value = $this->escape_dot($value);
                 $styles .= "$key =\"$value\", ";
             }
             trim($styles);
 
-            $connection = implode('" -> "', [$srcstep->name, $deststep->name]);
+            $connection = implode('" -> "', [$this->escape_dot($srcstep->name), $this->escape_dot($deststep->name)]);
             $link = "\"{$connection}\" [$styles]";
             $connections[] = $link;
         }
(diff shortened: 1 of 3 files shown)


classes/dataflow.php
```diff
@@ -754,4 +754,14 @@ public function save_config_version() {
...
     }
+
+    /**
+     * Escape string to be passed to the dot cli
+     *
+     * @param   string $stringtoescape
+     * @return  string
+     */
+    public function escape_dot(string $stringtoescape): string {
+        return addslashes($stringtoescape);;
```
brendanheywood, line 765: double ;;

also cover this with a data provider including all the failing examples we found the other day

tests/tool_dataflows_dot_escape_test.php
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    /**
+     * Test generating the image from dot cli after escaping.
+     *
+     * @dataProvider escape_dot_dataprovider
+     * @covers ::escape_dot
+     * @param string $input
+     */
+    public function test_dot_cli(string $input) {
```
brendanheywood, line 102: your data provider has an input and an output so you should have 2 arguments here, you do not have any assertions that the generated output matches the expected output

brendanheywood: I also think it should test escape_dot directly

tests/tool_dataflows_dot_escape_test.php
```diff
@@ -0,0 +1,126 @@
...
+        $writer->name = 'writer';
+        $writer->type = 'tool_dataflows\local\step\writer_debugging';
+        $writer->depends_on([$reader->id]);
+        $dataflow->add_step($writer);
+
+        $dotscript = $dataflow->get_dotscript();
+
+        // Execute dotscript to ensure no errors are thrown.
+        $output = \tool_dataflows\visualiser::generate($dotscript, 'svg');
+        $this->assertIsString($output);
```
keevan, line 124: For an SVG check I would want to know if my steps, links, etc in full or a partial amount were included. Rather than checking the type of output.

PR #380, opened 2022-07-29
Issue #374: Fix Permitted directories bug

### LINKED ISSUE #374: Permitted dirs must not evaluate to include empty strings
Apparently exploding an empty string gives an array of one empty string.

jaypha:
Resolves #374 

Add checks to get_permitted_dirs() to exclude empty values.
Fix unit tests.
Add more unit tests
Remove empty lines before saving value.
Moved admin_setting_permitted_directories to classes/admin.

### DIFF
--- a/lang/en/tool_dataflows.php
+++ b/lang/en/tool_dataflows.php
@@ -37,7 +37,7 @@ $string['enabled_help'] = '';
 $string['permitted_dirs'] = 'Permitted directories';
 $string['permitted_dirs_desc'] = "List directories here to allow them to be read from/written to by dataflow steps.
     One directory per line. Each directory must be an absolute path. You can use the place holder '{\$a}' for the
-    site's data root directory.\n
+    site's data root directory. You can also include comments using /* and */.\n
     Examples.
     /tmp
     /home/me/mydata
--- a/tests/tool_dataflows_helper_test.php
+++ b/tests/tool_dataflows_helper_test.php
@@ -37,15 +37,42 @@ class tool_dataflows_helper_test extends \advanced_testcase {
     /**
      * Tests the get_permitted_dirs() function.
      *
+     * @dataProvider dir_provider
      * @covers \tool_dataflows\helper::get_permitted_dirs
+     * @param string $data
+     * @param array $expected
      */
-    public function test_get_permitted_dirs() {
-        global $CFG;
-
-        set_config('permitted_dirs', "/home/me/tmp\n[dataroot]/tmp", 'tool_dataflows');
+    public function test_get_permitted_dirs(string $data, array $expected) {
+        set_config('permitted_dirs', $data, 'tool_dataflows');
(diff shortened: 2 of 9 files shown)


keevan:
Also, on a loosely related note, are you able to move the admin setting under the directory `classes/admin/*` as per https://github.com/catalyst/moodle-tool_dataflows/issues/145

Related: 
https://docs.moodle.org/dev/Coding_style#Namespaces
https://docs.moodle.org/dev/Core_APIs#Admin_settings_API_.28admin.29

tests/tool_dataflows_helper_test.php
```diff
@@ -37,15 +37,42 @@ protected function setUp(): void {
...
+                '/* A comment' . PHP_EOL . ' over two lines */' . PHP_EOL . '/tmp' . PHP_EOL . '/var',
+                ['/tmp', '/var']
+            ],
+            [
+                '# comment' . PHP_EOL . '/var/tmp   # more comments.',
+                ['/var/tmp']
+            ],
+            [
+                '/var/[dataroot]/tmp',
+                ['/var/[dataroot]/tmp']
```
keevan, line 74: I'm on the fence about whether or not this should be added as a multiline string, purely for readability. Reading the PHP_EOL and mentally pushing it onto the next line is a bit painful, and you cannot glance at the test to see if there's a case it's not yet covering.

Would you be keen to adjust this? I think it should also be expected that there might be some spacing before / after each line and it should be handled gracefully, and so you should be able to get what you want with the indenation.

e.g.
```
["
    /home/me/tmp
    [dataroot]/tmp
    ...other
    # a comment
", [.. valid options here]]
```

lang/en/tool_dataflows.php
```diff
@@ -37,7 +37,7 @@
 $string['permitted_dirs'] = 'Permitted directories';
 $string['permitted_dirs_desc'] = "List directories here to allow them to be read from/written to by dataflow steps.
     One directory per line. Each directory must be an absolute path. You can use the place holder '{\$a}' for the
-    site's data root directory.\n
+    site's data root directory. You can also include comments using /* and */.\n
```
keevan, line 40: Should also mention `#` based comments

brendanheywood: +1 but let’s keep moving too

tests/tool_dataflows_helper_test.php
jaypha: Do you mean replacing PHP_EOL with hard newlines?
As for spacing before/after, the lines get trimmed before processing.

lang/en/tool_dataflows.php
jaypha: I'll submit another PR as a followup.

jaypha: This is being fixed as  a part of PR #417.

PR #384, opened 2022-08-01
Issue #372: Fix step alias validation

### LINKED ISSUE #372: step alias should not allow thing containing a .
eg `send_mdl_quiz.json_to_s3`

This could easily be ambogous if there was another step with an alias of `send_mdl_quiz`

jaypha:
Resolves #372

Add rule to alias element in step_form.
Add validate_alias() method to step.
Add unit tests.

### DIFF
--- a/classes/form/step_form.php
+++ b/classes/form/step_form.php
@@ -19,6 +19,7 @@ namespace tool_dataflows\form;
 use Symfony\Component\Yaml\Yaml;
 use tool_dataflows\dataflow;
 use tool_dataflows\parser;
+use tool_dataflows\step;
 
 /**
  * Dataflow Step Form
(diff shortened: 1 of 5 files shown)


classes/form/step_form.php
```diff
@@ -70,6 +71,7 @@ public function definition() {
         // Alias for the step (e.g. id-like field of a yaml configured dataflow, if absent, the key for the step).
         $mform->addElement('text', 'alias', get_string('field_alias', 'tool_dataflows'));
         $mform->addElement('static', 'alias_help', '', get_string('field_alias_help', 'tool_dataflows'));
+        $mform->addRule('alias', get_string('invalid_value', 'tool_dataflows'), 'regex', step::ALIAS_REGEX, 'client');
```
brendanheywood, line 74: why not just PARAM_ALPHANUMEXT ?

jaypha: Unfortunately, there are no defined constants for the regexes for PARAM_*. I changed the regex to be compatible with PARAM_ALPHANUMEXT, but I still have to provide it myself.

PR #385, opened 2022-08-01
382-side-effects:

### LINKED ISSUE #382: Stream writers should be considered to have side effects
If its output path is outside the scratch directory

Default to has side effects if unconfigured

### LINKED ISSUE #383: writer_debugging does not have side effects

jaypha:
Resolves #382 
Resolves #383 

Make having a side effect is not the default for writers.
Add has_side_effect() to writer_stream.
Set hassideeffect to true for array_out_type.
Add unit test.


brendanheywood:
lgtm, auto merging

PR #388, opened 2022-08-01
Issue #378: has_side_effect for curl and s3

### LINKED ISSUE #378: s3 connector step should only have side effects if it copies to s3
If it reads from s3 then it doesn't have any effect

jaypha:
Resolves #378

Add has_side_effect() to connector_curl and connector_s3
Add unit tests.

### DIFF
--- a/classes/local/step/connector_curl.php
+++ b/classes/local/step/connector_curl.php
@@ -30,11 +30,42 @@ use tool_dataflows\helper;
  */
 class connector_curl extends connector_step {
 
-    /** @var bool whether or not this step type (potentially) contains a side effect or not */
-    protected $hassideeffect = true;
-
     /** @var int Time after which curl request is aborted */
     protected $timeout = 60;
+
+    /**
+     * Returns whether or not the step configured, has a side effect.
+     *
+     * For curl connectors, it is considered to have a side effect if the target is
+     * anywhere outside of the scratch directory, the method is anything other than
+     * 'get' or 'head', or if the 'has side effects' setting is checked.
+     *
+     * @return     bool whether or not this step has a side effect
+     * @link https://en.wikipedia.org/wiki/Side_effect_(computer_science)
+     */
+    public function has_side_effect(): bool {
+        if (isset($this->stepdef)) {
+            $config = $this->stepdef->config;
+
+            // Destination is outside of scratch directory.
+            if (!(empty($config->destination) || helper::path_is_relative($config->destination))) {
+                return true;
+            }
+
+            // Request is anything other than 'get' or 'head'.
+            if (!($config->method == 'get' || $config->method == 'head')) {
+                return true;
+            }
+
+            // Side effects setting is checked.
+            if (!empty($config->sideeffects)) {
+                return true;
+            }
+
+            return false;
+        }
+        return true;
+    }
 
     /**
      * Return the definition of the fields available in this form.
@@ -199,13 +230,6 @@ class connector_curl extends connector_step {
         }
 
         $curl = new \curl();
-
-        $this->hassideeffect = !empty($config->sideeffects);
-
-        // For put and post methods automatically overrides/has side effects.
-        if ($config->method != 'get') {
-            $this->hassideeffect = true;
-        }
 
         // Provided a header is specified add header to request.
         if (!empty($config->headers)) {
--- a/classes/local/step/connector_s3.php
+++ b/classes/local/step/connector_s3.php
@@ -29,11 +29,25 @@ use tool_dataflows\helper;
  */
 class connector_s3 extends connector_step {
 
-    /** @var bool whether or not this step type (potentially) contains a side effect or not */
-    protected $hassideeffect = true;
-
     /** @var string the prefix identifier for an s3 path, e.g. s3://path/to/file. */
     const S3_PREFIX = 's3://';
+
+    /**
+     * Returns whether or not the step configured, has a side effect.
+     *
+     * For s3 connectors, it is considered to have a side effect if the target is
+     * anywhere outside of the scratch directory.
+     *
+     * @return     bool whether or not this step has a side effect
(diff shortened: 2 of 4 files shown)


classes/local/step/connector_curl.php
```diff
@@ -28,12 +29,36 @@
...
+     * @link https://en.wikipedia.org/wiki/Side_effect_(computer_science)
+     */
+    public function has_side_effect(): bool {
+        if (isset($this->stepdef)) {
+            $config = $this->stepdef->config;
+            return !(
+                (empty($config->destination) || helper::path_is_relative($config->destination)) &&
+                ($config->method == 'get' || $config->method == 'head') &&
+                empty($config->sideeffects)
+            );
```
keevan, line 57: This is a pretty complicated if statement. Without the description from the phpdoc above, it would be difficult to determine the intentions.

Could you perhaps split this into multiple if blocks with a comment above each describing what the check is? This would make it easier to read, check and make adjustments for. This is made complicated further by the surrounding `!( .. )` around the main condition. Performance wise it would be more or less the same (negligible) but the gain in readability is huge. 

The psuedocode suggestion is as follows:
```php
// If destination is outside the scratch directory, it has a side effect.
if ( .. ) {
    return true;
}

// If the request method is not a GET or HEAD request, it has a side effect.
if ( .. ) {
    return true;
}

// Returns whether or not the user declared this step causes side effects
return !empty($config->sideeffects)
```

Also please use strict equality checks as the method should always be a string, e.g. use `$config->method === 'get`

classes/local/step/connector_s3.php
```diff
@@ -29,12 +29,33 @@
...
+     * operation.
+     *
+     * For s3 connectors, it is considered to have a side effect if the target is
+     * anywhere outside of the scratch directory.
+     *
+     * @return     bool whether or not this step has a side effect
+     * @link https://en.wikipedia.org/wiki/Side_effect_(computer_science)
+     */
+    public function has_side_effect(): bool {
+        if (isset($this->stepdef)) {
```
keevan, line 50: This isset check feels wrong. Are you able to find out why it needs this check and whether or not it can be easily removed? The step def should always be available for the step type to reference. 

Just thinking, the fact that this is needed means any new step authors would need to always add this check, and determine whether by default it should return as true or false, when it probably should have some sort of default or unknown value.

One spot I can think of is the step chooser (where it displays the types but hasn't linked them to anything yet). I think if a step type has potential to cause a side effect, it should be default be regarded as one having a side effect in the step chooser.

Whilst this can be split out into a separate PR, this seems very closely related to the cleanup down in this PR here. Do you agree?

classes/local/step/connector_s3.php
```diff
@@ -29,12 +29,33 @@
...
+     * A side effect if it modifies some state variable value(s) outside its
+     * local environment, which is to say if it has any observable effect other
+     * than its primary effect of returning a value to the invoker of the
+     * operation.
+     *
+     * For s3 connectors, it is considered to have a side effect if the target is
+     * anywhere outside of the scratch directory.
+     *
+     * @return     bool whether or not this step has a side effect
+     * @link https://en.wikipedia.org/wiki/Side_effect_(computer_science)
```
keevan, line 47: Not sure about any particular conventions, it can probably be discussed but I feel like the extra information is just unnecessary. The purpose of the original phpdocs was so step authors could read and know what the has_side_effect function is used for.

My edit would be as follows, which would keep the main title, and @return line. Everything else is tailored for this method's implementation.
```suggestion
     * For s3 connectors, it is considered to have a side effect if the target is
     * anywhere outside of the scratch directory.
     *
     * @return     bool whether or not this step has a side effect
```

keevan (commented):
Thanks for adding those tests @jaypha.

I've added some points I've noticed, and alsso phpdoc is returning this currently if you're able to take a look:
```
admin/tool/dataflows/tests/tool_dataflows_curl_connector_test.php
    Line 236: Phpdocs for function tool_dataflows_curl_connector_test::test_has_side_effect has incomplete parameters list
```

classes/local/step/connector_curl.php
keevan: Some references:

https://github.com/jupeter/clean-code-php#avoid-nesting-too-deeply-and-return-early-part-1
https://github.com/jupeter/clean-code-php#avoid-nesting-too-deeply-and-return-early-part-2

classes/local/step/connector_s3.php
jaypha: Can you clarify this? I don't understand what you are getting at. All of the functions I have defined have default values  that get returned.

jaypha: stepdef is not guaranteed to be set. So unless there is some guarantee that has_side_effect() will not be called under such circumstances then this check will need to be there.

If we want to get around this, I could try to create a has_side_effect trait that could have a skeleton function defined, but we would then need an inner function to contain the guts of the check, and give it a suitable name.

jaypha: Or we could alter what base_step::has_side_effect() does. We would still need some kind of inner function though.

keevan: Yeah it's a bit of a pain to have to declare it as such, but perhaps it's more straightforward though a bit more cumbersome to use for now.

Okay, perhaps we can ignore this.

classes/local/step/connector_curl.php
jaypha: done

classes/local/step/connector_s3.php
jaypha: done

keevan:
phpdocs
```
admin/tool/dataflows/tests/tool_dataflows_curl_connector_test.php
    Line 276: Phpdocs for function tool_dataflows_curl_connector_test::test_has_side_effect has incomplete parameters list
```

PR #394, opened 2022-08-02
Issue #391: validate S3 connector paths only for runs.

### LINKED ISSUE #391: S3 connectors should be able to import impermitted pathnames.
The same as #335, but for S3 connectors.

jaypha:
Resolves #391 

Add validate_for_run() to connector_s3
Add unit tests

### DIFF
--- a/classes/local/step/connector_s3.php
+++ b/classes/local/step/connector_s3.php
@@ -250,20 +250,27 @@ class connector_s3 extends connector_step {
             $errors['config_target'] = $errors['config_target'] ?? $errormsg;
         }
 
-        // Confirm the non s3 source is safe to use.
-        if (empty($sourceins3) && !isset($errors['config_source'])) {
-            $error = helper::path_validate($config->source);
-            if ($error !== true) {
-                $errors['config_source'] = $error;
-            }
+        return empty($errors) ? true : $errors;
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


classes/local/step/connector_s3.php
```diff
@@ -250,20 +250,27 @@ public function validate_config($config) {
...
+     * @return true|array Will return true or an array of errors.
+     */
+    public function validate_for_run() {
+        $config = $this->stepdef->config;
+
+        $errors = [];
+
+        $error = helper::path_validate($config->source);
+        if ($error !== true) {
+            $errors['config_source'] = $error;
```
keevan, line 268: Less of a point but more of a question: Do you think there's a better variable name that could be used? Error makes sense when there is an actual error, but the conditional will more or less skip the error setting if error === true, which is strange logically speaking.

jaypha: In the later stages I was thinking of switching to 'answer', or 'result'. I can do it for this if you really want me to.

keevan (approved):

classes/local/step/connector_s3.php
keevan: I think we'll clean them all up at once later on. But thanks for the suggestions. I believe I've used $validation in other places.

PR #395, opened 2022-08-02
Issue #351: Sort flows by enabled and name

### LINKED ISSUE #351: Bug when enabling / disabling the order of flows changes
The order should always be:
* [ ] first by enabled and then disabled in two groups
* [ ] then by alphabetical, ignoring case

petersistrom:
Resolves #351

### DIFF
--- a/classes/dataflows_table.php
+++ b/classes/dataflows_table.php
@@ -73,6 +73,15 @@ class dataflows_table extends \table_sql {
     }
 
     /**
+     * Set sql sort value
+     *
+     * @return string
+     */
+    public function get_sql_sort() {
+        return 'enabled DESC, name';
+    }
+
+    /**
      * Display a small preview of the workflow
      *
      * @param \stdClass $record


jaypha:
I get an error.

`Exception - Call to protected method flexible_table::set_sorting_preferences() from context 'tool_dataflows\visualiser'`

```
Stack trace:

    line 191 of /admin/tool/dataflows/classes/visualiser.php: Error thrown
    line 63 of /admin/tool/dataflows/index.php: call to tool_dataflows\visualiser::display_dataflows_table()
```

classes/dataflows_table.php
```diff
@@ -216,6 +220,20 @@ public function col_config(\stdClass $record): string {
...
 
+    /**
+     * Display if the dataflow is enabled
+     *
+     * @param \stdClass $record
+     * @return string
+     */
+    public function col_enabled(\stdClass $record): string {
+        if ($record->enabled) {
+            return 'Y';
```
brendanheywood, line 231: This needs to be translated, so either an icon (preferred) or a yes / no

jaypha: Is this even needed? There is already is enabled/disabled icon in the action column.

petersistrom: The column is required in order for the table to sort by enabled. One option would be to move the enable/disable button from the action column and have it in this column

jaypha: OK.

jaypha:
I find that clicking on 'name' does not change the sort priority.
screenie

classes/dataflows_table.php
```diff
@@ -72,6 +72,14 @@ public function make_columns(): void {
         $this->define_headers($headers);
     }
 
+    /**
+     * Set sql sort value
+     *
+     */
+    function get_sql_sort() {
```
jaypha, line 79: It appears you forgot to put in 'public'.
You also need to add a @return to the phpdoc.

PR #405, opened 2022-08-03
Issue #376: Case step - handle no matching cases

### LINKED ISSUE #376: Case step - add handling to ensure if no expressions match, the record flowing down is effectively skipped
For example, if there was a case step, with 2 output connections/steps, and the expressions were both falsey, I would expect the iterator to be exhausted and the downstreams steps effectively do nothing.

- [ ] Add log message to indicate the record was skipped
- [ ] Add a unit test to confirm this
- [ ] One way to do this is to store a list of checked downstreams, and once the count of checked downstreams matches the number of expected downstreams (all = false), pull the next record from the iterator.

This is also to prevent an infinite loop from occurring if the case configuration didn't have a `default` case (which always evals to true)

petersistrom:
Resolves #376

### DIFF
--- a/classes/local/step/flow_logic_case.php
+++ b/classes/local/step/flow_logic_case.php
@@ -207,6 +207,7 @@ class flow_logic_case extends flow_logic_step {
                 // efficiency.
                 // See issue #347 for more details.
                 $parser = new parser;
+                $casefailures = 0;
                 foreach ($this->cases as $caseindex => $case) {
                     $result = (bool) $parser->evaluate_or_fail('${{ ' . $case . ' }}', ['record' => $value]);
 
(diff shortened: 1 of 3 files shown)


classes/local/step/flow_logic_case.php
```diff
@@ -215,6 +216,15 @@ public function next($caller) {
...
                         break;
+                    } else {
+                        $casefailures++;
+                        // Check if all cases have failed, if so pull next record.
+                        if ($casefailures === count($this->cases)) {
+                            // Log details for when a no cases matches.
+                            $this->step->log(get_string('flow_logic_case:nomatchingcases', 'tool_dataflows'));
+                            $this->input->next($this);
+                            break;
+                        }
```
keevan, line 227: This works but feels like the wrong place to place it.

In particular the next conditional will break up to the case number, so to _actually_ skip, the outside iterator needs to be on the last upstream iterator.

Feels like there should be a more eloquent way to do this, e.g. an array that gets filled up every time a case doesn't match, and perhaps an array holding all the failing cases, such that it doesn't matter where you are in the loop, as long as the failing amount matches the number of cases, it should skip. This of course would reset every time a new record is pulled through.

keevan: Optimise handling of case step skipping to skip once enough cases have been checked. Currently it will only skip if the last iterator is checked, and all matching conditions have failed.

It would be slightly better if it skips when the current iterator value has checked all conditions and found no matching cases, and this could happen for any iterator.

PR #409, opened 2022-08-03
Copy variables to clipboard #310

### LINKED ISSUE #310: Click on any variable in the available fields help chunk should put the expression into the clipboard

brendanheywood:
Resolves #310

### DIFF
--- a/templates/available-fields.mustache
+++ b/templates/available-fields.mustache
@@ -30,8 +30,38 @@
 
     Example context (json):
     {
-        "groups": {
-        }
+        "groups": [
+            {
+                "name": "env",
+                "open": "open",
+                "fields": [
+                    {
+                        "text": "level2a",
+                        "expression": "env.level2a",
+                        "fields": [],
+                        "leaf": 1
+                    },
+                    {
+                        "text": "level2b",
+                        "expression": "env.level2b",
+                        "fields": [
+                            {
+                                "text": "level3a",
+                                "expression": "env.level2b.level3a",
+                                "fields": [],
+                                "leaf": 1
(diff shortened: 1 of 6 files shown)


brendanheywood:
Peek 2022-08-03 16-28

templates/available-fields.mustache
```diff
@@ -48,3 +48,22 @@
...
+                document.execCommand("copy");
+
+                addToast.add(getString.get_string('successfullycopiedtoclipboard', 'tool_dataflows', text.value), {
+                    type: 'info',
+                    autohide: true,
+                    closeButton: true,
+                });
+            });
+        }
+    });
```
keevan, line 68: Ah this is nice. I'll go through some points:

execCommand is deprecated, please consider using the clipboard API and writeText (better browser support)
- https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand
- https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API

Loving the use of const.

`for (let i = 0; i < expression.length; i++) {`

I would much prefer the usage of `expressions.forEach(expression => { ... })`

Also `const expression` should probably be `const expressions` since it's getting all matching leaf expressions.

`getString.get_string` .. hahahah that just sounds funny. Is this typically how these core classes are used? If I had to start from nothing except the imports, I would probably use `toast` instead of `addToast`, and str instead of `getString`.

What do you think?

brendanheywood: es6 isn't transpiled here as it is duped verbatim and doesn't go through grunt, so for the moment it will be a little old school

Issue #411, opened 2022-08-03
position passed to get_output_label happens to be null and throws errors


marcghaly:
![Screenshot from 2022-08-03 09-25-28](https://user-images.githubusercontent.com/76186988/182619221-4a13970e-312c-472a-9fc2-8464c4073791.png)

PR #412, opened 2022-08-03
fix: output label position error

marcghaly:
Hi @keevan, this resolves issue #411

Regards,

Marc


keevan:
I'm more interested in why it's null in the first place, and fixing that issue instead. You hadn't mentioned in the issue, but how were you able to produce this particular problem?

keevan:
Going to hijack this one since I encountered this myself during testing and fixed the root cause.

https://github.com/catalyst/moodle-tool_dataflows/pull/426/files

PR #417, opened 2022-08-03
Issue #354: Global variables and config settings

### LINKED ISSUE #354: Support global vars
These are variables that are defined in an admin setting and are available to all flows. The key business reasons you would want this are:
1) to reuse a single var which might changes across many flows, such as the base domain of external api
2) you can import and export flows to different environments and wire them differently without the flow itself needing to change
3) any secrets could be declared globally and so would not be including in the flow when exporting

* [x] new admin textarea with name: expression list of vars. These can be nested by having a prefix with dots eg
```
foo: bar
acme.api.base: https://acme.com/api
acme.api.secret: 12345
```
* [x] vars would be like `${{ global.vars.acme.api.secret}}` (needs to work side by side with #366)
* [x] help text explaining why and when to use it
* [x] add any defined vars to the variable help tree widget
* [x] need to be able to support values which can contain new lines. yaml supports this with this syntax

```
example: |\n
··Several lines of text,\n
··with some "quotes" of various 'types',\n
··and also a blank line:\n
··\n
··and some text with\n
····extra indentation\n
··on the next line,\n
··plus another line at the end.\n
```

Out of scope in followup issues:
* https://github.com/catalyst/moodle-tool_dataflows/issues/400

jaypha:
Resolves #354 

Add global vars box to plugin settings.
Add global.vars to variables tree.

### DIFF
--- a/classes/admin/admin_setting_yaml.php
+++ b/classes/admin/admin_setting_yaml.php
@@ -1,5 +1,5 @@
 <?php
-// This file is part of Moodle - http://moodle.org/
+// This file is part of Moodle - https://moodle.org/
 //
 // Moodle is free software: you can redistribute it and/or modify
 // it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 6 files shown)


brendanheywood:
Looks great, automerging

PR #419, opened 2022-08-04
fix: step to be extended from flow_step to correct output links

### LINKED ISSUE #418: Web service not extending from correct class
Results in numbered / labelled outputs when it should just be 1-1

keevan:
Resolves #418

### DIFF
--- a/classes/local/step/flow_web_service.php
+++ b/classes/local/step/flow_web_service.php
@@ -29,7 +29,7 @@ use tool_dataflows\parser;
  * @copyright  Catalyst IT, 2022
  * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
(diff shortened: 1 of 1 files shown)


classes/local/step/flow_web_service.php
```diff
@@ -29,7 +29,7 @@
  * @copyright  Catalyst IT, 2022
  * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
-class flow_web_service extends flow_logic_step {
+class flow_web_service extends flow_step {
```
brendanheywood, line 32: This makes the step disappear completely from the chooser? purged caches and still no dice

brendanheywood: made https://github.com/catalyst/moodle-tool_dataflows/issues/425

PR #421, opened 2022-08-04
Implement a noop step #44

### LINKED ISSUE #44: Have a 'null' or noop / no-op step which is just a placeholder
Have a 'null' step which is just a placeholder for some external task but we want to see it in the overall workflow so it is easier to reason about.

This could also be used in templates which you import and are designed so that you clearly have a thing that needs replacing. So some null steps might be valid, but others might be invalid and require you to swap them out with something proper. So might be two step types.

brendanheywood:
Unlike the debugging this step this does absolutely nothing with no
logs.

Resolves  #44

### DIFF
--- a/lang/en/tool_dataflows.php
+++ b/lang/en/tool_dataflows.php
@@ -105,6 +105,7 @@ $string['step_name_connector_sns_notify'] = 'AWS SNS Notification';
 $string['step_name_connector_wait'] = 'Wait';
 $string['step_name_flow_logic_case'] = 'Case';
 $string['step_name_flow_logic_join'] = 'Join';
+$string['step_name_flow_logic_noop'] = 'No-OP';
 $string['step_name_flow_transformer_filter'] = 'Filter transformer';
 $string['step_name_reader_json'] = 'JSON reader';
 $string['step_name_reader_sql'] = 'SQL reader';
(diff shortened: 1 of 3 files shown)


lang/en/tool_dataflows.php
```diff
@@ -105,6 +105,7 @@
 $string['step_name_connector_wait'] = 'Wait';
 $string['step_name_flow_logic_case'] = 'Case';
 $string['step_name_flow_logic_join'] = 'Join';
+$string['step_name_flow_logic_noop'] = 'No-OP';
```
keevan, line 108: You'll need to bump the version.php here.

classes/local/step/flow_logic_noop.php
```diff
@@ -0,0 +1,55 @@
...
+     * @var int[] number of input flows (min, max)
+     */
+    protected $inputflows = [1, 1];
+
+    /**
+     * For this step, it should have 2 or more inputs and for now, up to 20.
+     *
+     * @var int[] number of output flows (min, max)
+     */
+    protected $outputflows = [1, 1];
```
keevan, line 41: Once you change it to a flow step these can go. By default they should be already set to 1:1, and you won't have to update the descriptions.

lang/en/tool_dataflows.php
brendanheywood: done

PR #429, opened 2022-08-04
Issue #366: CFG vars

### LINKED ISSUE #366: Allow access to selected global config variables such as wwwroot, primary user email, support email etc
This would be available under vars something like

```
global.cfg.wwwroot
global.cfg.fullname
global.cfg.shortname
global.cfg.supportname
global.cfg.supportemail
global.cfg.supportpage
...
```

All keys should align 1-1 with the config values. This primary use case here is common elements that you would want to want to put into a message or email or notification eg #6

jaypha:
Add CFG list box to dataflows settings.
Add global.cfg to variables.

### DIFF
--- a/classes/admin/admin_setting_cfg_list.php
+++ b/classes/admin/admin_setting_cfg_list.php
@@ -1,5 +1,5 @@
 <?php
-// This file is part of Moodle - http://moodle.org/
+// This file is part of Moodle - https://moodle.org/
 //
 // Moodle is free software: you can redistribute it and/or modify
 // it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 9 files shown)


jaypha:
A nice screenie.

Screenshot from 2022-08-04 17-14-46

classes/admin/admin_setting_cfg_list.php
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * A list of $CFG settings.
+ *
+ * @package   tool_dataflows
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class admin_setting_cfg_list extends admin_setting_list {
```
brendanheywood, line 27: I wasn't clear in the issue, but my intention was that the list of cfg's you could access would be hard coded. I'm on the fence around this, there is utility in grabbing any cfg, but there is a huge risk here with grabbing certain ones, like this:

image

The main reason I wanted these in the first place was for making links and for bits which would be used in emails that a flow might generate and 95% of this is covered by a handful of the cfg items listed in the issue.

So more eyeballs wanted. If we can make this configurable but lock it down then that is ideal. One thing we could do is only accept cfg items which are exposed in the admin tree, so anything else which is config.php only cannot work. The other downside to this is it will be slow to validate it, but it is a once off validation.

The other consideration might also be admin settings which are secrets and which are forced in config.php are generally not visible to admins.

@Peterburnett in particular want you security perspective on this one.

jaypha: I would presume it would set as a forced config value.

brendanheywood: I am also ok with merging this as is as long as we can see a path forward to hardening this in another issue

settings.php
```diff
@@ -82,6 +82,14 @@
             '',
             PARAM_RAW
         ));
+
+        $settings->add(new admin_setting_cfg_list(
+            'tool_dataflows/config_vars',
+            get_string('config_vars', 'tool_dataflows'),
+            get_string('config_vars_desc', 'tool_dataflows'),
+            '',
```
brendanheywood, line 90: I would default this to the ones listed in the issue

jaypha:
This one has been put on the backburner

PR #430, opened 2022-08-04
feat: add more helper functions to the parser #413

### LINKED ISSUE #413: Have an array count expression function
ie if we have an array of stuff, then we should be able to get the count of that array into another var

* [ ] implement a count()  function
* [ ] implement an isset() function - ensure this works when the expression inside it may only partially evaulate, ie half way through it is null so it can't get to the child parts

keevan:
Added native php functions as per below:
- add count
- add isset

Also added supporting tests

Resolves #413

### DIFF
--- a/tests/tool_dataflows_variables_test.php
+++ b/tests/tool_dataflows_variables_test.php
@@ -323,4 +323,54 @@ class tool_dataflows_variables_test extends \advanced_testcase {
         $result = $expressionlanguage->evaluate("steps.{$stepdef->name}.customOutputKey", $variables);
         $this->assertEquals('H5P.Accordion', $result);
     }
+
+    /**
+     * Test the functions available through the parser
+     *
+     * @param  string $expression
+     * @param  array $variables
+     * @param  mixed $expected
+     *
+     * @covers        \tool_dataflows\parser
+     * @dataProvider  parser_functions_data_provider
+     */
+    public function test_parser_functions(string $expression, array $variables, $expected) {
+        $parser = new parser;
+        $result = $parser->evaluate('${{'.$expression.'}}', $variables);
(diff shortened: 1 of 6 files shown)


tests/tool_dataflows_variables_test.php
```diff
@@ -323,4 +323,33 @@ public function test_output_variables() {
...
+    }
+
+    /**
+     * Ensure these expressions return the expected values (happy path)
+     *
+     * @return array of data
+     */
+    public function parser_functions_data_provider() {
+        return [
+            ['isset(a[0])', ['a' => [3, 2, 1]], true],
```
brendanheywood, line 350: it would be good to flesh this out with some of the negative examples too and both . as well as array data

keevan: Yeah I thought about that. Negatives will just return the expression by default, but and I thought I'd test the happy path this time and add negative tests in a different provider + test. I don't like seeing a messy data provider.

brendanheywood: The entire point of implementing this in the first place is so we can do something like:

isset(grand.daddy.child.mother[2].goat) 

This must always return true or false

keevan: Ah, that's what you meant by negatives. Then yes I agree it should test for absence as well as when it does have it. 

Apologies, I thought you meant when isset wasn't included as a negative example.

keevan: I added a bit more code and testing, and they are all proving unfruitful. It seems the value (e.g. `isset(value)`) is being evaluated first, and the parser stops before isset even gets a chance to give it a go.

PR #438, opened 2022-08-04
Email connector action

### LINKED ISSUE #6: Connector step emit email
Title, to, cc, body are all mustache templates and / or atto editor with placeholders expressions

Some vars:
* rows process from source
* rows processed at this loader
* time started
* time finished
* other variables in the "global" scope of the stream rather than each item

### LINKED ISSUE #410: Make a new var which holds a deep link to the run instance summary page
Split from https://github.com/catalyst/moodle-tool_dataflows/issues/408

brendanheywood:
Resolves #410 
Resolves #6

### DIFF
--- a/classes/local/step/connector_email.php
+++ b/classes/local/step/connector_email.php
@@ -1,0 +1,127 @@
+<?php
+// This file is part of Moodle - https://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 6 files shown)


classes/local/step/connector_email.php
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        $config = $this->enginestep->stepdef->config;
+
+        $toemail = $config->to;
+        $to = (object)array(
+            'id' => 1,
+            'auth' => 'manual',
+            'email' => $toemail,
+            'username' => 'xxx',
```
brendanheywood, line 78: Need to fix this

keevan: On that note, does this support multiple addresses? e.g. if it wasn't for a single user but for a fixed number of predefined users instead (which wasn't sourced in Moodle itself)?

brendanheywood: No and at this point I'm happy for that to be out of scope and tackled later on

brendanheywood:
Example email flow:

[Send_some_emails!_20220805_1443.yml.txt](https://github.com/catalyst/moodle-tool_dataflows/files/9269626/Send_some_emails._20220805_1443.yml.txt)

Issue #416, opened 2022-08-03
Case step : reference mapped outputs / aliases


marcghaly:
At the moment case step can only reference record passed through steps, it should also be able to reference a mapped output value, for example `record.USERNAME == john && steps.get_users_ws.userid !== null`

![Screenshot from 2022-08-03 13-39-10](https://user-images.githubusercontent.com/76186988/182673560-3f188ea4-2155-4e73-a9a1-3dee0848fd1d.png)

marcghaly:
A part of this issue has been fixed with issue #454, part concerning nullifying expression is on hold and being worked around with count and isset

PR #440, opened 2022-08-04
Fix: nullify literal expression when not assessed - be able to evalut…

marcghaly:
Hi,

This issue resolves issue #416, it allows evaluating mapped outputs in case comparison as such

Screenshot from 2022-08-04 16-48-14

Regards,

Marc

### DIFF
--- a/classes/local/step/base_step.php
+++ b/classes/local/step/base_step.php
@@ -124,6 +124,16 @@ abstract class base_step {
             $parser = new parser;
             $allvariables = array_merge($this->stepdef->variables, $this->variables);
             $outputs = $parser->evaluate_recursive($config->outputs, $allvariables);
+            // Check that outputs have been assessed.
+            foreach ($outputs as $key => $output) {
+                // If type is an object it means it has been assessed.
+                if (!is_object($output)) {
+                    [$hasexpression] = $parser->has_expression($output);
+                    if ($hasexpression) {
+                        $outputs->$key = null;
+                    }
+                }
+            }
             $this->stepdef->set_output($outputs);
         }
     }
--- a/classes/local/step/flow_logic_case.php
+++ b/classes/local/step/flow_logic_case.php
@@ -192,6 +192,7 @@ class flow_logic_case extends flow_logic_step {
                     $this->passed = false;
                 }
                 $value = $this->input->current();
+                $steps = $caller->step->engine->get_variables()['steps'];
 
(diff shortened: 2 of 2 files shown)


classes/local/step/flow_logic_case.php
```diff
@@ -192,6 +192,7 @@ public function next($caller) {
                     $this->passed = false;
                 }
                 $value = $this->input->current();
+                $steps = $caller->step->engine->get_variables()['steps'];
```
keevan, line 195: Why not allow for all variables?
```suggestion
                $variables = $caller->step->engine->get_variables();
                $variables['record'] = $value;
```
And then pass this through instead.

classes/local/step/base_step.php
```diff
@@ -124,6 +124,16 @@ public function prepare_outputs() {
             $parser = new parser;
             $allvariables = array_merge($this->stepdef->variables, $this->variables);
             $outputs = $parser->evaluate_recursive($config->outputs, $allvariables);
+            // Check that outputs have been assessed.
+            foreach ($outputs as $key => $output) {
+                // If type is an object it means it has been assessed.
+                if (!is_object($output)) {
+                    [$hasexpression] = $parser->has_expression($output);
+                    if ($hasexpression) {
+                        $outputs->$key = null;
```
keevan, line 133: I would disagree on this approach. I've added `count` and `isset` variable expression functions which could be used  instead. 

Please read the unit tests to see some examples and limitations: 
https://github.com/catalyst/moodle-tool_dataflows/blob/381bacb/tests/tool_dataflows_variables_test.php#L348

keevan: I believe the output set, should always be some value because otherwise how can proceeding steps utilise it confidently?. 

It it cannot be resolved at this stage, it should fail / explode.

I'm always wary of potentially outputs which are building blocks for more expressions e.g. `select ${{ some.value }}` but I don't think we should support that at all here.

brendanheywood:
As mentioned elsewhere, lets pause work on this one, we should have a viable workaround using the count() function expression so we can handle the first case without relying on this.

classes/local/step/flow_logic_case.php
jaypha: And perhaps call it $stepsvariables to not confuse them with actual steps.

keevan: @jaypha There wouldn't be any more `$steps` variable so the suggestion is not relevant here?

PR #444, opened 2022-08-05
Issue #366: Add CFG settings as variables.

### LINKED ISSUE #366: Allow access to selected global config variables such as wwwroot, primary user email, support email etc

jaypha:
Resolves #366

### DIFF
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -234,9 +234,12 @@ class dataflow extends persistent {
 
         // Test reading a value directly.
         $variables = [
-            'global' => (object) ['vars' => $globalvars],
-            'dataflow'  => $dataflow,
-            'steps'     => $steps,
+            'global' => (object) [
+                'vars' => $globalvars,
+                'cfg'  => helper::get_cfg_vars(),
+            ],
+            'dataflow' => $dataflow,
+            'steps'    => $steps,
         ];
         return $variables;
     }
--- a/classes/helper.php
+++ b/classes/helper.php
@@ -41,6 +41,16 @@ class helper {
     /** Location of the dependency section of the readme. */
     public const README_DEPENDENCY_LINK = 'https://github.com/catalyst/moodle-tool_dataflows#dependencies';
 
+    /** CFG and SITE settings to be included as variables. */
+    public const CFG_VARS = [
+        'fullname',
+        'shortname',
+        'supportemail',
+        'supportname',
+        'supportpage',
+        'wwwroot',
+    ];
+
     /** @var null|bool Is this Windows?  */
     protected static $iswindows = null;
 
@@ -206,5 +216,24 @@ class helper {
 
         // Don't know where dot is, so look for it.
         return !empty(shell_exec('which dot'));
+    }
+
+    /**
+     * Creates a list of variables from CFG and SITE settings.
+     *
+     * @return \stdClass
+     */
+    public static function get_cfg_vars(): \stdClass {
+        global $CFG, $SITE;
+
+        $configvars = new \stdClass();
+        foreach (self::CFG_VARS as $name) {
+            if (property_exists($CFG, $name)) {
+                $configvars->{$name} = $CFG->{$name};
+            } else if (property_exists($SITE, $name)) {
+                $configvars->{$name} = $SITE->{$name};
+            }
+        }
+        return $configvars;
     }
 }


classes/helper.php
```diff
@@ -207,4 +217,23 @@ public static function is_graphviz_dot_installed(): bool {
...
+        global $CFG, $SITE;
+
+        $configvars = new \stdClass();
+        foreach (self::CFG_VARS as $name) {
+            if (property_exists($CFG, $name)) {
+                $configvars->{$name} = $CFG->{$name};
+            } else if (property_exists($SITE, $name)) {
+                $configvars->{$name} = $SITE->{$name};
+            }
+        }
```
keevan, line 236: I wonder if it's better to have cfg values under `global.cfg.*` and site values under `global.site.*`, purely to know where to look and set the value if it wasn't already set or needs to be corrected.

classes/dataflow.php
```diff
@@ -234,9 +234,9 @@ public function get_variables(): array {
 
         // Test reading a value directly.
         $variables = [
-            'global' => (object) ['vars' => $globalvars],
-            'dataflow'  => $dataflow,
-            'steps'     => $steps,
+            'global'   => (object) ['vars' => $globalvars, 'cfg' => helper::get_cfg_vars()],
```
keevan, line 237: Could you split this into multiple lines so it's easier to see what's included? (Multi key arrays with a key per line)

classes/helper.php
jaypha: fullname and shortname are front page settings, so maybe `global.frontpage.*`?

keevan: Hmm, `supportemail` and `supportname` might be under `$CFG`, but that doesn't mean it should necessarily live under `global.support.email` and `global.support.name`.

My suggestion was to keep things more closely matching where it came from. I'm not sure if there are ever any overlap (e.g. CFG has an option with a key that SITE also has and now you have a conflict of interest, and if there are never instances of this, I would be fine with bundling them together.

classes/helper.php
```diff
@@ -41,6 +41,16 @@ class helper {
...
     public const README_DEPENDENCY_LINK = 'https://github.com/catalyst/moodle-tool_dataflows#dependencies';
 
+    /** CFG and SITE settings to be included as variables. */
+    public const CFG_VARS = [
+        'wwwroot',
+        'fullname',
+        'shortname',
+        'supportname',
+        'supportemail',
+        'supportpage',
```
keevan, line 51: Could you also sort this list alphabetically?

classes/helper.php
jaypha: I never said anything about supportemail or supportname. I suggested frontpage for fullname and shortname because that's what the user sees.

Anyway, I'm not so sure about this at all. The original specs said to put them under .cfg, and that's what I've done.

classes/dataflow.php
jaypha: done

classes/helper.php

classes/dataflow.php
```diff
@@ -234,9 +234,12 @@ public function get_variables(): array {
 
         // Test reading a value directly.
         $variables = [
-            'global' => (object) ['vars' => $globalvars],
-            'dataflow'  => $dataflow,
-            'steps'     => $steps,
+            'global' => (object) [
```
brendanheywood, line 237: one downside of the (object) [ casting is that I think you lose the explicit order of keys. Can you have a play to see if you can make this always be cfg then vars?

This isn't critical here, but for instance the key order under steps is correctly explicit

classes/helper.php
brendanheywood: Lets keep this simple and namespaced exactly the way core does which is no namespace. It would only have a namespace if we add plugin config into the brew which might look like global.cfg.tool.heartbeat.foo

classes/helper.php
```diff
@@ -207,4 +217,23 @@ public static function is_graphviz_dot_installed(): bool {
...
+    /**
+     * Creates a list of variables from CFG and SITE settings.
+     *
+     * @return \stdClass
+     */
+    public static function get_cfg_vars(): \stdClass {
+        global $CFG, $SITE;
+
+        $configvars = new \stdClass();
+        foreach (self::CFG_VARS as $name) {
```
brendanheywood, line 230: This seems to be putting them in reverse alpha order, can you make them alpha

image

brendanheywood: See also this bug probably the same root cause

https://github.com/catalyst/moodle-tool_dataflows/issues/446

jaypha: Most likely is. I noticed that it displays in the opposite order that it was given. It should be addressed as part of $446.

classes/dataflow.php
jaypha: I can look into this, but it will require a separate issue to fix.

jaypha: I had a quick look , and it appears that type casting to an object does preserve the order.

PR #445, opened 2022-08-05
Issue #353: Add dataflow config help and allow nested variables

### LINKED ISSUE #353: Polish dataflow configuration field
This field needs to be clearer what it is and how it is used.

* [ ] help text with examples of how to define vars and then how they can be used
* [ ] what the point of this field is
* [ ] add vars in here to the variable tag data
* [ ] if a var is used which is not defined need to clearly show what var is missing and where it should be set

petersistrom:
Resolves #353

### DIFF
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -262,13 +262,7 @@ class dataflow extends persistent {
 
         // Prepare this as a php object (stdClass), as it makes expressions easier to write.
         $parser = new parser();
-        foreach ($yaml as &$string) {
-            // TODO: Perhaps some $key should not be evaluated?
-
-            $string = $parser->evaluate_recursive($string, $this->variables);
-        }
-
-        return $yaml;
+        return $parser->evaluate_recursive($yaml, $this->variables);
     }
 
     /**
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -265,8 +265,7 @@ class dataflow extends persistent {
         foreach ($yaml as &$string) {
             // TODO: Perhaps some $key should not be evaluated?
 
-            // NOTE: This does not support nested expressions.
-            $string = $parser->evaluate($string, $this->variables);
+            $string = $parser->evaluate_recursive($string, $this->variables);
         }
 
         return $yaml;
--- a/classes/form/dataflow_form.php
+++ b/classes/form/dataflow_form.php
@@ -71,9 +71,11 @@ class dataflow_form extends \core\form\persistent {
             get_string('field_config', 'tool_dataflows'),
             ['cols' => 50, 'rows' => 7]
         );
-
-        $mform->addElement('static', 'config_help', '',\html_writer::nonempty_tag('pre',
-                get_string('field_config_help', 'tool_dataflows')));
+        $outputsexample['reference'] = \html_writer::tag('code', ' ${{dataflows.config.&lt;name&gt;}}');
+        $outputsexample['example'] = \html_writer::tag('code', ' ${{dataflows.config.abc.def}}');
+        $mform->addElement('static', 'config_help', '',
+        \html_writer::nonempty_tag('pre',
+        get_string('field_config_help', 'tool_dataflows', $outputsexample)));
 
         $this->add_action_buttons();
     }
--- a/classes/form/dataflow_form.php
+++ b/classes/form/dataflow_form.php
@@ -72,6 +72,9 @@ class dataflow_form extends \core\form\persistent {
             ['cols' => 50, 'rows' => 7]
         );
 
+        $mform->addElement('static', 'config_help', '',\html_writer::nonempty_tag('pre',
+                get_string('field_config_help', 'tool_dataflows')));
+
         $this->add_action_buttons();
     }
 }
--- a/classes/form/dataflow_form.php
+++ b/classes/form/dataflow_form.php
@@ -71,11 +71,13 @@ class dataflow_form extends \core\form\persistent {
             get_string('field_config', 'tool_dataflows'),
             ['cols' => 50, 'rows' => 7]
         );
-        $outputsexample['reference'] = \html_writer::tag('code', ' ${{dataflows.config.&lt;name&gt;}}');
-        $outputsexample['example'] = \html_writer::tag('code', ' ${{dataflows.config.abc.def}}');
-        $mform->addElement('static', 'config_help', '',
-        \html_writer::nonempty_tag('pre',
-        get_string('field_config_help', 'tool_dataflows', $outputsexample)));
+        $outputsexample['reference'] = \html_writer::tag('code',  htmlentities('${{dataflow.config.<name>}}'));
+        $examplestring = <<<'EOT'
+data:
+  configvar: 12     # Accessed as ${{dataflow.config.data.configvar}}
+EOT;
+        $outputsexample['example'] = \html_writer::tag('pre', $examplestring);
+        $mform->addElement('static', 'config_help', '', get_string('field_config_help', 'tool_dataflows', $outputsexample));
 
         $this->add_action_buttons();
     }
--- a/lang/en/tool_dataflows.php
+++ b/lang/en/tool_dataflows.php
@@ -138,11 +138,7 @@ $string['field_description'] = 'Description';
 $string['field_alias'] = 'Alias';
 $string['field_alias_help'] = 'A reference to this step, unique to this dataflow. This can be used in expressions to access the step. It must be made up of only letters, numbers, or underscores (\'_\'). If empty, then a snake case version of the name will be used. For example, if the name is "My Step", it will be populate this field as "my_step"';
 $string['field_config'] = 'Configuration';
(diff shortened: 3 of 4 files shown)


classes/form/dataflow_form.php
```diff
@@ -72,6 +72,9 @@ public function definition() {
             ['cols' => 50, 'rows' => 7]
         );
 
+        $mform->addElement('static', 'config_help', '',\html_writer::nonempty_tag('pre',
+                get_string('field_config_help', 'tool_dataflows')));
```
keevan, line 76: Could you reduce the indentation here by 4 spaces? You only need to use 8 indentations when wrapping some conditionals, functions, and class declarations

keevan: Alternatively, split each param into it's own line.

If you want to keep it though, you might want to fix up the spacing issue here (note the comma):
`config_help', '',\html_writer::`

classes/dataflow.php
```diff
@@ -265,8 +265,7 @@ protected function get_config($expressions = true): \stdClass {
         foreach ($yaml as &$string) {
             // TODO: Perhaps some $key should not be evaluated?
```
brendanheywood, line 266: is this comment still valid?

lang/en/tool_dataflows.php
```diff
@@ -138,6 +138,11 @@
 $string['field_alias'] = 'Alias';
 $string['field_alias_help'] = 'A reference to this step, unique to this dataflow. This can be used in expressions to access the step. It must be made up of only letters, numbers, or underscores (\'_\'). If empty, then a snake case version of the name will be used. For example, if the name is "My Step", it will be populate this field as "my_step"';
 $string['field_config'] = 'Configuration';
+$string['field_config_help'] = "Config variables that can be accessed via expressions within all dataflow steps as \${{dataflows.config.&lt;name&gt;}}.
+    Values are defined in YAML format, with nested values being converted into a dot separated sequence.\n
+    Example.
+    abc:
```
brendanheywood, line 144: whereever possible I'd like to avoid code samples in the string, they should be either hard coded outside the string, or hardcoded and passed in as a param

Also wrap them in <pre> or <code>

classes/dataflow.php
petersistrom: Turns out this for loop is redundant when calling evaluate_recursive(), so removed

classes/form/dataflow_form.php
```diff
@@ -71,6 +71,11 @@ public function definition() {
             get_string('field_config', 'tool_dataflows'),
             ['cols' => 50, 'rows' => 7]
         );
+        $outputsexample['reference'] = \html_writer::tag('code', ' ${{dataflows.config.&lt;name&gt;}}');
+        $outputsexample['example'] = \html_writer::tag('code', ' ${{dataflows.config.abc.def}}');
+        $mform->addElement('static', 'config_help', '',
+        \html_writer::nonempty_tag('pre',
+        get_string('field_config_help', 'tool_dataflows', $outputsexample)));
```
keevan, line 78: Few questions:
- What's the benefit in wrapping the whole help in a pre tag? Shouldn't only the relevant parts of it (e.g. the code examples) be?
- For the `&lt;` and `&gt;` you can use htmlentities on the whole example if you wanted instead with `<name>` and it should just work. (Fun fact, the opposite is called html_entity_decode which is in snake case. PHP consistencies)
- As mentioned by Brendan I would pull out all the examples from the lang string and prepare them e.g. here.

classes/dataflow.php
```diff
@@ -262,14 +262,7 @@ protected function get_config($expressions = true): \stdClass {
...
         $parser = new parser();
-        foreach ($yaml as &$string) {
-            // TODO: Perhaps some $key should not be evaluated?
-
-            // NOTE: This does not support nested expressions.
-            $string = $parser->evaluate($string, $this->variables);
-        }
-
-        return $yaml;
+        return $parser->evaluate_recursive($yaml, $this->variables);
```
keevan, line 265: Nice refactor :)

keevan (commented):
Also, since you are updating lang/en you'll need to bump the version.php

classes/form/dataflow_form.php
petersistrom: The pre tag is used to render the indentation and line breaks, is there a better way handle that?

keevan: Please use the pre tag only on the examples you supply into the string as per Brendan's suggestion here https://github.com/catalyst/moodle-tool_dataflows/pull/445#discussion_r938387290

classes/form/dataflow_form.php
```diff
@@ -71,6 +71,12 @@ public function definition() {
             get_string('field_config', 'tool_dataflows'),
             ['cols' => 50, 'rows' => 7]
         );
+        $outputsexample['reference'] = \html_writer::tag('code',  htmlentities('${{dataflows.config.<name>}}'));
+        $outputsexample['example'] = \html_writer::tag('pre', ' 
+        abc: 
+            def: 12          # Accessed as ${{dataflows.config.abc.def}}');
+        $mform->addElement('static', 'config_help', '',
+        get_string('field_config_help', 'tool_dataflows', $outputsexample));
```
keevan, line 79: Could you also re-indent this so it's easier to read?

classes/form/dataflow_form.php
```diff
@@ -71,6 +71,12 @@ public function definition() {
             get_string('field_config', 'tool_dataflows'),
             ['cols' => 50, 'rows' => 7]
         );
+        $outputsexample['reference'] = \html_writer::tag('code',  htmlentities('${{dataflows.config.<name>}}'));
+        $outputsexample['example'] = \html_writer::tag('pre', '
+        abc:
```
brendanheywood, line 76: IS this indenting correct? looks too deep and it should be the indenting 'inside' the textarea not the indenting of the surrounding code here, ie use heredoc / nowdoc instead

lang/en/tool_dataflows.php
```diff
@@ -138,6 +138,8 @@
 $string['field_alias'] = 'Alias';
 $string['field_alias_help'] = 'A reference to this step, unique to this dataflow. This can be used in expressions to access the step. It must be made up of only letters, numbers, or underscores (\'_\'). If empty, then a snake case version of the name will be used. For example, if the name is "My Step", it will be populate this field as "my_step"';
 $string['field_config'] = 'Configuration';
+$string['field_config_help'] = 'Config variables that can be accessed via expressions within all dataflow steps as {$a->reference}.
+    Values are defined in YAML format, with nested values being converted into a dot separated sequence. Example:{$a->example}';
```
brendanheywood, line 142: is the line break intended here? it is ignore in the ui in hml

classes/form/dataflow_form.php
```diff
@@ -71,6 +71,12 @@ public function definition() {
...
+            def: 12          # Accessed as ${{dataflows.config.abc.def}}');
```
brendanheywood, line 77: When you cut and paste this example into the field and save it then it explodes,

a) it shouldn't explode, at worst it should be a validation error
b) the example should be a perfect example which is valid

```
Unable to parse at line 1 (near " abc:").

File: classes/form/dataflow_form.php [hunk 10]
```diff
@@ -71,6 +71,14 @@ public function definition() {
...
             ['cols' => 50, 'rows' => 7]
         );
+        $outputsexample['reference'] = \html_writer::tag('code',  htmlentities('${{dataflows.config.<name>}}'));
+        $examplestring = <<<'EOT'
+data:
+  configvar: 12     # Accessed as ${{dataflows.config.data.configvar}}
+EOT;
+        $outputsexample['example'] = \html_writer::tag('pre', $examplestring);
+        $mform->addElement('static', 'config_help', '',
+                get_string('field_config_help', 'tool_dataflows', $outputsexample));
```
keevan, line 81: So I'm not sure what's happening here, but are you able to use 4 spaces for indentation?

```suggestion
            get_string('field_config_help', 'tool_dataflows', $outputsexample));
```

keevan: Potentially it might be a case where setting this on a single line is fine also?

classes/form/dataflow_form.php
```diff
@@ -71,6 +71,13 @@ public function definition() {
             get_string('field_config', 'tool_dataflows'),
             ['cols' => 50, 'rows' => 7]
         );
+        $outputsexample['reference'] = \html_writer::tag('code',  htmlentities('${{dataflows.config.<name>}}'));
```
brendanheywood, line 74: 2 types, 
dataflows.config
vs
dataflow.config  <<- no s

classes/form/dataflow_form.php
```diff
@@ -71,6 +71,13 @@ public function definition() {
...
+        $examplestring = <<<'EOT'
+data:
+  configvar: 12     # Accessed as ${{dataflows.config.data.configvar}}
```
keevan, line 77: Just in case this one was missed also.

Issue #381, opened 2022-07-29
Improve the logging outputs and appenders


brendanheywood:
There is a few things I'd like to see improved and they all interrelate:

* [x] I want to be able to have the engine and each step declare a log level for each message like debug, warn, error, critical etc and these will map to similar concepts in moodle, log4php, psr logs, etc
* [x] When running a flow you'll be able to set the log level depending on how much data you want to see. This can be set in the run now page as well as at the flow level
* [x] I want each message to be able to push both a free form string message as well as a structured data objects, usually this will be the state of the record which is flowing
* [x] I want there to be a concept similar to log4php of a log appender and we'd have a default one which appends using mtrace, but when running it in the run now page we could also have a browser devtool console appender
* [x] https://developer.mozilla.org/en-US/docs/Web/API/console/group

~~Rather than reinventing the wheel we just pull in log4php via composer and then write a couple appenders~~

keevan:
>  When running a flow you'll be able to set the log level depending on how much data you want to see. This can be set in the run now page as well as at the flow level

This is set at the flow/site level, not on a run by run basis. If setting this on the run now page is still desired, it can be raised as a new issue.

PR #448, opened 2022-08-05
Align log statements to read easier #406

### LINKED ISSUE #406: Align the log statements in log / mtrace
When logging padd out the names of each step with spaces so all the logs align

image

brendanheywood:
Resolves #406

### DIFF
--- a/classes/local/execution/logging_context.php
+++ b/classes/local/execution/logging_context.php
@@ -62,11 +62,15 @@ class logging_context {
             return;
         }
 
-        $logstr = 'Engine "' . $this->engine->name . '"';
+        $name = $this->engine->name;
+        $logstr = "Engine '{$name}'";
+        $strlen = min(20, strlen($name));
+        $dots = str_repeat('.' , 25 - $strlen);
         if (!is_null($this->enginestep)) {
-            $logstr .= ', step "' . $this->enginestep->name . '"';
+            $name = $this->enginestep->name;
+            $logstr .= " step {$name}{$dots}";
         }
-        $logstr .= ': ' . $message;
+        $logstr .= ' ' . $message;
         mtrace($logstr);
     }
 }
--- a/run.php
+++ b/run.php
@@ -100,6 +100,7 @@ require_sesskey();
 echo $OUTPUT->single_button($runnowurl, get_string('run_again', 'tool_dataflows'), 'post', ['class' => 'mb-3']);
 
 // Prepare to handle output via mtrace.
+echo html_writer::start_tag('small');
 echo html_writer::start_tag('pre');
 $CFG->mtrace_wrapper = 'tool_dataflows_mtrace_wrapper';
 try {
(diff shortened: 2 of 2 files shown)


brendanheywood:
image

keevan:
Oddly CI is throwing a hissy fit.

I was  wondering if we can simplify common bits, in particular when the engine says something vs when a step says it:
e.g. if the dataflow was called dataflow it might look something like
```
13:10:12.123 [dataflow][enrol user WS]: ............. Iteration 2: { something }
13:10:12.123 [dataflow][another step]: .............. Iteration 2: { something }
13:10:12.123 [dataflow][enrol user WS]: ............. Iteration 3: { something }
13:10:12.123 [dataflow]: ............................ Aborting, something went horribly wrong!
```

brendanheywood:
> I was wondering if we can simplify common bits, in particular when the engine says something vs when a step says it:
e.g. if the dataflow was called dataflow it might look something like

yes but not now. I want to redo the whole way logging is done so that when we log we are dealing a bunch of different objects we can do things with and not just a lump of text. I also want to add logging levels. Right now I just want the reading pain to go away slightly. See https://github.com/catalyst/moodle-tool_dataflows/issues/381

classes/local/execution/logging_context.php
```diff
@@ -62,11 +62,13 @@ public function log($message) {
...
-        $logstr = 'Engine "' . $this->engine->name . '"';
+        $logstr = "Engine '{$this->engine->name}'";
+        $strlen = min(20, strlen($this->enginestep->name));
+        $dots = str_repeat('.' , 25 - $strlen);
         if (!is_null($this->enginestep)) {
-            $logstr .= ', step "' . $this->enginestep->name . '"';
+            $logstr .= " step {$this->enginestep->name}{$dots}";
         }
-        $logstr .= ': ' . $message;
+        $logstr .= ' ' . $message;
```
keevan, line 71: So it seems passing the `enginestep->name` (which is derived from a magic getter) into any function is the problem.

You can fix this immediate issue by putting it into a variable first
```suggestion
        $name = $this->engine->name;
        $logstr = "Engine '{$name}'";
        $strlen = min(20, strlen($name));
        $dots = str_repeat('.' , 25 - $strlen);
        if (!is_null($this->enginestep)) {
            $logstr .= " step {$name}{$dots}";
        }
        $logstr .= ' ' . $message;
```

classes/local/execution/logging_context.php
```diff
@@ -62,11 +62,15 @@ public function log($message) {
...
         }
 
-        $logstr = 'Engine "' . $this->engine->name . '"';
+        $name = $this->engine->name;
+        $logstr = "Engine '{$name}'";
+        $strlen = min(20, strlen($name));
+        $dots = str_repeat('.' , 25 - $strlen);
         if (!is_null($this->enginestep)) {
-            $logstr .= ', step "' . $this->enginestep->name . '"';
+            $name = $this->engineistep->name;
```
keevan, line 70: Slight typo here:
```suggestion
            $name = $this->enginestep->name;
```

brendanheywood: my typing is haunted today

PR #454, opened 2022-08-05
fix: add full variables context when evaluating case expressions

### LINKED ISSUE #452: When evaluating an expression  inside a case the full context is not passed in
2022 Aug  5, 14:14:27.005 Engine 'BCPS enrolment PROD' Variable "steps" is not valid around position 40 for expression `record.COURSE_STATE == 'Drop' && count(steps.get_users_ws.result) == 0`. 

Ie the only thing we can reference in the case expressions is anything under record.* and not *anything* 

https://github.com/catalyst/moodle-tool_dataflows/blob/MOODLE_35_STABLE/classes/local/step/flow_logic_case.php#L212

keevan:
Resolves #452


keevan:
Will need to add tests

PR #457, opened 2022-08-05
Issue #451: has_expression() is called with an array

### LINKED ISSUE #451: Whan assigning a mapped output to an expression it cannot accept an array
2022 Aug  5, 14:07:21.618 Engine 'enrolment PROD' Argument 1 passed to tool_dataflows\parser::has_expression() must be of the type 
string, array given, called in /var/www/moodle.local/admin/tool/dataflows/classes/parser.php on line 92 

ie with webservice step it should return an array of users, either 0 1 or many

```
webservice: core_user_get_users_by_field
user: admin
parameters: |
  field: email
  values: 
              - ${{ data.EMAIL }}
  
failure: abortstep
outputs:
  users: '${{ result }}'  <<--- this doesn't work
```

jaypha:
Resolves #451

### DIFF
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -212,7 +212,9 @@ class dataflow extends persistent {
 
                     if (is_object($field)) {
                         foreach ($field as &$fieldlet) {
-                            $localparse($fieldlet);
+                            if (!is_object($fieldlet) && !is_array($fieldlet)) {
+                                $localparse($fieldlet);
+                            }
                         }
                     } else {
                         $localparse($field);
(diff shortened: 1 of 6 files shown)


jaypha:
The only I could find where an object could possibly be passed to has_expression() is in reader_sql. So I put a check.

jaypha:
A sample flow. Make sure to set globals.vars.stuff to an array.
[zrun_20220805_0659.yml.txt](https://github.com/catalyst/moodle-tool_dataflows/files/9266180/zrun_20220805_0659.yml.txt)

keevan (commented):
Can you add a unit test to ensure it throws some error during run time, and bump the version.php as there have been langstring changes

Issue #392, opened 2022-08-01
Fix and improve the curl headers feature


brendanheywood:
Tasks
* [x] remove the json and yaml parsing and just make this a raw text chunj
* [x] simply the text to just show the format for header: value
* [x] make it work with multiple headers
* [x] make the bash export also work with multiple headers
* [x] fix bash escaping

brendanheywood:
Just put this back on the queue and re-done the task list here to be complete

jaypha:
The headers field is currently designed to require RFC2616 compliance as requested in the comments to #466.
Could it be, perhaps, a bit looser to allow comments and blank lines, which would be removed before adding to curl?

PR #466, opened 2022-08-05
fix: simplifying curl header field

marcghaly:
Hi,

This is a PR that solves issue #392

Regards,

Marc

### DIFF
--- a/classes/local/step/connector_curl.php
+++ b/classes/local/step/connector_curl.php
@@ -118,11 +118,6 @@ class connector_curl extends connector_step {
      * @param \MoodleQuickForm $mform
      */
     public function form_add_custom_inputs(\MoodleQuickForm &$mform) {
-        $jsonexample = [
-            'header-key' => 'header value',
-            'another-key' => '1234',
-        ];
-        $ex['json'] = \html_writer::nonempty_tag('pre', json_encode($jsonexample, JSON_PRETTY_PRINT));
         $ex['yaml'] = '<pre>header-key: header value<br>another-key: 1234</pre>';
 
         $urlarray = [];
--- a/lang/en/tool_dataflows.php
+++ b/lang/en/tool_dataflows.php
@@ -342,7 +342,7 @@ $string['connector_curl:sideeffects'] = 'Does this request have side effects?';
 $string['connector_curl:sideeffects_help'] = 'Most read requests done via http GET do not have side effects ie they change state on the remote server, and generally any POST or PUT does have side effects. Curl calls with side effects are not actually executed in dry run mode.';
 $string['connector_curl:timeout'] = 'Timeout';
 $string['connector_curl:timeout_help'] = 'Time in seconds after which the request will abort, default is 60 seconds.';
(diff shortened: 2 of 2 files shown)


classes/local/step/connector_curl.php
```diff
@@ -120,10 +120,9 @@ public function define_outputs(): array {
     public function form_add_custom_inputs(\MoodleQuickForm &$mform) {
         $jsonexample = [
             'header-key' => 'header value',
-            'another-key' => '1234',
         ];
         $ex['json'] = \html_writer::nonempty_tag('pre', json_encode($jsonexample, JSON_PRETTY_PRINT));
-        $ex['yaml'] = '<pre>header-key: header value<br>another-key: 1234</pre>';
```
brendanheywood, line 126: I would keep this example, I'd also kill the dead json code above

marcghaly: fixed

lang/en/tool_dataflows.php
```diff
@@ -342,7 +342,7 @@
 $string['connector_curl:sideeffects_help'] = 'Most read requests done via http GET do not have side effects ie they change state on the remote server, and generally any POST or PUT does have side effects. Curl calls with side effects are not actually executed in dry run mode.';
 $string['connector_curl:timeout'] = 'Timeout';
 $string['connector_curl:timeout_help'] = 'Time in seconds after which the request will abort, default is 60 seconds.';
-$string['connector_curl:field_headers_help'] = 'Headers should be in the following JSON format: {$a->json} You can also use the following YAML format: {$a->yaml}';
+$string['connector_curl:field_headers_help'] = 'Headers should be in the following YAML format: {$a->yaml}';
```
brendanheywood, line 345: ok so this isn't a bug in your code in this patch but can you please fix this anyway: the headers simply do not work!

If I cut and paste the example headers in the help text into the box then I expect those 2 headers to come through. Right now I am:

a) not getting the curl working at all
b) the header format which is being put into the dry run log is also wrong:

```
2022 Aug  6, 13:29:35.219 Engine 'Curl into myself' step meta curl......... curl -X GET http://moodle.local/ -H 'header-key: header value
another-key: 1234
' 
```
this curl cli command is not valid, it should be multiple -H cli arguments. You should be able to cut and paste this command into bash and run it and have it be identical. 

```
curl -X GET http://moodle.local/ -H 'header-key: header value' -H 'another-key: 1234'
```

Also this CLI output is not being escaped correctly, it needs to be specifically bash escaped so it works on the cli when you cut and paste it

brendanheywood: The more I think about it I'd like to remove all the json and yaml parsing from this specific field as it just doesn't make sense here. We are not defining any variables or anything where they add value. We are dealing with a chunk of text over multiple lines which is verbatim the same as you'd see over the wire and so this field should just be plain text and conform the rfc

https://www.rfc-editor.org/rfc/rfc2616.html#section-4.2

So all we need to do here is split it on new lines, evaluate each to sort the expressions, and then push each line out to curl and the curl cli args

keevan:
As pointed out by @jaypha, this can be closed off as it has been superseded by PR #604 and the affected changes are now redundant.

jaypha:
Closing as the issue has been resolved by #604

Issue #458, opened 2022-08-05
Cache internal small fragments of svg


brendanheywood:
If we ever want to render a fragment of a a graph, eg a single node which is used in the table, then pass through a MUC definition first

* [ ] only do this for the inline bits of svg which block render of the main html page
* [ ] The cache key could be based off a hash of the graphviz code that goes in. This cache will be completely deterministic and can be localizable 
* [ ]  this shouldn't be needed for the whole graphs are they are cached in the browser as a separate request

PR #467, opened 2022-08-05
feat: Cache internal dot binaries fragments

marcghaly:
Hi @brendanheywood,

This PR solves issue #458, wanted to use a 256 hash but not sure of the use in `admin/tool/dataflows/visual.php`

Regards,

Marc

### DIFF
--- a/db/caches.php
+++ b/db/caches.php
@@ -15,21 +15,20 @@
 // along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
 
 /**
- * Version
+ * tool_dataflows cache definitions.
  *
- * @package   tool_dataflows
- * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
- * @copyright  2022, Catalyst IT
+ * @package    tool_dataflows
+ * @category   cache
+ * @author     Ghaly Marc-Alexandre <marc-alexandreghaly@catalyst-ca.net>
+ * @copyright  Catalyst IT, 2022
  * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 
 defined('MOODLE_INTERNAL') || die();
 
-$plugin->version = 2022080600;
-$plugin->release = 2022080600;
-$plugin->requires = 2017051500;    // Our lowest supported Moodle (3.3.0).
-$plugin->supported = [35, 401];    // Available as of Moodle 3.9.0 or later.
-// TODO $plugin->incompatible = ;  // Available as of Moodle 3.9.0 or later.
-$plugin->component = 'tool_dataflows';
-$plugin->maturity = MATURITY_ALPHA;
-$plugin->dependencies = [];
+$definitions = [
+    'dot' => [
+        'mode' => cache_store::MODE_APPLICATION,
+        'canuselocalstore' => true,
+    ]
+];
--- a/classes/visualiser.php
+++ b/classes/visualiser.php
@@ -75,6 +75,13 @@ class visualiser {
     public static function generate(string $dotscript, ?string $type = 'svg') {
         global $CFG, $OUTPUT;
 
+        $cache = \cache::make('tool_dataflows', 'dot');
+        $hash = hash('sha256', $dotscript . $type);
+
+        if ($data = $cache->get($hash)) {
+            return $data;
+        }
+
         if (!helper::is_graphviz_dot_installed()) {
             return $OUTPUT->render(
                 new \pix_icon(helper::GRAPHVIZ_ALT_ICON,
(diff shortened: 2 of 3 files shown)


visual.php
```diff
@@ -36,12 +36,24 @@
...
+
 // Generate DOT script based on the configured dataflow.
 $dataflow = new dataflow($id);
 $dotscript = $dataflow->get_dotscript();
 
-// Generate the image based on the DOT script.
-$contents = \tool_dataflows\visualiser::generate($dotscript, $type);
+$hash = hash('sha256', $dotscript);
+
+if ($data = $cache->get($hash) && $type === 'svg') {
```
brendanheywood, line 47: Do the svg check so we don’t go to cache for no reason

rest looks good

brendanheywood: except code standards is red

visual.php
```diff
@@ -36,12 +36,24 @@
...
 
+$cache = cache::make('tool_dataflows', 'svg');
+
 // Generate DOT script based on the configured dataflow.
 $dataflow = new dataflow($id);
 $dotscript = $dataflow->get_dotscript();
 
-// Generate the image based on the DOT script.
-$contents = \tool_dataflows\visualiser::generate($dotscript, $type);
+$hash = hash('sha256', $dotscript);
```
brendanheywood, line 45: More of a general feedback thing which doesnt super matter here, if you want to cache something like here and there are 2 or more outputs that it could have for the cache, try to analyse all the dimensions of the inputs going into that cache and then make sure that all of the those inputs are represented in the cache key.

So in this case something like this would be fine:

$hash = hash('sha256', $dotscript . $type);

Then you get automatic cache key collision avoided for free

visual.php
```diff
@@ -36,12 +36,24 @@
...
+
+if ($data = $cache->get($hash) && $type === 'svg') {
+    $contents = $data;
+} else {
+    // Generate the image based on the DOT script.
+    $contents = \tool_dataflows\visualiser::generate($dotscript, $type);
```
brendanheywood, line 51: Actually this patch isn't having the performance win I expected. The way I am testing this is profiling and perf testing a page like admin/tool/dataflows/view.php?id=10 on a flow which has heaps of steps and each step image turn into its own cli call to render the graph.

None of this display logic comes through visual.php it is all inline svg not svg inside an image tag. So all of this logic needs to be moved up to an internal cache inside generate()

A good generic way of testing these types of things is temporarily make the mildly expensive thing you are trying to improve massively expensive by adding a sleep(N) call strategically right next to the cli exec, and then profile the whole page and get a base line for how slow it is then do you patch and get a before and after comparison.

So broadly whenever you see a task like 'add a cache' never just add the cache, instead make sure you have a clear performance metric to make sure you are making the thing faster that needs to be faster.

visual.php
brendanheywood: Also lets dump the requirement I added about only caching svg. We can also cache the binary versions as well

db/caches.php
```diff
@@ -0,0 +1,21 @@
...
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+$definitions = [
+    'svg' => [
```
brendanheywood, line 18: if we cache other types lets call this dot or graphvis instead of svg

Also because this cache is purely deterministic (ie for a given hash key they value will always be identical) this cache can be distributed which means you can declare it as canuselocalstore

https://github.com/moodle/moodle/blob/master/lib/db/caches.php#L40

visual.php
marcghaly: Fixed.

visual.php
marcghaly: Fixed.

visual.php
marcghaly: I have moved the logic inside the generate method and tested time consumption 

Here is the before and after 
Screenshot from 2022-08-06 01-51-55
Screenshot from 2022-08-06 01-49-05

db/caches.php
marcghaly: Fixed

marcghaly:
Hi @brendanheywood,

I have fixed issues above,

Regards,

Marc

brendanheywood:
Ci is still red

marcghaly:
Just fixed

brendanheywood:
thanks @marcghaly this is spot on

Issue #470, opened 2022-08-06
Add a caching layer to the symphony expression tokenize / parser


brendanheywood:
On many pages this is the biggest bottleneck outside of things with genuine side effects. It comes with a caching layer already:

https://symfony.com/doc/current/components/expression_language/caching.html

https://symfony.com/doc/current/components/cache.html

So two approaches come to mind:
1) we write a symphony cache adapter for moodle MUC, this should not be a huge amount of work but isn't trivial but gives us the best long term flexibility
2) we use one of the built in ones which we can wire to work with moodle. The most promising would be the file system adapter and then point it at either $CFG->cachedir (or $CFG->localcachedir but ONLY if we 100% know these caches are deterministic)

Start with the second and see what the speed up is. If we do know these are deterministic then we can put these into apcu as well and make them really fast

https://symfony.com/doc/current/components/cache/adapters/apcu_adapter.html

PR #493, opened 2022-08-07
Added expression caches #470

brendanheywood:
I've added both a an apcu cache or a file cache if apcu isn't installed but the result are so far anti climactic and the expressions are still a major bottleneck. I can also see the cache being populated so it is clearly working.

There a few steps which are not using parser.php but that can't fully account for it

#470 470

### DIFF
--- a/classes/parser.php
+++ b/classes/parser.php
@@ -16,6 +16,8 @@
 
 namespace tool_dataflows;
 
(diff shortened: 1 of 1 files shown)


brendanheywood:
Merged anyway it is a marginal improvement

PR #500, opened 2022-08-08
various step authoring enhancements

keevan:
- feat: add simpler methods from the step type to call
- feat: config now automatically resolves "record[*]" expressions

### DIFF
--- a/classes/local/execution/engine_step.php
+++ b/classes/local/execution/engine_step.php
@@ -293,4 +293,26 @@ abstract class engine_step {
                 break;
         }
     }
+
+    /**
+     * Returns an array with all the variables available, with the context of the step
+     *
+     * @return  array
+     */
+    public function get_variables(): array {
+        // Config values are directly referenceable, step values go through
+        // step.fieldname, everything else is available through expressions,
+        // such as 'dataflow.id' and 'steps.mystep.name' for example.
+        $variables = $this->engine->get_variables();
+        $step = $variables['steps']->{$this->stepdef->alias};
+
+        // Pull out the config.
+        $config = $step->config ?? new \stdClass;
+
+        return array_merge(
+            $variables,
+            ['step' => $step],
+            (array) $config
+        );
+    }
 }
--- a/classes/local/execution/flow_engine_step.php
+++ b/classes/local/execution/flow_engine_step.php
@@ -16,6 +16,8 @@
 
 namespace tool_dataflows\local\execution;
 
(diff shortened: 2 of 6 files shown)


classes/local/execution/engine_step.php
```diff
@@ -293,4 +293,26 @@ protected function on_change_status() {
...
+        $variables = $this->engine->get_variables();
+        $step = $variables['steps']->{$this->stepdef->alias};
+
+        // Pull out the config.
+        $config = $step->config ?? new \stdClass;
+
+        return array_merge(
+            $variables,
+            ['step' => $step],
+            (array) $config,
```
jaypha, line 315: This is not supported in older versions of PHP.

classes/local/execution/flow_engine_step.php
```diff
@@ -84,10 +86,24 @@ public function get_variables(): array {
...
+            // step type doesn't want to (e.g. SQL reader does it own handling).
+            $parser = new parser;
+            $parser->evaluate_recursive($config, $variables);
+        }
+
         return array_merge(
             $variables,
             ['step' => $step],
-            (array) $step->config
+            (array) $config,
```
jaypha, line 106: Here too.

classes/local/step/base_step.php
```diff
@@ -611,4 +611,37 @@ public function get_output_label(int $position): string {
...
+
+    /**
+     * Get the step's (definition) current config.
+     *
+     * Helper method to reduce the complexity when authoring step types.
+     *
+     * @return  \stdClass configuration object
+     */
+    protected function get_config(): \stdClass {
+        return $this->enginestep->stepdef->config ?? $this->stepdef->config;
```
jaypha, line 623: This is redundant. If enginestep is defined, then stepdef will also be defined, and be the same value as enginestep->stepdef.

keevan: This is required because the engine step's configuration resolves things for `record.*` values, where the other one does not as it has no context of the engine.

Would this work better backwards? e.g. Add the enginestep check at the stepdef layer? (which would probably need to go through steptype, and get the engine's stepdef's config)

jaypha: I need more explanation. Don't they both point to the same object?

keevan: Nevermind, you were right. It is redundant. 

Thankfully, from re-testing I found another issue which I've pushed a fix to also relating to outputs.

PR #502, opened 2022-08-08
Issue #495: Fix dry run issue in curl connectors

### LINKED ISSUE #495: connector_curl should still read from the site in a dry run when there is no side effect.

jaypha:
Resolves #495

### DIFF
--- a/classes/local/step/connector_curl.php
+++ b/classes/local/step/connector_curl.php
@@ -30,6 +30,9 @@ use tool_dataflows\helper;
  */
 class connector_curl extends connector_step {
 
+    /** Lowest number for HTTP errors. */
+    public const HTTP_ERROR = 400;
+
     /** @var int Time after which curl request is aborted */
     protected $timeout = 60;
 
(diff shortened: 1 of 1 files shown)


classes/local/step/connector_curl.php
```diff
@@ -250,7 +250,7 @@ public function execute(): bool {
         }
 
         // Download response to file provided destination is set.
-        if (!empty($config->destination) && !$isdryrun) {
+        if (!empty($config->destination) && $doingcurl) {
```
brendanheywood, line 253: can we make this logic simpler and just exit earlier if it isn't curling?

jaypha: done

classes/local/step/connector_curl.php
```diff
@@ -30,6 +30,12 @@
  */
 class connector_curl extends connector_step {
 
+    /** Error number for curl timeout. */
+    public const CURL_TIMEOUT_ERROR = 28;
```
brendanheywood, line 34: can this use the php curl constants? is this the same as CURLE_OPERATION_TIMEOUTED

jaypha: done

PR #503, opened 2022-08-08
feat: when creating a new step the name will be prefilled automatically

### LINKED ISSUE #488: When creating a new step the name should default to the simple version of the step type
Grab the class then strip it right down

keevan:
This will be equal to just the class name, and later on, the class names will be simplified to something like 'curl' or 'sql' which will be even better, so opted to not do anything too opinionated here.

Resolves #488

### DIFF
--- a/classes/form/step_form.php
+++ b/classes/form/step_form.php
@@ -404,6 +404,11 @@ class step_form extends \core\form\persistent {
             $data = $steptype->form_get_default_data($data);
         }
 
+        // Automatically fill in the name for new steps.
+        if (empty($data->id)) {
+            $data->name = $steptype->get_name();
+        }
+
         return $data;
     }
 


classes/form/step_form.php
```diff
@@ -404,6 +404,16 @@ protected function get_default_data() {
             $data = $steptype->form_get_default_data($data);
         }
 
+        // Automatically fill in the name for new steps.
+        if (empty($data->id)) {
+            $classname = $type;
+            $position = strrpos($classname, '\\');
+            $basename = substr($classname, $position + 1);
```
brendanheywood, line 411: can you go a bit deeper and strip out the 'type' words like 'connector' 'flow_logic' etc

OR a different approach start with the lang string rather than the classname might be better results

keevan: That's a great idea. Starting with the lang string might be better actually, since it would allow for more localization. Note that we still have those connector/flow terminology in our existing lang strings, but I'll make a note to clean those up as part of the other task.

e.g.

image

Issue #517, opened 2022-08-08
get_variables method is returning objects preventing count case step


marcghaly:

marcghaly:
This cast prevent the use of `count` https://github.com/catalyst/moodle-tool_dataflows/blob/MOODLE_35_STABLE/classes/step.php#L908 in case step.

Adding the `get_object_vars` to expression provider solves the issue.

![Screenshot from 2022-08-08 11-37-29](https://user-images.githubusercontent.com/76186988/183458019-93ef5b32-17a4-4b4a-9813-0d8ff9ffa783.png)

keevan:
Were you able to use the isset expression function instead? This issue is lacking an example of the expression used and an example of something stored which it is referencing.

marcghaly:
Tried using isset with the id instead but isset is being passed the litteral expression giving false results:

![Screenshot from 2022-08-08 21-27-15](https://user-images.githubusercontent.com/76186988/183542810-57fc6d33-0b76-4e41-af09-ee01379523e1.png)

![Screenshot from 2022-08-08 21-26-46](https://user-images.githubusercontent.com/76186988/183542811-f83f268c-de95-415c-84c1-4627b0dec89e.png)

![Screenshot from 2022-08-08 21-26-11](https://user-images.githubusercontent.com/76186988/183542813-de9da29c-8476-41f4-93da-4acccf12b004.png)

PR #518, opened 2022-08-08
feat: add get_object_vars to expression provider

marcghaly:


marcghaly:
This resolves issue #517

Issue #481, opened 2022-08-06
Statically cache calls the dependancies or dependants


brendanheywood:
If either of these are called then see how aggressive we can be in loading the whole dag into memory and then not going back to the db at all

Lots of pages call this inside loops which creates 100s of db calls

brendanheywood:
Viewing a big flow makes a whopping 1k calls: DB reads/writes: 1063/1

600+ of these are dependencies

Another 236 of these are loading steps one by one. I think if we load a flow then grab one step out of it then we should load them all in.

PR #529, opened 2022-08-09
Issue #481: Reduce db calls

petersistrom:
#481

### DIFF
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -533,9 +533,14 @@ class dataflow extends persistent {
 
         // Fetch the dot script node from each step to construct them.
         $steps = $this->steps;
+
         $nodes = [];
         $contentonly = true;
+        $steparray = [];
+        $outputlabels = [];
         foreach ($steps as $step) {
+            $steparray[$step->id] = $step;
+            $outputlabels[$step->id] = $step->steptype->get_output_labels();
             $nodes[] = $step->get_dotscript($contentonly);
         }
         $nodes = implode(';' . PHP_EOL, $nodes);
@@ -550,8 +555,8 @@ class dataflow extends persistent {
         ];
         foreach ($edges as $edge) {
             [$srcid, $destid] = $edge;
-            $srcstep = new step($srcid);
-            $deststep = new step($destid);
+            $srcstep = $steparray[$srcid];
+            $deststep = $steparray[$destid];
 
             $localstyles = [];
             $typesvalid = class_exists($srcstep->type) && class_exists($deststep->type);
@@ -578,7 +583,8 @@ class dataflow extends persistent {
                 'stepid' => $destid,
             ]);
             if (isset($dependency->position)) {
-                $outputlabel = $srcstep->steptype->get_output_label($dependency->position);
+                $outputlabel = $outputlabels[$srcid][$dependency->position];
+
                 $connectionstyles = [
                     'label' => $outputlabel,
                     'fontsize'  => '10',
--- a/classes/step.php
+++ b/classes/step.php
@@ -49,6 +49,12 @@ class step extends persistent {
     /** @var array $dependson */
     private $dependson = [];
 
+    /** @var array $dependents array for lazy loading step dependents */
+    private $dependants = null;
+
+    /** @var array $dependents array for lazy loading step dependants */
+    private $dependents = null;
+
     /** @var \stdClass of engine step states and timestamps */
     private $states;
 
@@ -414,9 +420,10 @@ class step extends persistent {
     /**
      * Returns a list of other steps that depend on this step before they can run.
      *
-     * @return  array step dependencies
+     * @param bool $reload
+     * @return  array|null step dependencies
      */
-    public function dependants(): array {
+    public function dependants($reload = false): array {
         global $DB;
         $sql = "SELECT step.id,
                        step.name,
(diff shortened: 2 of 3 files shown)


classes/dataflow.php
```diff
@@ -550,8 +560,8 @@ public function get_dotscript(): string {
         ];
         foreach ($edges as $edge) {
             [$srcid, $destid] = $edge;
-            $srcstep = new step($srcid);
-            $deststep = new step($destid);
+            $srcstep = $steparray[$srcid];
```
brendanheywood, line 563: This bit make sense so keep this

classes/steps_table.php
```diff
@@ -79,6 +79,7 @@ public function make_columns(): void {
      */
     public function col_name(\stdClass $record): string {
         $step = new step($record->id);
+        $step->states->valid = true;
```
brendanheywood, line 82: this feels out of place, how does this one help?

visual.php
```diff
@@ -38,6 +38,7 @@
 
 // Generate DOT script based on the configured dataflow.
 $dataflow = new dataflow($id);
+$dataflow->validate_dataflow();
```
brendanheywood, line 41: what side effects does this have here which help?

classes/steps_table.php
petersistrom: Changing this line https://github.com/catalyst/moodle-tool_dataflows/pull/529/files#diff-642f9f96d3487795294cba491b20d8119c2d15032bb22abeb72c475a88191cf5L859 made all the steps in the table false but it definitely needs some tweaking

visual.php
petersistrom: Sets the valid state for each step, so each step doesn't have to be validated individually again

classes/steps_table.php
brendanheywood: ok not sure this is the right approach so lets land the other bits which are fine and the changes yet to come for the persistent

classes/step.php
```diff
@@ -449,9 +460,12 @@ public function dependents() {
              LEFT JOIN {tool_dataflows_steps} step ON sd.stepid = step.id
                  WHERE sd.dependson = :stepid";
 
-        $deps = $DB->get_records_sql($sql, [
-            'stepid' => $this->id,
-        ]);
+        if (empty($this->dependents) || $reload) {
```
brendanheywood, line 463: should this be isset rather than empty and the attribute should default to null rather than an empty array?

classes/step.php
```diff
@@ -635,6 +649,7 @@ public function validate_step() {
                 $errors = array_merge($errors, $extravalidation);
             }
         }
+
```
brendanheywood, line 652: try to avoid unrelated changes

classes/step.php
```diff
@@ -849,8 +864,8 @@ public function get_dotscript($contentonly = false): string {
         } else {
             // Invalid step type styles.
             $stepstyles = [
-               'fillcolor' => '#ca3120',
-               'fontcolor' => '#ffffff',
+                'fillcolor' => '#ca3120',
```
brendanheywood, line 867: things like this are good to cleanup, but better to be in a different pr or commit

classes/step.php
```diff
@@ -427,19 +434,23 @@ public function dependants(): array {
...
-        $deps = $DB->get_records_sql($sql, [
-            'dependson' => $this->id,
-        ]);
+        if (!isset($this->dependants) || $reload) {
+            $deps = $DB->get_records_sql($sql, [
+                'dependson' => $this->id,
+            ]);
+            $this->dependants = $deps;
+        }
         return $deps ?? [];
```
brendanheywood, line 443: this should always return $this->dependants;

classes/step.php
```diff
@@ -449,10 +460,13 @@ public function dependents() {
...
-            'stepid' => $this->id,
-        ]);
-        return $deps;
+        if (!isset($this->dependents) || $reload) {
+            $deps = $DB->get_records_sql($sql, [
+                'stepid' => $this->id,
+            ]);
+            $this->dependents = $deps;
+        }
+        return $deps ?? [];
```
brendanheywood, line 469: same

classes/dataflow.php
```diff
@@ -533,9 +533,14 @@ public function get_dotscript(): string {
 
         // Fetch the dot script node from each step to construct them.
         $steps = $this->steps;
+
         $nodes = [];
         $contentonly = true;
+        $steparray = [];
+        $oputlabels = [];
```
keevan, line 540:
```suggestion
        $outputlabels = [];
```
Did you mean this?

classes/step.php
```diff
@@ -49,6 +49,12 @@ class step extends persistent {
     /** @var array $dependson */
     private $dependson = [];
 
+    /** @var array $dependson */
+    private $dependants = null;
+
+    /** @var array $dependson */
+    private $dependents = null;
```
keevan, line 56: The variable it's referencing is incorrect (copy + paste but no edit).

Are you able to add a brief description to the comment, because it adds no value currently (aside from type specification)? i.e. it would be good to know what the variable means and is for.

jaypha:
The image in the dataflows manager screen is not updating in line with changes to the dataflow.
For example
Screenshot from 2022-09-06 17-17-07
When it should be.
Screenshot from 2022-09-06 17-17-45

brendanheywood:
> The image in the dataflows manager screen is not updating in line with changes to the dataflow. For example 

This hints that there is probably deeper issues which the cache not being invalidated and there is likely a range of related bugs with the same root cause

petersistrom:
I'm not sure this is related to this PR. Seems to be happening on STABLE as well. A hard refresh of the browser will load the correct preview image.

jaypha:
I am seeing this too.

PR #531, opened 2022-08-10
525 output mappings

### LINKED ISSUE #525: Move output mappings under its own sub tree eg ${{steps.ws.outputs.myvar}}
* [ ] move step level vars under 'vars' 
* [ ] rename flow level config to 'vars' (do this deeply, var names and method names)
* [ ] add flow level config with enabled as the real config ie dataflow.config.enabled

jaypha:
Resolves #525 

Moved outputs to 'vars'.
Replaced dataflows.config with dataflows.vars.
Added config to dataflows for other settings.

### DIFF
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -96,7 +96,7 @@ class dataflow extends persistent {
             'name' => ['type' => PARAM_TEXT],
             'enabled' => ['type' => PARAM_BOOL, 'default' => false],
             'concurrencyenabled' => ['type' => PARAM_BOOL, 'default' => false],
-            'config' => ['type' => PARAM_TEXT, 'default' => ''],
+            'vars' => ['type' => PARAM_TEXT, 'default' => ''],
             'timecreated' => ['type' => PARAM_INT, 'default' => 0],
             'userid' => ['type' => PARAM_INT, 'default' => 0],
             'timemodified' => ['type' => PARAM_INT, 'default' => 0],
@@ -221,7 +221,7 @@ class dataflow extends persistent {
         // Prepare variable data for the dataflow key.
         $dataflow = (object) $this->get_export_data();
         unset($dataflow->steps);
-        $dataflow->config = $this->get_config(false);
+        $dataflow->vars = $this->get_vars(false);
         $dataflow->states = $this->states;
         $dataflow->id     = $this->id;
 
@@ -245,15 +245,15 @@ class dataflow extends persistent {
     }
 
     /**
-     * Return the configuration of the dataflow, parsed such that any
+     * Return the vars of the dataflow, parsed such that any
      * expressions are evaluated at this point in time.
      *
-     * @param   bool $expressions whether or not to parse expressions when returning the config
-     * @return  \stdClass configuration object
+     * @param   bool $expressions whether or not to parse expressions when returning the vars
+     * @return  \stdClass vars object
      */
-    protected function get_config($expressions = true): \stdClass {
-        $yaml = Yaml::parse($this->raw_get('config'), Yaml::PARSE_OBJECT_FOR_MAP);
-        // If there is no config, return an empty object.
+    protected function get_vars($expressions = true): \stdClass {
+        $yaml = Yaml::parse($this->raw_get('vars'), Yaml::PARSE_OBJECT_FOR_MAP);
+        // If there is no vars, return an empty object.
         if (empty($yaml)) {
             return new \stdClass();
         }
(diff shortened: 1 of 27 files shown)


db/upgrade.php
```diff
@@ -141,5 +141,34 @@ function xmldb_tool_dataflows_upgrade($oldversion) {
...
+    if ($oldversion < 2022081000) {
+        // Changing type of field config on table tool_dataflows_steps to text.
+        $table = new xmldb_table('tool_dataflows');
+
+        $field = new xmldb_field('config', XMLDB_TYPE_TEXT, null, null, null, null, null, 'concurrencyenabled');
+        if ($dbman->field_exists($table, $field)) {
+            $dbman->rename_field($table, $field, 'vars');
+        }
+
+        $field = new xmldb_field('confighash', XMLDB_TYPE_TEXT, null, null, null, null, null, 'usermodified');
```
brendanheywood, line 153: confighash should not have been caught up these changes, confighash represents the entirety of the config of the entire flow, steps, flow config, vars all of its. If anything changed at all then the "config" in the looser definition of the word has changed and so the hash should change

jaypha:
confighash should not have changed

classes/dataflow.php
```diff
@@ -746,49 +759,49 @@ public function get_export_data() {
...
             Yaml::DUMP_MULTI_LINE_LITERAL_BLOCK
         );
     }
 
     /**
-     * Saves the current config to the tool_dataflows_versions table
+     * Saves the current vars to the tool_dataflows_versions table
      */
-    public function save_config_version() {
+    public function save_vars_version() {
```
brendanheywood, line 786: this all looks like it should not be touched too

db/upgrade.php
jaypha: done

classes/dataflow.php
jaypha: done

classes/dataflow.php
```diff
@@ -773,21 +786,21 @@ public function set_var($name, $value) {
...
             $newconfighash = sha1($configyaml);
 
-            // Inserts the stored config if it is a new config.
+            // Inserts the stored vars if it is a new vars.
             if (!$DB->record_exists('tool_dataflows_versions', ['dataflowid' => $this->id, 'confighash' => $newconfighash])) {
                 $DB->insert_record('tool_dataflows_versions',
                     (object) ['dataflowid' => $this->id, 'confighash' => $newconfighash, 'configyaml' => $configyaml]);
             }
-            // Set the config hash in dataflows table.
+            // Set the vars hash in dataflows table.
```
brendanheywood, line 803: this comment can go?

classes/dataflow.php
```diff
@@ -773,21 +786,21 @@ public function set_var($name, $value) {
...
-                (array) $config,
+                (array) $vars,
                 helper::YAML_DUMP_INLINE_LEVEL,
                 helper::YAML_DUMP_INDENT_LEVEL,
                 Yaml::DUMP_MULTI_LINE_LITERAL_BLOCK
             );
             $newconfighash = sha1($configyaml);
 
-            // Inserts the stored config if it is a new config.
+            // Inserts the stored vars if it is a new vars.
```
brendanheywood, line 798: I think this comment can go too

brendanheywood: the change I mean

db/install.xml
```diff
@@ -10,7 +10,7 @@
         <FIELD NAME="name" TYPE="char" LENGTH="200" NOTNULL="true" SEQUENCE="false" COMMENT="The name of this dataflow"/>
         <FIELD NAME="enabled" TYPE="int" LENGTH="1" NOTNULL="true" DEFAULT="0" SEQUENCE="false"/>
         <FIELD NAME="concurrencyenabled" TYPE="int" LENGTH="1" NOTNULL="true" DEFAULT="0" SEQUENCE="false"/>
-        <FIELD NAME="config" TYPE="text" NOTNULL="false" SEQUENCE="false"/>
+        <FIELD NAME="vars" TYPE="text" NOTNULL="false" SEQUENCE="false" COMMENT="YAML encoded varaibles"/>
```
brendanheywood, line 13: typo in spelling

edit.php
```diff
@@ -84,8 +84,8 @@
             // Or, do the following if you don't want capability checks (discouraged).
             $persistent->from_record($data);
 
-            // Unset config hash if config has changed, so it will be resaved on next run.
-            if ($currentconfig !== $persistent->get('config')) {
+            // Unset config hash if vars has changed, so it will be resaved on next run.
+            if ($currentvars !== $persistent->get('vars')) {
```
brendanheywood, line 88: I think this was not quite right in the original code, if *anything* changes in the configuration of the flow then the config hash should change, so changes to enable or concurrency or the vars. I think just make it easy and if you save the record then we should reset the hash. If it regenerates and comes back with the same hash then no big deal

jaypha: fixed

db/install.xml
jaypha: done

classes/dataflow.php
jaypha: done

classes/dataflow.php
jaypha: done

brendanheywood:
+1 from me, the most important thing is the upgrade script, if there are other little bugs its ok and we can fix them in followup issues

Issue #524, opened 2022-08-09
BUG Setting an output mapping to a previous value doesn't work if the previous value is an array


brendanheywood:
[Single_web_service_call_20220809_0217.yml.txt](https://github.com/catalyst/moodle-tool_dataflows/files/9286227/Single_web_service_call_20220809_0217.yml.txt)

In the output setting the vars the first time works but then copying those vars works for ints but not for arrays:

![image](https://user-images.githubusercontent.com/187449/183549128-3e54f462-d8f7-452a-8cbe-6c9aad9fe79f.png)

jaypha:
The good news is that the outputs are defined correctly, it is just the display which is in error. 
The display is in error most likely because of #517. (The arrays had been converted to objects. When I converted them back to arrays, they displayed correctly).
So fixing #517 should also fix this.

marcghaly:
Also @jaypha, the `set_output` is not the main culprit here, the cast here is proper https://github.com/catalyst/moodle-tool_dataflows/blob/MOODLE_35_STABLE/classes/step.php#L908, issue is coming from here  https://github.com/catalyst/moodle-tool_dataflows/blob/MOODLE_35_STABLE/classes/local/step/flow_logic_case.php#L211 when retrieved all is an stdclass

PR #535, opened 2022-08-10
fix: output mapping display issue

marcghaly:
This PR solves issue #524

### DIFF
--- a/classes/local/step/base_step.php
+++ b/classes/local/step/base_step.php
@@ -132,6 +132,12 @@ abstract class base_step {
             $outputs = $parser->evaluate_recursive($config->outputs, $allvariables);
             $this->stepdef->set_output($outputs);
 
+            array_walk_recursive($outputs, function(&$item){
+                if (is_object($item)) {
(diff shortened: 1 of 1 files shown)


classes/local/step/base_step.php
```diff
@@ -132,6 +132,12 @@ public function prepare_outputs() {
             $outputs = $parser->evaluate_recursive($config->outputs, $allvariables);
             $this->stepdef->set_output($outputs);
 
+            array_walk_recursive($outputs, function(&$item){
+                if (is_object($item)) {
+                    $item = (array) $item;
+                };
+            });
+
```
keevan, line 140: I am worried that this is trying to do something it shouldn't, which is to change the structure of the actual data.

jaypha:
This PR is redundant. Fixing #517 should also fix #524.

marcghaly:
Hi @jaypha, #524 is only a display issue while #517 actually prevents the use of the count method in the case step for example due to the nature of returned element by `get_variables`

Issue #512, opened 2022-08-08
Exception in shutdown function after unit tests


jaypha:
This happens after a unit test run
```
Exception ignored in shutdown function Closure::__invoke: Attempting to change the status of a dataflow engine after it has concluded.
```
Whatever it is, I'm fairly sure we don't want it.

brendanheywood:
@petersistrom fyi regression, need to handle the case where one, or possible many flows have started and stopped and then finally the php script ends we need to only ever touch the last one and only if it wasn't actually running

PR #541, opened 2022-08-12
Issue #512: Fix exception in shutdown function

petersistrom:
Issue #512

### DIFF
--- a/classes/local/execution/engine.php
+++ b/classes/local/execution/engine.php
@@ -282,8 +282,11 @@ class engine {
         }
         // Register shutdown handler - if request is ended by client, abort and finalise flow.
         \core_shutdown_manager::register_function(function (){
-            $this->set_status(self::STATUS_ABORTED);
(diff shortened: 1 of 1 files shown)


classes/local/execution/engine.php
```diff
@@ -282,8 +282,12 @@ public function initialise() {
         }
         // Register shutdown handler - if request is ended by client, abort and finalise flow.
         \core_shutdown_manager::register_function(function (){
-            $this->set_status(self::STATUS_ABORTED);
-            $this->run->finalise($this->status, $this->export());
+            // If the script has stopped and flow is not finalised then abort.
+            if ($this->status != self::STATUS_FINISHED && $this->status != self::STATUS_FINALISED &&
```
jaypha, line 286: The engine does not use the STATUS_FINISHED status. You only need to check against STATUS_FINALISED and STATUS_ABORT.

PR #543, opened 2022-08-15
Issue #532: Remove configuration field from generic step form.

### LINKED ISSUE #532: Remove 'configuration' from steps without custom configs.
Is it needed?

jaypha:
Resolves #532

### DIFF
--- a/classes/local/step/base_step.php
+++ b/classes/local/step/base_step.php
@@ -163,9 +163,7 @@ abstract class base_step {
      * @return array
      */
     public static function form_define_fields(): array {
-        return [
-            'config' => ['type' => PARAM_TEXT, 'default' => ''],
-        ];
+        return [];
     }
 
     /**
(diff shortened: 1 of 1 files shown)


classes/local/step/base_step.php
```diff
@@ -381,16 +374,19 @@ public function form_set_input_rules(\MoodleQuickForm &$mform) {
...
-        if ($validation === true) {
-            return [];
-        }
+        if (count(static::form_define_fields())) {
+            $yaml = Yaml::parse($data->config, Yaml::PARSE_OBJECT_FOR_MAP);
+            if (empty($yaml)) {
+                $yaml = new \stdClass();
+            }
+            $validation = $this->validate_config($yaml);
+            if ($validation === true) {
```
jaypha, line 383: Note to self. This should test for !== and return $validation.

classes/local/step/base_step.php
```diff
@@ -381,16 +374,17 @@ public function form_set_input_rules(\MoodleQuickForm &$mform) {
...
+                $yaml = new \stdClass();
+            }
+            $validation = $this->validate_config($yaml);
+            if ($validation !== true) {
+                return $validation;
+            }
         }
-
-        return $validation;
+        return [];
```
keevan, line 387: You should return early here instead, it would keep everything else simpler to follow.
e.g.
```
if (empty(static::form_define_fields())) {
    return [];
}
```

classes/local/step/base_step.php
```diff
@@ -313,13 +311,15 @@ public function validate_config($config) {
...
     public function form_setup(\MoodleQuickForm &$mform) {
-        $mform->addElement('header', 'stepheader', $this->get_name());
-        $this->form_add_custom_inputs($mform);
+        // Add custom configuration fields under their own header, but only if there are custom fields to add.
+        if (count(static::form_define_fields())) {
+            $mform->addElement('header', 'stepheader', $this->get_name());
+            $this->form_add_custom_inputs($mform);
+            $this->form_set_input_types($mform);
+            $this->form_set_input_rules($mform);
+        }
```
keevan, line 320: You should change the `count` to an `!empty` check instead. As you are checking for whether there are fields or not, as opposed to needing to know the specific count of items.

classes/local/step/base_step.php
jaypha: done

classes/local/step/base_step.php
jaypha: done

keevan (approved):
Looks good to me, thanks @jaypha

PR #544, opened 2022-08-15
Issue #399: Rename case to switch

### LINKED ISSUE #399: Rename case step to switch step
I thin if we are gonna do this then lets do it everywhere in code and not just in the UI. It will be a breaking change so the sooner the better

jaypha:
Resolves #399


keevan:
Might need an update in upgrade.php where the existing step types set in the DB are converted to the new name, e.g. to ensure it doesn't break support for flows currently in BCPS

Issue #545, opened 2022-08-16
Convert email connector to email flow


marcghaly:
An email flow will allow the passing of data and evaluated expression to send specific emails.

PR #546, opened 2022-08-16
feat: email connector to emailflow, and record expression support

marcghaly:
Hi,

This PR solves issue #545 

Cheers,

Marc


brendanheywood:
I'm not sure about this, at the moment steps can be flow or connector and not both which is one of the driving reasons for the refactor. We need to have a connector email, if we also want a flow email then I think we need to clone it in the mean time

PR #557, opened 2022-08-29
fix: templates containing variables in pix helpers

keevan:
Check signature: 
mustache_pix_helper_testcase::test_no_exploitable_pix_helper_uses

### DIFF
--- a/classes/visualiser.php
+++ b/classes/visualiser.php
@@ -226,16 +226,19 @@ class visualiser {
 
         // Validate current dataflow, displaying any reason why the flow is not valid.
         $validation = $dataflow->validate_dataflow();
+        $valid = $validation === true;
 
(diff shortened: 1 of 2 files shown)


classes/visualiser.php
```diff
@@ -226,16 +226,19 @@ public static function display_dataflows_view_page(int $dataflowid, steps_table
 
         // Validate current dataflow, displaying any reason why the flow is not valid.
         $validation = $dataflow->validate_dataflow();
+        $valid = $validation === true;
```
Peterburnett, line 229: Is validation `bool|array` here? not sure if thats why you need the $valid var vs just using $validation

keevan: Yes, it's `true|object`

PR #562, opened 2022-08-29
fix: remove usage of close_recordset for table output

### LINKED ISSUE #553: T16 - Exception - Call to undefined method tool_dataflows\steps_table::close_recordset()
Seems like it's missing a backport expected for the table SQL class, which might be similar to an issue found previously for Moodle 3.5 clean.

image

Happens after creating a new dataflow, and going in to view the details page @ `admin/tool/dataflows/view.php?id=1`

keevan:
Resolves #553

### DIFF
--- a/classes/visualiser.php
+++ b/classes/visualiser.php
@@ -191,13 +191,7 @@ class visualiser {
 
         echo \html_writer::tag('div', $newdataflow . $importdataflow, ['class' => 'mb-3']);
 
-        // No hide/show links under each column.
-        $table->collapsible(false);
-        // Columns are presorted.
-        $table->sortable(false);
-        // Table does not show download options by default, an import/export option will be available instead.
-        $table->is_downloadable(false);
-        $table->out($table->pagesize, false);
+        $table->build();
 
         echo $output->footer();
     }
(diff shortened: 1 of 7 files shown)


classes/visualiser.php
```diff
@@ -318,7 +318,6 @@ public static function display_dataflows_view_page(int $dataflowid, steps_table
         $table->setup();
         $table->query_db(0); // No limit, fetch all rows.
         $table->pageable(false);
-        $table->close_recordset();
```
Peterburnett, line 321: Is there any info on why this was needed in the first place?

Would it be worth doing a method_exists() check to maintain it where we can?

Peterburnett: If the table isnt closing its recordset without the method, that feels like a pretty significant bug

keevan: So, in Moodle, it needed to be explicitly closed after calling `build_table` (which this block didn't seem to be following strictly).

In Totara, huge recordsets are handled using it's own implementation, which did not create a publicly available function `close_recordset` in the table_sql class.

So there were 2 options I had considered. Do a partial backport but not actually backport anything aside from an empty `close_recordset()` function in Totara - which means in Totara it just does nothing special since it's not required. Or remove it from being called in the first place which seemed like a better option at the time. 

I did a bit of digging, it seems for the most part resources are freed/closed up once they are no longer used. In particular since these things only display somewhat small sized and simple tables I thought about letting PHP handle the more internal bits.

Ref: https://github.com/moodle/moodle/blob/952d9e9/lib/tablelib.php#L1997

Peterburnett: Havent looked at the link yet, is the recordset field public? if so we could just call ->close() on it ourselves, or we could subclass the table_sql class to make it public? Otherwise, lets just gate it behind method_exists() and yes we can let the rest get cleaned up by the lang management classes

PR #564, opened 2022-08-29
fix: step links only when there is a step record

### LINKED ISSUE #555: Step chooser issue with creation of new steps
image

When picking a step, there is an outer link which redirects to the correct "new step" form page, but the link when you hover and click directly on the step type text goes to something like `/admin/tool/dataflows/step.php?id=0` which is always wrong.

This is likely due to the rendering of the step and dot script, and should probably NOT have a link if the step's dataflow id resolves to zero, allowing to it to be used for other purposes.

keevan:
This was previously setting it even if the value was zero (in string form)

Resolves #555

### DIFF
--- a/classes/step.php
+++ b/classes/step.php
@@ -835,8 +835,9 @@ class step extends persistent {
      */
     public function get_dotscript($contentonly = false): string {
         $localstyles = ['tooltip' => $this->description ?: $this->name];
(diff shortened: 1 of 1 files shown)


keevan:
Seems to cause a regression in Moodle (on the dataflows details page) so looking into that before undrafting.

PR #570, opened 2022-09-04
Stream out content into pre after the footer is rendered

brendanheywood:


keevan (approved):
This seems cool, I like it.

PR #575, opened 2022-09-09
Issue #574: Ensure dataflow's confighash has gets cleared whenever a step changes.

### LINKED ISSUE #574: Visual.php not updating is some circumstances
I you open up a dataflow, edit it, and then click on the 'manage flows' breadcrumb, the picture in the table is not always updated.

jaypha:
Resolves #574

### DIFF
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -44,6 +44,8 @@ class dataflow extends persistent {
     /** @var steps[] cache of steps connected to this dataflow */
     private $stepscache = [];
 
+    /** @var bool Set to true when the dataflow is in the process of deleting. */
+    private $isdeleting = false;
     /**
      * When initialising the persistent, ensure some internal fields have been set up.
      *
@@ -667,12 +669,33 @@ class dataflow extends persistent {
     }
 
     /**
+     * Called when steps are created, modified or removed.
+     */
+    public function on_steps_save() {
+        if (!$this->isdeleting) {
+            // Not needed if we are going to just delete the dataflow.
+            $this->set('confighash', '');
+            $this->save();
+        }
+    }
+
+    /**
      * Remove all the steps first.
      */
     protected function before_delete() {
+        $this->isdeleting = true;
         foreach ($this->steps as $step) {
             $this->remove_step($step);
         }
+    }
+
+    /**
+     * Called after deletion
+     *
+     * @param bool $result
+     */
+    protected function after_delete($result) {
+        $this->isdeleting = false;
     }
 
     /**
--- a/classes/step.php
+++ b/classes/step.php
@@ -542,6 +542,7 @@ class step extends persistent {
         $this->update_depends_on();
 
         $this->steptype->on_save();
+        $this->get_dataflow()->on_steps_save();
 
         return $this;
     }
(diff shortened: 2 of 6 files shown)


classes/dataflow.php
```diff
@@ -657,6 +657,16 @@ public function remove_step(step $step) {
         return $this;
     }
 
+    /**
+     * Called when steps are created, modified or removed.
+     *
+     * @throws \coding_exception
+     */
+    public function on_steps_save() {
+        $this->set('confighash', '');
```
keevan, line 666: Shouldn't this set the confighash to whatever it should be at the time? Or is this not changing for some other reason (e.g. another factor not considered)

classes/dataflow.php
```diff
@@ -657,6 +657,16 @@ public function remove_step(step $step) {
         return $this;
     }
 
+    /**
+     * Called when steps are created, modified or removed.
+     *
+     * @throws \coding_exception
```
keevan, line 663: Seems unnecessary to have the throws phpdoc here

remove-step.php
```diff
@@ -45,9 +45,13 @@
 // Start output.
 $dataflowdetailsurl = new moodle_url('/admin/tool/dataflows/view.php', ['id' => $step->dataflowid]);
 
+$dataflow = $step->get_dataflow();
+
 // Remove the step.
 $step->delete();
 
+$dataflow->on_steps_save();
+
```
keevan, line 54: Seems like the step should trigger the dataflow to recalculate its config hash, instead of calling the on_steps_save directly. 

Perhaps this should be added in some on_delete/on_remove method inside the step class instead? Internally calling the get_dataflow equivalent in the step class.

step.php
```diff
@@ -136,6 +136,8 @@
             $persistent->depends_on($dependson);
             $persistent->upsert();
         }
+        $persistent->get_dataflow()->on_steps_save();
```
keevan, line 139: This feels like it should be called somewhere along the lines of the upsert() call instead of directly as done here.

classes/dataflow.php
jaypha: I was following the current convention which was to simply clear it. I can change it to reset it here, as well as the other place where this value is cleared (I.

remove-step.php
jaypha: this was my original idea, but dataflows calls delete for every step when removing itself. Didn't unnecessary saves.

step.php
jaypha: Same as above

classes/dataflow.php
jaypha: I just read issue #143, where this behaviour was explicitly requested.

remove-step.php
keevan: It's true that there may be some unnecessary saves. 

Also consider the point where editing the dataflow shouldn't happen as often and is quite low traffic. Some extra DB calls per update aren't that big of an issue. Dataflows aren't being dynamically updated in a way where this poses an issue. And there's probably a more elegant way to resolve this issue but keep the call clean. 

Alternatively, needing to fetch the dataflow, delete the step, and call `$dataflow->on_steps_save()` because of a confighash field needing an update, requires some mental overhead (in remembering to do that), and is more likely to cause issues should someone forget to add it in a different location in the codebase.

keevan:
Hi @jaypha, the changes seem okay but CI is failing. Could you please have a look?

PR #586, opened 2022-09-12
Issue #577: Fix problem with run continuing after aborted.

### LINKED ISSUE #577: Calling engine::abort directly does not actually halt the execution.
When an exception is thrown, the iteration loop is interrupted. Calling abort() directly bypasses this.

- [ ] After abort is called, all steps should be in the state 'ABORTED'.
- [ ] Any loops should be halted with or without an exception.
- [ ] Any other safeguards to ensure that nothing gets done if the step is in state ABORTED.

jaypha:
Resolves #577
Add code to check for ABORTED status
Refactor names to minimize confusion.
Add flow abort step.

### DIFF
--- a/classes/local/execution/connector_engine_step.php
+++ b/classes/local/execution/connector_engine_step.php
@@ -53,7 +53,7 @@ class connector_engine_step extends engine_step {
                     }
                 } catch (\Throwable $thrown) {
                     $this->exception = $thrown;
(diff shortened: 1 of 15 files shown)


keevan (approved):
The changes look good. If there are any lingering issues they will be fixed in a follow up patch.

Thanks @jaypha

PR #588, opened 2022-09-13
feat: add reader_csv step

### LINKED ISSUE #573: CSV / Line reader step
- [ ] Option to specify if the file contains header on the first line
- [ ] Option to specify a custom delimiter

Very basic implementation for now, might leverage Moodle core API? Or defer that for later and get a basic working version implemented initially.

keevan:
Resolves #573

### DIFF
--- a/classes/local/step/reader_csv.php
+++ b/classes/local/step/reader_csv.php
@@ -1,0 +1,131 @@
+<?php
+// This file is part of Moodle - https://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_dataflows\local\step;
+
+use tool_dataflows\helper;
+use tool_dataflows\local\execution\iterators\iterator;
+use tool_dataflows\local\execution\iterators\dataflow_iterator;
+
+/**
+ * CSV reader step
+ *
+ * @package   tool_dataflows
+ * @author    Kevin Pham <kevinpham@catalyst-au.net>
+ * @copyright Catalyst IT, 2022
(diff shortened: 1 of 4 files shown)


classes/local/step/flow_s3.php
```diff
@@ -0,0 +1,299 @@
...
+/**
+ * S3 connector step type
+ *
+ * @package    tool_dataflows
+ * @author     Peter Burnettt <peterburnett@catalyst-au.net>
+ * @author     Kevin Pham <kevinpham@catalyst-au.net>
+ * @copyright  Catalyst IT, 2022
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class connector_s3 extends flow_step {
```
Peterburnett, line 30: Classname and comment here should be flow?

I also wonder if the guts of this should be pulled out to a trait so we can do multiple inheritance in both the flow and connector extending their base, but inheriting execute from s3_trait

Peterburnett: From memory, the form fields, side effects, execute, and resolve path should all be identical

classes/local/step/reader_csv.php
```diff
@@ -0,0 +1,131 @@
...
+                }
+
+                // At this point, if headers is false, then the file is empty, and
+                // so it should continue as if the file finished.
+                if ($strheaders === false) {
+                    return;
+                }
+
+                // Convert header string to an actual headers array.
+                $headers = str_getcsv($strheaders, $config->delimiter);
```
Peterburnett, line 79: I dont think this is correct for cases where you have a CSV with no headers, and you dont really need to add them manually in config. in this case, you would end up with 1 row mapped against another row as header => data which isnt right. It should just default to numerical keys. I think there should be a seperate option for "infer headers from file" which controls whether you read the first line as headers, or just use numeric keys

keevan: I think this is where unit tests would come in handy. 

Unfortunately, you wouldn't be able to use them in their numerical key format, as they are converted into objects and numeric keys on objects just _do not work_ with the expression parser we're using (currently). 

Originally, I had a thought about a setting for infer headers from file, and then I switched it to a custom "headers" field because I thought it would have been one or the other. 

That said, you are correct in saying, there might be a case where the first row is NOT header information and the "headers" might not need to be known or labelled in any meaningful way. Until there is such a case, I'm leaning towards giving the user the ability to set a custom header row if they choose to otherwise it takes the first row as headers. I think it makes it easier to configure, and if it's not clear from the _help I will add in this explanation.

classes/local/step/flow_s3.php
keevan: Yes this should be moved to a trait, but I don't think this is the appropriate PR for the change. I would like to move this file to the s3 PR instead.

PR #604, opened 2022-09-16
Issue #392: Improve handling of curl headers in connector_curl

### LINKED ISSUE #392: Fix and improve the curl headers feature

jaypha:
Resolves #392 

Refactor code to handle headers as plain text
Handle multiple headers in bash command.
Add bash escaping.
Other fixes.

### DIFF
--- a/classes/helper.php
+++ b/classes/helper.php
@@ -41,6 +41,9 @@ class helper {
     /** Location of the dependency section of the readme. */
     public const README_DEPENDENCY_LINK = 'https://github.com/catalyst/moodle-tool_dataflows#dependencies';
 
+    /** Regular expression for a valid HTTP header name. */
+    public const HTTP_HEADER_REGEX = '/^[A-Za-z0-9!#$%&\'*+-.=^_`|]+$/'; // phpcs:ignore moodle.Strings.ForbiddenStrings.Found
+
     /** CFG and SITE settings to be included as variables. */
     public const CFG_VARS = [
         'fullname',
@@ -259,5 +262,66 @@ class helper {
             return false;
         }
         return true;
+    }
+
+    /**
+     * Takes a string and turns it into an escaped bash string.
+     *
+     * @param string $content
+     * @return string
+     */
+    public static function bash_escape(string $content): string {
+        return "'" . str_replace("'", "'\''", $content) . "'";
+    }
+
+    /**
+     * Extract headers from given content in accordance with RFC2616.
+     *
+     * See https://www.rfc-editor.org/rfc/rfc2616.html#section-4.2
+     *
+     * @param string $content
+     * @return array|bool The array of header => value or false if the headers were invalid.
+     */
+    public static function extract_http_headers(string $content) {
+        $content = trim($content);
+        if ($content === '') {
+            return [];
+        }
+        $lines = explode(PHP_EOL, $content);
+        $headerlines = [];
+        $current = null;
+
+        // Look for multiline headers and contract them into a single line for each.
+        foreach ($lines as $line) {
+            if (trim($line) === '') {
+                return false; // Cannot have empty lines.
+            }
+            if (ctype_space(ord($line))) {
+                $current .= $line;
+            } else {
+                if (!is_null($current)) {
+                    $headerlines[] = $current;
+                }
+                $current = $line;
+            }
+        }
+        $headerlines[] = $current;
+
+        $headers = [];
+        // Check for validity.
+        foreach ($headerlines as $line) {
+            $pair = explode(':', $line, 2);
+            // There must be a value, even if it is an empty string (No value indicates that the ':' was not present.
+            if (!isset($pair[1])) {
+                return false;
+            }
+            list($header, $value) = $pair;
+            // Name must be of the correct syntax.
+            if (preg_match(self::HTTP_HEADER_REGEX, $header) !== 1) {
+                return false;
+            }
+            $headers[$header] = $value;
+        }
+        return $headers;
     }
 }
--- a/classes/local/step/connector_curl.php
+++ b/classes/local/step/connector_curl.php
@@ -34,7 +34,7 @@ class connector_curl extends connector_step {
     public const HTTP_ERROR = 400;
 
     /** @var int Time after which curl request is aborted */
-    protected $timeout = 60;
+    public const DEFAULT_TIMEOUT = 60;
 
     /**
      * Returns whether or not the step configured, has a side effect.
@@ -120,12 +120,10 @@ class connector_curl extends connector_step {
      * @param \MoodleQuickForm $mform
      */
     public function form_add_custom_inputs(\MoodleQuickForm &$mform) {
-        $jsonexample = [
-            'header-key' => 'header value',
-            'another-key' => '1234',
-        ];
-        $ex['json'] = \html_writer::nonempty_tag('pre', json_encode($jsonexample, JSON_PRETTY_PRINT));
-        $ex['yaml'] = '<pre>header-key: header value<br>another-key: 1234</pre>';
+        $ex = \html_writer::nonempty_tag(
+            'pre',
+            htmlspecialchars(get_string('connector_curl:header_format', 'tool_dataflows') . PHP_EOL)
+        );
 
         $urlarray = [];
         $urlarray[] =& $mform->createElement('select', 'config_method', '', [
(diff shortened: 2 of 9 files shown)


classes/helper.php
```diff
@@ -260,4 +263,64 @@ public static function obj_empty(\stdClass $obj): bool {
...
     }
+
+    /**
+     * Takes a string and turns it into an escaped bash string.
+     *
+     * @param string $content
+     * @return string
+     */
+    public static function bash_escape(string $content): string {
+        return "'" . str_replace("'", "'\''", $content) . "'";
```
keevan, line 274: Could we update this to use single quotes here instead?

https://docs.moodle.org/dev/Coding_style#Single_quotes

classes/helper.php
```diff
@@ -41,6 +41,9 @@ class helper {
     /** Location of the dependency section of the readme. */
     public const README_DEPENDENCY_LINK = 'https://github.com/catalyst/moodle-tool_dataflows#dependencies';
 
+    /** Regular expression for a valid HTTP header name. */
+    public const HTTP_HEADER_REGEX = "/^[A-Za-z0-9!#$%&'*+-.=^_`|]+$/"; // phpcs:ignore moodle.Strings.ForbiddenStrings.Found
```
keevan, line 45: Could we update this to use single quotes here instead?

https://docs.moodle.org/dev/Coding_style#Single_quotes

classes/helper.php
```diff
@@ -260,4 +263,64 @@ public static function obj_empty(\stdClass $obj): bool {
...
+     * Extract headers from given content in accordance with RFC2616.
+     *
+     * See https://www.rfc-editor.org/rfc/rfc2616.html#section-4.2
+     *
+     * @param string $content
+     * @return array|bool The array of header => value or false if the headers were invalid.
+     */
+    public static function extract_http_headers(string $content) {
+        $content = trim($content);
+        if ($content == '') {
```
keevan, line 287: It would be slightly better to change this to a strict comparison.

tests/tool_dataflows_connector_curl_test.php
```diff
@@ -220,6 +229,7 @@ public function test_validate_config() {
         $config = (object) [
             'curl' => $this->get_mock_url('/h5puuid.json'),
             'method' => 'get',
+            'headers' => "X-One:one\nX-Two:\nX-Three:three"
```
keevan, line 232: Could we add a trailing comma to the end of this multiline array please.

> Note the last item has a trailing comma left there which allows to extend the list of items later with a cleaner diff. For the same reason, it is better not to align the assignment operators. 
https://docs.moodle.org/dev/Coding_style#Wrapping_arrays

classes/helper.php
```diff
@@ -260,4 +263,64 @@ public static function obj_empty(\stdClass $obj): bool {
...
+            return [];
+        }
+        $lines = explode(PHP_EOL, $content);
+        $headerlines = [];
+        $current = null;
+
+        // Look for multiline headers and contract them into a single line for each.
+        foreach ($lines as $line) {
+            if (trim($line) == '') {
```
keevan, line 296: Similar to the above, it would be slightly better to change this to a strict comparison.

classes/local/step/connector_curl.php
```diff
@@ -213,24 +221,51 @@ public function execute($input = null) {
...
+        $headers = helper::extract_http_headers($config->headers);
+        if ($headers === false) {
+            throw new \moodle_exception(get_string('connector_curl:headers_invalid', 'tool_dataflows'));
         }
 
-        $options = ['CURLOPT_TIMEOUT' => $this->timeout];
+        // Add headers to bash command. Headers with no value are ended with a ';' in accordance with the man page.
+        if (!empty($headers)) {
+            foreach ($headers as $name => $value) {
+                if (trim($value) != '') {
```
keevan, line 245: Similar to the above, it would be slightly better to change this to a strict comparison.

classes/helper.php
```diff
@@ -260,4 +263,64 @@ public static function obj_empty(\stdClass $obj): bool {
...
+                }
+                $current = $line;
+            }
+        }
+        $headerlines[] = $current;
+
+        $headers = [];
+        // Check for validity.
+        foreach ($headerlines as $line) {
+            $pair = explode(':', $line, 2);
```
keevan, line 313: What about changing this to use the `list` functionality, to give the variables more meaningful names, which would make this easier to follow. 

e.g.
```
[$header, $value] = explode(':', $line, 2);
```

If that isn't possible, it would be good to set them to meaningful variables before they are used.

https://github.com/jupeter/clean-code-php#use-explanatory-variables

classes/local/step/connector_curl.php
```diff
@@ -213,24 +221,51 @@ public function execute($input = null) {
...
+        // See https://manpages.org/curl.
+        $dbgcommand = 'curl -s -X ' .  strtoupper($method) . ' ' . $config->curl;
 
-        if (!empty($config->timeout)) {
-            $this->timeout = (int) $config->timeout;
-            $dbgcommand .= ' --max-time ' . $config->timeout;
+        // Extract timeout.
+        $timeout = (int) $config->timeout;
+        if (empty($timeout)) {
+            $timeout = self::DEFAULT_TIMEOUT;
```
keevan, line 231: This may be a potential candidate for a ternary operator?
```suggestion
        $timeout = (int) $config->timeout ?: self::DEFAULT_TIMEOUT;
```

keevan (commented):
Other than various styling adjustments, I think the code is in mostly good shape.

I did get this during the upgrade, so potentially some adjustments need to be made here:
```
Warning: Invalid argument supplied for foreach() in /var/www/moodle/admin/tool/dataflows/db/upgrade.php on line 224
```

lang/en/tool_dataflows.php
```diff
@@ -376,8 +376,10 @@
 $string['connector_curl:sideeffects_help'] = 'Most read requests done via http GET do not have side effects ie they change state on the remote server, and generally any POST or PUT does have side effects. Curl calls with side effects are not actually executed in dry run mode.';
 $string['connector_curl:timeout'] = 'Timeout';
 $string['connector_curl:timeout_help'] = 'Time in seconds after which the request will abort, default is 60 seconds.';
-$string['connector_curl:field_headers_help'] = 'Headers should be in the following JSON format: {$a->json} You can also use the following YAML format: {$a->yaml}';
+$string['connector_curl:field_headers_help'] = 'Headers should be in valid HTTP format: {$a} One per line.';
 $string['connector_curl:output_response_result'] = 'Returns a string that contains the response to the request as text, or null if the request was unsuccessful or has not yet been sent.';
+$string['connector_curl:header_format'] = '<header>:<value>';
+$string['connector_curl:headers_invalid'] = 'Curl connector headers are invalid';
```
keevan, line 382: Based on other validation messages, this one probably should end with a full stop as well.

classes/helper.php
```diff
@@ -260,4 +263,65 @@ public static function obj_empty(\stdClass $obj): bool {
...
+            return [];
+        }
+        $lines = explode(PHP_EOL, $content);
+        $headerlines = [];
+        $current = null;
+
+        // Look for multiline headers and contract them into a single line for each.
+        foreach ($lines as $line) {
+            if (trim($line) === '') {
+                return false; // Cannot have empty lines.
```
keevan, line 297: I wonder if the intention here was meant to be a `continue` statement instead? For instance, you may want to space out and group the header information, potentially later on even adding in-line comments for each block.

Instead of returning false, it may be better to continue to the next line for processing. What do you think?

keevan (commented):
One question mentioned below, otherwise everything else looks good. Thanks @jaypha.

classes/helper.php
jaypha: It is designed to require RFC2616 compliance. It was asked for at https://github.com/catalyst/moodle-tool_dataflows/pull/466#discussion_r939495608.
I suppose it could be 'looser' in its requirements, but that would be a matter of discussion on the issue page.

keevan: Would skipping over empty lines be in contradiction with its requirements? It looks like this function was only preparing the would-be curl output, so the output should conform to the spec. 

Since the spec doesn't mention anything about new lines (from what I can tell), this is likely only a UX concern with regards to configuring the headers in the step form page. 

Perhaps flipping the question: if the user entered headers with whitespace between them (such as 2 new lines), would this explode during execution (or validation) and clearly inform the user to the reason?

jaypha: I've moved this discission to the issue page. https://github.com/catalyst/moodle-tool_dataflows/issues/392#issuecomment-1248951599

PR #621, opened 2022-09-29
Issue #37: Add an sftp connector

### LINKED ISSUE #37: Have a copy to and copy from sftp connector

jaypha:
Resolves #37


jaypha:
Need to add keyfile support.

jaypha:
Need to switch to use the phpseclib (see tool_etl), as php ssh is buggy.

jaypha:
Switched to use phpseclib3.
Tested for uploading and downloading.
Tested for password and keyfile authentication.
Tested for bad login
Tested for copy failure.

To test it yourself you will need to setup a local SFTP server. I used a docker container with atmoz/sftp

classes/local/step/connector_sftp.php
```diff
@@ -0,0 +1,293 @@
...
+            throw new \moodle_exception('connector_sftp:bad_host', 'tool_dataflows');
+        }
+
+        // At this point we need to disconnect once we are finished.
+        try {
+            if (!empty($config->hostpubkey) && $config->hostpubkey !== $hostpubkey) {
+                throw new \moodle_exception('connector_sftp:bad_hostpubkey', 'tool_dataflows');
+            } else {
+                $this->enginestep->set_var('hostpubkey', $hostpubkey);
+            }
```
keevan, line 226: No need for this to be in an else-block here, as the if block throws an exception.

classes/local/step/connector_sftp.php
```diff
@@ -0,0 +1,293 @@
...
+                $this->log("Uploading from '$tmppath' to '$config->target'");
+                if (!$sftp->put($topath, $tmppath, SFTP::SOURCE_LOCAL_FILE)) {
+                    throw new \moodle_exception('connector_sftp:copy_fail', 'tool_dataflows', '', $sftp->getLastSFTPError());
+                }
+            } else {
+                $this->log("Writing from '$tmppath' to '$topath'");
+                if (!@copy($tmppath, $topath)) {
+                    throw new \moodle_exception('connector_sftp:copy_fail', 'tool_dataflows', '', error_get_last()['message']);
+                }
+            }
```
keevan, line 273: Were there reasons it needed 2 hops instead of doing only one? For example, if I wanted to download a file from and to a particular location, it would download the file to a temp path, then copy it from that path to the destination path.

It seems like ops that could be avoided.

classes/local/step/connector_sftp.php
```diff
@@ -0,0 +1,293 @@
...
+    /**
+     * Executes the step
+     *
+     * Performs an SFTP call according to config parameters.
+     *
+     * @param  mixed|null $input
+     * @return mixed
+     */
+    public function execute($input = null) {
+        $config = $this->stepdef->config;
```
keevan, line 210: We should change this to use `$this->get_config()` here.

keevan (commented):
Doing a quick first pass, I'll have a more thorough test come Monday.

Thanks Jason, have a good weekend.

classes/local/step/connector_sftp.php
jaypha: This allows the code to be simpler and allows the reading and writing portions to be independant. To transfer directly would require more code overall.

classes/local/step/connector_sftp.php
```diff
@@ -0,0 +1,293 @@
...
+ * SFTP connector step type.
+ *
+ * Uses phpseclib. See https://phpseclib.com
+ *
+ * @package   tool_dataflows
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class connector_sftp extends connector_step {
```
keevan, line 33: Can we add an empty line here for whitespace. This would make it consistent the majority of other classes.

classes/local/step/connector_sftp.php
```diff
@@ -0,0 +1,293 @@
...
+
+    /**
+     * Return the definition of the fields available in this form.
+     *
+     * @return array
+     */
+    public static function form_define_fields(): array {
+        return [
+            'host' => ['type' => PARAM_TEXT, 'required' => true],
+            'port' => ['type' => PARAM_INT, 'required' => true, 'default' => 22],
```
keevan, line 59: I noticed we're using a default of 22 for the port. Could we make this a class constant just to make it less of a magic number? Thanks @jaypha

classes/local/step/connector_sftp.php
```diff
@@ -0,0 +1,302 @@
...
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class connector_sftp extends connector_step {
+
+    /** Shorthand sftp scheme for use in config. */
+    const SFTP_PREFIX = 'sftp';
+    /** Default port to connect to. */
+    const DEFAULT_PORT = '22';
```
keevan, line 38: Bit of a nitpick, I believe this doesn't need to be a string. 

But it's so small I personally don't mind which way it lands.

classes/local/step/connector_sftp.php
```diff
@@ -0,0 +1,302 @@
...
+     *
+     * Performs an SFTP call according to config parameters.
+     *
+     * @param  mixed|null $input
+     * @return mixed
+     */
+    public function execute($input = null) {
+        $config = $this->get_config();
+
+        $this->log("Connecting to $config->host:$config->port");
```
keevan, line 217:
```suggestion
        $this->log("Connecting to {$config->host}:{$config->port}");
```

I'm testing out an alternative way to weave it into the main string. What do you think? Would this suggestion make it easier to identify the individual variables?

classes/local/step/connector_sftp.php
```diff
@@ -0,0 +1,302 @@
...
+            }
+            $this->enginestep->set_var('hostpubkey', $hostpubkey);
+
+            if (!empty($config->privkeyfile)) {
+                // Use key authorisation.
+                $key = PublicKeyLoader::load(file_get_contents($config->privkeyfile), $config->password ?: false);
+            } else {
+                // Use password authorisation.
+                $key = $config->password;
+            }
```
keevan, line 238: Could we:
- pull this out into a private method, e.g. `get_key` or `prepare_key`, passing in the config.
- ensure file_get_contents is provided with a sanitized / verified path, which would be done by reaching in to the engine, and calling `resolve_path` (e.g. it is on the allowlist)

This will also remove an if-else statement from this main block, leaving it with something like
```
$key = $this->prepare_key($config);
```

classes/local/step/connector_sftp.php
```diff
@@ -0,0 +1,302 @@
...
+    /**
+     * Custom elements for editing the connector.
+     *
+     * @param \MoodleQuickForm $mform
+     */
+    public function form_add_custom_inputs(\MoodleQuickForm &$mform) {
+        $mform->addElement('text', 'config_host', get_string('connector_sftp:host', 'tool_dataflows'));
+        $mform->addElement('static', 'config_host_desc', '',  get_string('connector_sftp:host_desc', 'tool_dataflows'));
+        $mform->addElement('text', 'config_port', get_string('connector_sftp:port', 'tool_dataflows'));
+        $mform->setDefault('config_port', 22);
```
keevan, line 81: Could we update this to use the DEFAULT_PORT constant

classes/helper.php
```diff
@@ -82,6 +82,24 @@ public static function path_has_scheme(string $path): bool {
...
+    public static function path_is_scheme(string $path, string $scheme): bool {
+        $splitpath = explode('://', $path, 2);
+        if (count($splitpath) !== 2) {
+            return false;
+        }
+        if (is_null($scheme)) {
+            return true;
+        }
+        return strtolower($splitpath[0]) === strtolower($scheme);
+    }
```
keevan, line 101: Thanks for pulling this out.

Could we instead use this:
```
substr($path, 0, strlen($scheme)) === $scheme
```

this is based on the solution here: https://stackoverflow.com/a/2790919/4757871

classes/local/step/connector_sftp.php
keevan, line 38: Could we please change this to a number instead of a string?

classes/local/step/connector_sftp.php
keevan, line 81: Could we use the default port constant here. `self::DEFAULT_PORT`

classes/local/step/connector_sftp.php
```diff
@@ -0,0 +1,302 @@
...
+                    throw new \moodle_exception('connector_sftp:copy_fail', 'tool_dataflows', '', $sftp->getLastSFTPError());
+                }
+                return true;
+            }
+
+            // Upload to remote.
+            $this->log("Uploading from '$config->source' to '$config->target'");
+            if (!$sftp->put($targetpath, $sourcepath, SFTP::SOURCE_LOCAL_FILE)) {
+                throw new \moodle_exception('connector_sftp:copy_fail', 'tool_dataflows', '', $sftp->getLastSFTPError());
+            }
```
keevan, line 282: This looks much better. I can see the logical groups / use cases.

Are we able to:
- move the comment from inside the single if statement above the if instead of being inside the block? Reasoning is I tend to read top down and only dive into the blocks if they are relevant. The condition itself tells me nothing about what it's trying to do / why it's there, and so a leading comment would help with this
- move the handling within each block into its own method. Based on the comments, the names could be `download`, `upload`, and `remote_copy`.

classes/local/step/connector_sftp.php
```diff
@@ -0,0 +1,302 @@
...
+ * @package   tool_dataflows
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class connector_sftp extends connector_step {
+
+    /** Shorthand sftp scheme for use in config. */
+    const SFTP_PREFIX = 'sftp';
+    /** Default port to connect to. */
```
keevan, line 37: Could we add a whitespace above this comment?

classes/local/step/connector_sftp.php
```diff
@@ -0,0 +1,302 @@
...
+        if ($hostpubkey === false) {
+            throw new \moodle_exception('connector_sftp:bad_host', 'tool_dataflows');
+        }
+
+        // At this point we need to disconnect once we are finished.
+        try {
+            if (!empty($config->hostpubkey) && $config->hostpubkey !== $hostpubkey) {
+                throw new \moodle_exception('connector_sftp:bad_hostpubkey', 'tool_dataflows');
+            }
+            $this->enginestep->set_var('hostpubkey', $hostpubkey);
```
keevan, line 230: It might be a balancing at of sorts, but security wise, would it be better if it only populates this if the setting is currently empty, otherwise it keeps the current value even if it ends up being invalid. This could mitigate concerns such as MITM attacks, and an "easy" way to re-sync it, would be to let the user clear the value so it applies the current working entry again.

What do you think? Should we make this more secure-ish by default?

keevan: Ah sorry, just noticed the if check right above this checks and blocks this from happening.

keevan (commented):
Aside from the comments provided, it passes functional testing.



Thanks @jaypha.

jaypha:
I've made the fixes. I didn't add the new functions. They are unnecessary and I no longer have time.

keevan (approved):
Thanks @jaypha for spending time on this. It is in a good state to be merged so I'll do that.

I'll make the remaining adjustments in a follow up PR.

Issue #614, opened 2022-09-21
Variables code needs to be refactored


jaypha:
There are several issues with the variables code.

- It is distributed amongst several classes, sharing codespace with code that does other things.
- There is repetition of code fragments that do the same thing. (E.g dataflows::variables() and step_form::get_available_references() both construct the step->vars subtree).
- There is unnecessary execution. (E.g. step::get_export_data() is called at least twice for each step).
- There is a mixture of objects and associative arrays to store structures, causing confusion.
- It is causing performance issues in field installations.

The code should be refactored to have a separate set of classes to handle variables, including the structure and expression substitution.

jaypha:
A branch containing an experimental alternative is available at https://github.com/catalyst/moodle-tool_dataflows/tree/614-experimental-alternative.
It is non-functional, but should give an idea of what I had in mind.

keevan:
Some more concerns which may or may not be easily resolved:
- parser usage in various step types (using `$this->get_config()` might do what you need out of the box). 
- needing to evaluate the whole variable tree and all its expressions regardless of complexity, nesting or other. It would be nice if there was a way to simply specify what you need, and get it and only it. This might actually eliminate a lot of the performance concerns since you don't need to build the whole tree.
- nested expressions
- undefined property warning/notices for `fields.that.might.be.really.deep.or.not.exist.yet`. Handling this in a class might allow us to handle this more gracefully.

PR #622, opened 2022-09-29
Issue #614: Introduce a dirty flag for variables.

jaypha:


classes/dataflow.php
```diff
@@ -47,6 +47,12 @@ class dataflow extends persistent {
     /** @var bool Set to true when the dataflow is in the process of deleting. */
     private $isdeleting = false;
 
+    /** @var bool If true, then the variables tree will be rebuilt when get_variables() is called.  */
+    private $rebuildvariables = true;
```
keevan, line 51: This matches the method name which could get a bit messy.

Are we able to change this to be something else such as `shouldrebuildvariables` (and maybe the variable tree should be its own class instead - to hold its own internal state?)

keevan (commented):
A minor rename comment, and a sugquestion.

classes/dataflow.php
jaypha: Changing the variable name is fine.
Putting into it's own class? Well, yeah, that's kinda the point of the issue, but that is VERY complicated, and this is supposed to be a quick, quick fix.

keevan (approved):
There's a clear performance improvement from local testing, and the flows I've tested with seem to run as expected.

So approving this one.

PR #626, opened 2022-09-29
[Issue #625] feat: add directory file count step

keevan:


classes/parser.php
```diff
@@ -165,7 +165,7 @@ private function internal_evaluator(string $string, array $variables, ?callable
                 // the wrong spot.
                 try {
                     $parsethis = trim($match['expression']);
-                    error_reporting(E_ALL & ~E_NOTICE);
+                    error_reporting(E_ALL & ~E_NOTICE & ~E_WARNING);
```
keevan, line 168: This resolves a small issue I noticed:

image

keevan: This is little bit of a limitation of the expression parser. The currently doesn't have any way of specifying a fallback, and deeply nested objects would result in a similar issue. As this is coming from referencing the output of a step in a future step, whilst we could default it to some value such as now it probably wouldn't be correct either (null could be expected)

The variables refactor would actually resolve this issue by giving control back to the class level to handle such conditions.

I hope that clarifies things.

keevan: I've moved the parser change to a different PR / issue and I think it's an issue with php8.0+

#629

PR #630, opened 2022-09-30
fix(parser): warning for undefined properties

### LINKED ISSUE #629: Warnings for undefined properties, should be suppressed in the parser for now
Will be resolved probably with the variables refactor? But currently due to perhaps PHP version differences, the undefined properties notices are now warnings, which weren't suppressed previously.

image

keevan:
This used to come up as a notice, but now has been changed to a warning which was not suppressed. This removes warnings from display also during parsing.

Resolves #629

### DIFF
--- a/classes/parser.php
+++ b/classes/parser.php
@@ -165,7 +165,7 @@ class parser {
                 // the wrong spot.
                 try {
                     $parsethis = trim($match['expression']);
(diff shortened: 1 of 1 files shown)


keevan:
Going to close this as the variable handling and logic has likely changed the behaviour of this. Which will require a different (probably) change to be done.

PR #648, opened 2022-10-07
Issue #647: Refactor parser class to be a singleton.

### LINKED ISSUE #647: Make parser a singelton
Slightly faster, and if we ever want to do things like registering functions, it will be easier.

jaypha:
Resolves #647

### DIFF
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -187,7 +187,7 @@ class dataflow extends persistent {
             $step = (object) $step;
         }
         $steps = (object) $steps;
-        $parser = new parser();
+        $parser = parser::get_parser();
 
         $variables = ['steps' => $steps];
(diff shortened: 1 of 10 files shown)


keevan:
@jaypha thanks for preparing the PR.

Can we get a profile or graph to demonstrate the performance benefit here as described in the issue?

jaypha:
I would not be that significant an improvement. Doing some experiments showed that using a static get is roughly 7-8% faster than using 'new', but don't expect that to result in the same improvement overall. I don't think it will show up in any profiles.

jaypha:
I ran some profiles, there is nothing to see. This change is still a good thing as it cleans the code somewhat.

keevan (approved):

Issue #646, opened 2022-10-05
Static variables


jaypha:
There is an ability to update a configuration setting of a step during a run. This is currently only used by reader_sql to update the countervalue during a run.

This is not good as it poisons the dataflow definition. Ideally, the dataflow definition should be immutable to a run.

It would be better to have a separate static variables store. It would have the following features.
- [ ] Key-value database store.
- [ ] Declare variables as static in the step type.
- [ ] Reads in static variables at the beginning of a run.
- [ ] Saves variables at end of run (maybe as it goes).
- [ ] Can see/edit what the current values are.
- [ ] Import/export.

Anything else?

PR #652, opened 2022-10-13
614 variables II

### LINKED ISSUE #614: Variables code needs to be refactored

jaypha:
Resolves #614

New module to handle variables.
More efficient management of expression handling.
Should get substantial speed improvements.

image

### DIFF
--- a/tests/local/execution/tool_dataflows_basic_execution_test.php
+++ b/tests/local/execution/tool_dataflows_basic_execution_test.php
@@ -16,16 +16,16 @@
 
 namespace tool_dataflows\local\execution;
 
+use Symfony\Component\Yaml\Yaml;
 use tool_dataflows\local\execution\engine;
 use tool_dataflows\dataflow;
 use tool_dataflows\step;
+use tool_dataflows\test_dataflows;
 
 defined('MOODLE_INTERNAL') || die();
 
 // This is needed. File will not be automatically included.
-require_once(__DIR__ . '/array_in_type.php');
-// This is needed. File will not be automatically included.
-require_once(__DIR__ . '/array_out_type.php');
+require_once(__DIR__ . '/../../test_dataflows.php');
 
 /**
  * Unit tests for the execution engine
@@ -57,22 +57,10 @@ class tool_dataflows_basic_execution_test extends \advanced_testcase {
      */
     public function test_in_and_out() {
         // Create the dataflow.
-        $dataflow = new dataflow();
-        $dataflow->name = 'two-step';
-        $dataflow->enabled = true;
-        $dataflow->save();
-
-        $reader = new step();
-        $reader->name = 'reader';
-        $reader->type = 'tool_dataflows\local\execution\array_in_type';
-        $dataflow->add_step($reader);
-
-        $writer = new step();
-        $writer->name = 'writer';
-        $writer->type = 'tool_dataflows\local\execution\array_out_type';
-
-        $writer->depends_on([$reader]);
-        $dataflow->add_step($writer);
+        [$dataflow, $steps] = test_dataflows::sequence([
+            'reader' => array_in_type::class,
+            'writer' => array_out_type::class,
+        ]);
 
         // Define the input.
         $json = '[{"a": 1, "b": 2, "c": 3}, {"a": 4, "b": 5, "c": 6}]';
@@ -87,6 +75,32 @@ class tool_dataflows_basic_execution_test extends \advanced_testcase {
         ob_get_clean();
 
         $this->assertEquals(json_encode(array_in_type::$source), json_encode(array_out_type::$dest));
+        $this->assertEquals(engine::STATUS_FINALISED, $engine->status);
+    }
+
+    public function test_direct_and_out() {
+        // Create the dataflow.
+        [$dataflow, $steps] = test_dataflows::sequence([
+            'reader' => direct_in_type::class,
+            'writer' => array_out_type::class,
+        ]);
+
+        // Define the input.
+        $source = [
+            ['a' => 9, 'b' => 8, 'c' => 7],
+            ['a' => 6, 'b' => 5, 'c' => 4],
+        ];
+
+        $steps['reader']->config = Yaml::dump(['source' => $source]);
+        array_out_type::$dest = [];
+
+        // Execute.
+        ob_start();
+        $engine = new engine($dataflow);
+        $engine->execute();
+        ob_get_clean();
+
+        $this->assertEquals(json_encode($source), json_encode(array_out_type::$dest));
         $this->assertEquals(engine::STATUS_FINALISED, $engine->status);
     }
 }
--- a/tests/local/execution/tool_dataflows_basic_execution_test.php
+++ b/tests/local/execution/tool_dataflows_basic_execution_test.php
@@ -78,6 +78,11 @@ class tool_dataflows_basic_execution_test extends \advanced_testcase {
         $this->assertEquals(engine::STATUS_FINALISED, $engine->status);
     }
 
+    /**
+     * Tests using direct_in step type.
+     *
+     * @covers \tool_dataflows\local\execution\direct_in_type
+     */
     public function test_direct_and_out() {
         // Create the dataflow.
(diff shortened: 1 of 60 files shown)


tests/local/execution/tool_dataflows_basic_execution_test.php
```diff
@@ -57,22 +57,10 @@ protected function setUp(): void {
...
-        $reader->type = 'tool_dataflows\local\execution\array_in_type';
-        $dataflow->add_step($reader);
-
-        $writer = new step();
-        $writer->name = 'writer';
-        $writer->type = 'tool_dataflows\local\execution\array_out_type';
-
-        $writer->depends_on([$reader]);
-        $dataflow->add_step($writer);
+        [$dataflow, $steps] = test_dataflows::sequence([
```
keevan, line 60: This is great. I think the sequence method should become a main feature, and allow anyone to quickly build a dataflow or parts of one by typing out / filtering and selecting what they want in order. 

Then they can configure each step as required later.

tests/test_dataflows.php
```diff
@@ -0,0 +1,127 @@
...
+ * @package   tool_dataflows
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class test_dataflows {
+    /**
+     * Create a two step reader-writer using the static array in and array out test steps.
+     *
+     * @return array [<dataflow>, <steps]
```
keevan, line 42:
```suggestion
     * @return array [<dataflow>, <steps>]
```

tests/test_dataflows.php
```diff
@@ -0,0 +1,127 @@
...
+        $reader->name = 'reader';
+        $reader->type = 'tool_dataflows\local\execution\array_in_type';
+        $dataflow->add_step($reader);
+
+        $writer = new step();
+        $writer->name = 'writer';
+        $writer->type = 'tool_dataflows\local\execution\array_out_type';
+
+        $writer->depends_on([$reader]);
+        $dataflow->add_step($writer);
```
keevan, line 60: Could use the new sequence creation method here to make the test method shorter. 

Just a thought, not required.

tests/test_dataflows.php
```diff
@@ -0,0 +1,127 @@
...
+        return [$dataflow, ['reader' => $reader, 'writer' => $writer]];
+    }
+
+    /**
+     * Create a three step reader-callback-writer dataflow.
+     *
+     * @param string $readertype
+     * @param string $writertype
+     * @param callable $fn
+     * @return array [<dataflow>, <steps]
```
keevan, line 71:
```suggestion
     * @return array [<dataflow>, <steps>]
```

tests/tool_dataflows_variables_dataflow_test.php
```diff
@@ -0,0 +1,173 @@
...
+     */
+    public function test_persistence() {
+        [$dataflow, $steps] = test_dataflows::array_in_array_out();
+
+        // Define the input.
+        $json = '[{"a": 1, "b": 2, "c": 3}, {"a": 4, "b": 5, "c": 6}]';
+
+        array_in_type::$source = json_decode($json);
+        array_out_type::$dest = [];
+        $dataflow->set('vars', Yaml::dump(['a' => '${{steps.writer.record.c}}']));
```
keevan, line 161: I believe this may have some unintended side effects, at least from a particular dataflow I'm using it in.

I have a dataflow which sets a bunch of AWS variables at the dataflows vars level. They are set as expressions. Would this change mean that after the initial run, they would be set to their evaluated value instead of being retained as an expression? This might be okay within the same lifecycle of a dataflow, but in terms of keeping the value in the database, there should be a different way to explicitly update these values.

Is this something that would be difficult to support? What were other cases where a dataflow vars should be evaluated and persisted?

tests/tool_dataflows_variables_new_test.php
```diff
@@ -0,0 +1,259 @@
...
+        $vars->set('a.b', 'double ${{a.x}} and ${{a.z}}');
+        $vars->set('a.x', 'six');
+        $vars->set('a.z', 'four');
+        $this->assertEquals('double six and four', $vars->get('a.b'));
+        $vars->set('a.z', 'five');
+        $this->assertEquals('double six and five', $vars->get('a.b'));
+    }
+
+    /**
+     * Test expressions that reference expressions that themselves referenece other values.
```
keevan, line 157:
```suggestion
     * Test expressions that reference expressions that themselves reference other values.
```

keevan (commented):
Hi @jaypha, it's looking super duper!

I like how much of the previous variables handling / mangling has been removed
from the other classes (local/execution/*, dataflow, steps) and has been moved
to its own class. This removes much of the noise and complexity that was
sprinkled all over.

I wanted to also highlight the fact that you've made an enormous effort to
improve readability, such as when you split it out to a $stepvar variable,
instead of keeping it chained, even if if technically could be on a single
line. This does help readability and that is greatly appreciated.

Thanks for going through and filling in the missing define_outputs for all the
various step types as well as adding very thorough tests for all the new
variable changes.

The variable localisation changes are very nice. I do like the fact that you've
got it working for local subtrees as well as absolute paths.

I was going to merge this but there were a couple of small fixes and one main
issue regarding dataflow.vars persistence I would like to address before going
ahead.

tests/tool_dataflows_variables_dataflow_test.php
keevan: Just updating to confirm that this will be an issue since I'm setting something like the following in vars, and referencing it as a path in a step.
```
archivelocation: 'archive/${{ date(''YmdHi'',time()) }}'
```

tests/local/execution/tool_dataflows_basic_execution_test.php
jaypha: I got this idea when I found myself repeatedly cutting and pasting dataflow creation scripts. There is a problem in that any 'required' field needs to be set before saving, which limited it's ability to be used.
I think this is worth getting to work, probably as an issue for itself.

tests/test_dataflows.php
jaypha: done

tests/test_dataflows.php
jaypha: done

tests/tool_dataflows_variables_dataflow_test.php
jaypha: I noticed this myself, but it was my understanding that this is the way it behaved before. I personally am not a fan of updating the database during a run. I wrote an issue proposing the idea of 'static variables' (#646), Variables that are stored in it's own persistent store, and read in at the beginning of each run. I'll have another look at this, but we may need to discuss what to do.

tests/tool_dataflows_variables_new_test.php
jaypha: done

tests/tool_dataflows_variables_dataflow_test.php
jaypha: I've had another look, and it is now apparent that the database would only be updated when the variable is explicitly set, not expanded. There is currently no step that calls either `set_dataflow_var` or `set_global_var`, so my incling is to remove this for now, and have it resolved in a followup issue. What do you think?

keevan: Agreed.

keevan (approved):
Thanks Jason, this all looks good to merge.

:rocket:

Issue #663, opened 2022-11-04
A reader step whose group is 'trigger' does not execute the first item in an iterator.


matthewhilton:
While working on #31 I came across a behaviour that I'm not sure is a bug or if i'm doing something I shouldn't.

Basically I have a `reader_step` that I want to appear as a trigger, so i've set the `get_group` function to return `trigger`. This makes it appear as a trigger in the UI which is what I want.

I then have an iterator in the step that reads some stuff from the database. But I found that when i set the group to `trigger` it only calls the execute function N - 1 times. For e.g. if returns 4 records, then the execute function is only called 3 times, the first item in the iterator is always skipped. If there is just 1 record, nothing is run. Note I checked if it was making an adhoc task for the first one, but it isn't.

If I change the group type to `reader`, they all execute as expected.

I think either get_group should:
1. Be more clear that its not just UI but also logic.
2. Refactor the logic that is relying on `get_group`. For e.g. if its expecting an iterator but checking if its group is a `reader` before getting the iterator, it should just check for the existence of an iterator instead (this might not be the case, just a random idea)

matthewhilton:
https://github.com/catalyst/moodle-tool_dataflows/blob/05a7ed801bd2cb7674a3fd16530abdd70fb597dc/classes/local/execution/iterators/dataflow_iterator.php#L188-L192

The first item is not being called here. There issue is likely further up the chain.

matthewhilton:
I've worked it out.

Here it checks if `should_pull_next` as a way to NOT increment the iterator on the first run.

https://github.com/catalyst/moodle-tool_dataflows/blob/05a7ed801bd2cb7674a3fd16530abdd70fb597dc/classes/local/execution/iterators/dataflow_iterator.php#L159-L170

However this function under the hood 

https://github.com/catalyst/moodle-tool_dataflows/blob/05a7ed801bd2cb7674a3fd16530abdd70fb597dc/classes/local/execution/iterators/dataflow_iterator.php#L132-L140

Is checking the grouptype is `!== readers`, which is why any non-reader steps are skipping the first iteration in in iterator.

matthewhilton:
Confirmed that removing the steptype check in `should_pull_next` fixes the bug. However, I am unsure if this may lead to regressions elsewhere without understanding the reason why this check was there in the first place.

keevan:
Internal comms but here for completness:

[...]

Issue #639, opened 2022-10-04
Missing method reschedule_or_queue_adhoc_task in core\task\manager (was added after 3.7)


keevan:
https://github.com/catalyst/moodle-tool_dataflows/blob/629a7f552f3cb7fd5121ae66017ba75f89df632b/classes/task/process_dataflows.php#L60

So 35, 36 and Totara won't have this function available out of the box. This will require some backport patches to be added or alternative handling added.

PR #664, opened 2022-11-06
31 event trigger

### LINKED ISSUE #31: Moodle event as a source trigger step
This is an interesting idea, you can setup a flow which is triggered by a single event. This would create a stream which only has a single item in it. To some a large extent if this is all implemented really well tool_trigger would be redundant. 

If it is purely single instance then it is easy, we handle the event, put it into a tiny buffer, then call the flow execution manager which then pulls it through the flow. Possibly there is a case where many identical events are emitted by the same script as we could handle them as an actual stream. But this may be tricky as we don't want the flow manager to take over control of the script and be trying to pull more events out when it can't. OR even better, we can tell the flow manager to just run as much as it can, and then we'll pass execution back. We would probably need another method like 'is_waiting'

This will have some interesting edge cases to make sure we handle with filters. Because of a filter we can't know if there is items available until we actually ask for it. Because the filter won't know if it will return a record or not until it's gotten its data. So might fundamentally break us from using the iterator interface and we might need to introduce our own one. We can either return an item, or we can throw a 'flow is blocked' exception which exactly which source it is blocked on. We'd then want a resume flow method.

An extension of this concept is that all events are queued and there is config to say how much we should throttle / debounce the activation of the workflow. This means we'd pull out of the event queue at a latency no higher than X for event coming so we have some QoS around the overall latency but we are not smashing the dataflow manager with a high turnover of workflow instances.

### LINKED ISSUE #665: A bad webservice on another plugin causes the webservice writer to throw exceptions.
If another plugin is installed that has a dodgy web service definition, when you go to add a webservice writer it will throw an exception

```
Coding error detected, it must be fixed by a programmer: Cannot find file /externallib.php with external function implementation

[More information about this error](http://docs.moodle.org/39/en/error/debug/codingerror)
Debug info:
Error code: codingerror
Stack trace:

    line 87 of /lib/externallib.php: coding_exception thrown
    line 106 of /admin/tool/dataflows/classes/local/step/flow_web_service.php: call to external_api::external_function_info()
    line 281 of /admin/tool/dataflows/classes/local/step/base_step.php: call to tool_dataflows\local\step\flow_web_service->form_add_custom_inputs()
    line 108 of /admin/tool/dataflows/classes/form/step_form.php: call to tool_dataflows\local\step\base_step->form_setup()
    line 214 of /lib/formslib.php: call to tool_dataflows\form\step_form->definition()
    line 110 of /lib/classes/form/persistent.php: call to moodleform->__construct()
    line 87 of /admin/tool/dataflows/step.php: call to core\form\persistent->__construct()
```

I think if a webservice is throwing an exception, it should just be skipped and excluded from the list (and maybe log some debugging)

### LINKED ISSUE #666: Update references to `set_variables` in the docs
Since https://github.com/catalyst/moodle-tool_dataflows/pull/652/files `set_variables` has been removed. However, there are still some references to it in the docs that need to be updated.

matthewhilton:
PR to add event triggers. 

### New event trigger
This event trigger works the following:

1. The plugin (tool_dataflows) listens to EVERY moodle event.
2. When it observes a Moodle event, it checks if any `event_reader` steps are configured to listen to this event.
3. If they are, it stores the event data in the `tool_dataflows_events` table
4. Then depending on the execution policy of the event_reader step, it will either:
4.1. Execute the dataflow immediately
4.2. Create an adhoc task to execute the dataflow
4.3. Wait until the next CRON call to dataflow to run it.
5. When the dataflow is run, it gets the record from the database and loads it as an array iterator and makes the `event` variable available to later steps.

## What has been changed
- `nextruntime` is now optional (since now event triggered dataflows don't have a nextruntime like cron ones do)
- Handle bad webservices causing the form to be unusable
- Updated docs around the recent variable handling change
- Refactored the calling of a dataflow so it can be reused in a couple of new places

## Things left to do:

- [x] Unit tests - will do once the structure is settled
- [x] A user guide for how to use the event trigger
- [x] Codechecker, etc...

## Some areas I would like comment / feedback / ideas on

- Should this be a trigger, reader, or a mix of both - currently it's a reader but styled as if its a trigger.
- Is a race condition possible the current db queue setup? I had a good think and I don't think this is possible even with concurrent running, but i'm not 100% confident.
- General comments regarding the way everything has been setup. I'll admit OOP isn't my strongest suite but happy for any tips

## Closes issues
Closes #31 

Closes #665 

Closes #666 

## Example dataflow
See example dataflow below (captures course viewed event, then calls course_get_contents webservice)
Just replaced the user with the admin user on your site
```
name: 'Test dataflow'
config:
  enabled: '1'
  concurrencyenabled: '0'
steps:
  course_viewed:
    name: 'Course viewed'
    type: tool_dataflows\local\step\trigger_event
    config:
      eventname: \core\event\course_viewed
      executionpolicy: adhocqueued
    vars:
      courseid: '${{ event.courseid }}'
  debugging_writer:
    name: 'Debugging writer'
    depends_on:
      1: course_viewed
    type: tool_dataflows\local\step\writer_debugging
  flow_web_service:
    name: 'Flow web service'
    depends_on: debugging_writer
    type: tool_dataflows\local\step\flow_web_service
    config:
      webservice: core_course_get_contents
      user: admin
      parameters: 'courseid: ${{steps.course_viewed.event.courseid}}'
      failure: abortflow
```


matthewhilton:
This should be ready for an initial review now
Not 100% complete but would like feedback regarding the setup before continuing

classes/local/event_processor.php
```diff
@@ -0,0 +1,173 @@
...
+                    // Only record that this event was triggered.
+                    // The next CRON call will then execute the event.
+                    self::add_triggered_event($dataflow->id, $step->id, json_encode($eventdata));
+                    break;
+                default:
+                    throw new \coding_exception("Unknown execution policy");
+            }
+        }
+
+        return;
```
keevan, line 152: Could we remove this? An empty return statement at the end of a function isn't needed.

classes/local/event_processor.php
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    /**
+     * Finds trigger steps that are listening to this event.
+     * @param string $eventname
+     * @return array array of steps
+     */
+    private static function get_listening_triggers(string $eventname): array {
+        // Get all event triggers.
+        global $DB;
```
keevan, line 162:
```suggestion
        global $DB;
        // Get all event triggers.
```
Just flipping the order. Globals have consistently stuck to the top, and they generally don't have a comment prior to declaration.

classes/local/event_processor.php
```diff
@@ -0,0 +1,173 @@
...
+        $type = $DB->sql_compare_text('tool_dataflows\local\step\reader_event');
+        $eventsteps = \tool_dataflows\step::get_records(['type' => $type]);
+
+        // Check which ones are configured to listen to this event.
+        $stepslistening = array_filter($eventsteps, function($step) use ($eventname) {
```
keevan, line 167:
```suggestion
        $stepslistening = array_filter($eventsteps, function ($step) use ($eventname) {
```

classes/local/step/reader_event.php
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        $variables = $this->get_variables();
+        $variables->set('event', $input->eventdata);
+
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
```
keevan, line 47: Usage of `@inheritdoc` is forbidden as per https://docs.moodle.org/dev/Coding_style#Functions

classes/local/step/reader_event.php
```diff
@@ -0,0 +1,223 @@
...
+    public function define_outputs(): array {
+        return ['event' => get_string('reader_event:variable:event', 'tool_dataflows')];
+    }
+
+    /**
+     * Returns the event iterator. Reads events from the database queue table and returns them as an array iterator.
+     * @return iterator Array iterator containing the events queued for this dataflow.
+     */
+    public function get_iterator(): iterator {
+        $dataflowid = $this->enginestep->engine->dataflow->get('id');
```
keevan, line 58: This chain makes me cry.

Do you know of a more concise way to reference this? If not, don't worry about it for now.

classes/local/step/reader_event.php
```diff
@@ -0,0 +1,223 @@
...
+        $eventarray = event_processor::get_events_for_dataflow($dataflowid);
+
+        // Re-parse the stringified JSON event data for each.
+        $eventarray = array_map(function($e) {
```
keevan, line 62:
```suggestion
        $eventarray = array_map(function ($e) {
```

classes/local/step/reader_event.php
```diff
@@ -0,0 +1,223 @@
...
+            $e->eventdata = json_decode($e->eventdata);
+            return $e;
+        }, $eventarray);
+
+        return new dataflow_iterator($this->enginestep, new \ArrayIterator($eventarray));
```
keevan, line 67: Bit of a microoptimisation, but potentially could move the handling for decoding the json in another private method, and use that as a generator function.

See: https://github.com/catalyst/moodle-tool_dataflows/blob/e947e37/classes/local/step/reader_csv.php#L84

keevan (commented):
Submitting feedback in parts.

I'm going to skip most phpdoc related ones, as they are indicated on the CI. But noted the ones which might not have appeared.

classes/local/step/reader_event.php
```diff
@@ -0,0 +1,223 @@
...
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class reader_event extends reader_step {
+    /**
+     * Executes the reader event by consuming the event from the queue and add the event data to the variables for this step.
+     */
+    public function execute($input = null) {
+        if (!$this->enginestep->engine->isdryrun) {
+            event_processor::consume_event($input->id);
+        }
```
keevan, line 38: How does a dry run event data get passed to the run at the moment? I scanned the class and didn't see a form field for this sort of data, and below it seems to be referenced.
```
$variables->set('event', $input->eventdata);
```

version.php
```diff
@@ -25,7 +25,7 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-$plugin->version = 2022102600;
+$plugin->version = 2022102601;
```
keevan, line 28: This version needs to be bumped to `..03` to match the upgrade script. During functional testing it complains the events table is missing 
```
Exception encountered in event observer '\tool_dataflows\local\event_processor::process_event': Table "tool_dataflows_events" does not exist
```

db/upgrade.php
```diff
@@ -205,7 +205,7 @@ function xmldb_tool_dataflows_upgrade($oldversion) {
         upgrade_plugin_savepoint(true, 2022091200, 'tool', 'dataflows');
     }
 
-    if ($oldversion < 2022091600) {
+    if ($oldversion < 2022102601) {
```
keevan, line 208: This appears to be a temporary change that should be reverted.

db/install.xml
```diff
@@ -71,6 +71,17 @@
         <KEY NAME="dataflow" TYPE="unique" FIELDS="dataflowid, stepid"/>
       </KEYS>
     </TABLE>
+    <TABLE NAME="tool_dataflows_events" COMMENT="Times for last run, and the next scheduled run.">
```
keevan, line 74: Is the description for this table correct?

keevan (commented):
A few more comments.

Thanks for making the changes @matthewhilton, I'm still a bit on the fence regarding whether or not it should appear under the reader list or trigger list.

image

In its current state, it can't act like a typical trigger-connector step. I think the sweet spot would be to tweak the handling of the dataflow iterator to allow for trigger steps that are "flows" to act like readers. This is because you wouldn't have a step that comes before this.

ref: https://github.com/catalyst/moodle-tool_dataflows/issues/663

keevan (commented):
It would be good if we could add some unit tests as well. This should work and be testable for the 'run immediately' type of execution, and can verify the event data comes through as expected.

classes/local/event_processor.php
```diff
@@ -0,0 +1,173 @@
...
+
+                    break;
+                case self::EXECUTE_ADHOC:
+                    // Record the triggered event to save the event data.
+                    self::add_triggered_event($dataflow->id, $step->id, json_encode($eventdata));
+
+                    // Create adhoc task to execute.
+                    $record = [ // TODO refactor ?
+                        'dataflowid' => $dataflow->id
+                    ];
```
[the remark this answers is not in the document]
matthewhilton: I think maybe I was thinking of adding a helper function e.g. `create_adhoc_task_from_id` so you don't have to create a object with just 1 property

classes/local/step/reader_event.php
matthewhilton: Oh interesting, I saw it used elsewhere in the plugin so I just used it here. I can write them out though no biggie :smile:

version.php
matthewhilton: Yes that is correct. I'll fix up the versions once its all ready to go

matthewhilton:
Test case ideas:

- [ ] dataflow disabled, shouldn't run / capture events
- [ ] the different execution policies
- [ ] form fuzzing
- [ ] config validation
- [ ] delete step, events recorded are deleted
- [ ] get events list (ensure it works and doesn't throw exception)
- [ ] get iterator

classes/local/event_processor.php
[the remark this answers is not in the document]
matthewhilton: Thought about this some more and I think I might leave it, my comment above might be a bit of over-optimisation. I'll probably leave as-is

classes/local/step/reader_event.php
matthewhilton: Smart idea although I worry that it reduces code readability for what could be a negligable performance gain.

I think for now we should keep it as is as it is more straightforward and in the future if people are running into bottlenecks here we can optimise it a bit more.

VARIABLES.md
```diff
@@ -0,0 +1,28 @@
...
+
+I **must** add `courseid: ${{ row.courseid }}` to the 'Variables' field (under extra settings) for it to show up in the variable tree. Then any other step can access it via either:
+
+`{{steps.course_sql_reader.vars.courseid}}`
+
+OR
+
+`{{steps.course_sql_reader.row.courseid}}`
+
+The latter option is only available when the variable if asked for by some step, since then the config 'knows' what the shape of the data will be.
```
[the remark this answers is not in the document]
matthewhilton: I have added a new example section with some pictures and more description of what I mean

classes/local/event_processor.php
```diff
@@ -0,0 +1,173 @@
...
+    public const EXECUTE_ADHOCQUEUED = 'adhocqueued';
+
+    /**
+     * Table for storing captured events.
+     * @var string
+     */
+    public const TABLE = 'tool_dataflows_events';
+
+    /**
+     * Records a triggered event.
```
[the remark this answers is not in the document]
matthewhilton: I updated some of the ones I found but will keep in mind for the future.

classes/local/step/reader_event.php
matthewhilton: Was able to replace it with a much simpler `$this->stepdef->get('dataflowid');` :+1:

db/upgrade.php
matthewhilton: yep that was a mistake - have fixed it :wrench:

classes/local/step/reader_event.php
matthewhilton: I just tested the dry run and the `get_iterator` is called to get the event data and it is passed to the `execute` function just as like any other run.

When I go to dry-run there is no form that pops up - is this a feature that steps can implement? I'm not too familiar with ti

matthewhilton:
I added a new method to the base_step:

https://github.com/catalyst/moodle-tool_dataflows/blob/04522d97f42155fb66873f1feba1e2ef625ea879/classes/local/step/base_step.php#L642-L651

And overrode it in the reader step

https://github.com/catalyst/moodle-tool_dataflows/blob/04522d97f42155fb66873f1feba1e2ef625ea879/classes/local/step/reader_step.php#L42-L50

And so instead of checking `group == 'reader'`, we can check instead the output of this function:

https://github.com/catalyst/moodle-tool_dataflows/blob/04522d97f42155fb66873f1feba1e2ef625ea879/classes/local/execution/iterators/dataflow_iterator.php#L133-L140

This is equivalent to the code before; readers will not pull the next. However, this now means the new `reader_event` class can be in the `trigger` group but still specify that it has a producing iterator (since it is a reader) meaning we don't encounter the bug #663 

Furthermore, I styled the trigger as blue to indicate to the user 'This is a trigger, but its also kinda a reader' since ultimately it is a reader with a bit of trigger'ry stuff sprinkled on top

image

I thought about going the other way and making it based on the `trigger` class with iterator stuff added, but because triggers are based off the `connector` class things get hairy pretty quickly. And since this step is probably 80% reader I though it would be best to keep it that way.

Would like some feedback on ^^ though.

classes/local/step/reader_event.php
keevan: There is no such feature currently. It would be nice, for the interface to display / prompt the user for an input when required (or stop until the input has been supplied).

A similar issue had been created here, but this is for the initial trigger (form trigger)
https://github.com/catalyst/moodle-tool_dataflows/issues/443

If / once that has been implemented, it could potentially be expanded to other steps and allow for gating.

For testing purposes, perhaps the data could be set up in the step itself? And would only be used for testing (e.g. dry runs) due to the absense of a form feature. 

Though, at the same time it might be good to make a generic field to hold dry-run test data that is used if certain conditions were checked. But I think this is somewhat scope creep, and should be implemented in another issue. 

I'm leaning towards adding a generically named field to hold the test (event) data for now, and building on that concept later down the track.

keevan (commented):
Everything looks mostly good. As discussed internally regarding the colour choices, I think keep it the same as the normal trigger step. It will ultimately head towards that direction after the future refactor.

I'll wait until you've added the rest of the things you plan to add such as those tests before taking another look.

Also I'm curious, have you tried testing if you can reference the event data using `${{ record.eventdata.courseid }}` based on your example in `VARIABLES.md`, as that should by default work.

One test case I'd like to cover as well, is one where on an event, you read some SQL data based on the data from the event and write it somewhere. It'd be interesting to inspect the shape of the $input that gets passed through.

classes/local/step/reader_event.php
matthewhilton: > I'm leaning towards adding a generically named field to hold the test (event) data for now, and building on that concept later down the track.

I think this should probably be implemented later since to me it seems a bit complex and not necessary at the moment:
- Different events have different structures, so would we store a test event for every event type? What about installed plugin events?
- If I wanted to dry-run for e.g. a `course_viewed` event, it would probably have to load the test data and then extra form field for e.g. what user, what course id

matthewhilton:
Was running into #639 causing the unit tests on 3.5 and 3.6 to fail. I implemented a check that if the function exists, use it, otherwise just queue it. 

Apart from just copying the entire logic of the function in, I think this is the best solution.

https://github.com/catalyst/moodle-tool_dataflows/blob/635d810bd01870dbd648bf61f0799d3d5b9dceba/classes/task/process_dataflow_ad_hoc.php#L57-L62

matthewhilton:
> One test case I'd like to cover as well, is one where on an event, you read some SQL data based on the data from the event and write it somewhere. It'd be interesting to inspect the shape of the $input that gets passed through.

This currently isn't allowed since the `sql_reader` step does not allow upstream flows. #671 would cover this functionality

classes/local/step/trigger_event.php
```diff
@@ -0,0 +1,281 @@
...
+/**
+ * Event reader step
+ *
+ * @package   tool_dataflows
+ * @author    Matthew Hilton <matthewhilton@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class trigger_event extends trigger_step {
+    /**
```
keevan, line 34: Could we add a whitespace before this phpdoc block.

classes/local/step/trigger_event.php
```diff
@@ -0,0 +1,281 @@
...
+
+        return new dataflow_iterator($this->enginestep, new \ArrayIterator($eventarray));
+    }
+
+    /**
+     * Get the default data.
+     *
+     * @param \stdClass $data
+     * @return stdClass
+     */
```
keevan, line 83:
```suggestion
     * @param \stdClass $data
     * @return \stdClass
     */
```
Types should be consistently declared.

classes/task/process_dataflow_ad_hoc.php
```diff
@@ -34,22 +33,32 @@ class process_dataflow_ad_hoc extends \core\task\adhoc_task {
...
-                            userdate($metadata->timestamp) . ", process ID: $metadata->processid.");
-                }
-            }
-        } catch (\Throwable $thrown) {
-            mtrace("Dataflow run failed for ID: $dataflowrecord->dataflowid, " . $thrown->getMessage());
+        // Only available 3.7 onwards.
+        if (function_exists('\core\task\manager::reschedule_or_queue_adhoc_task')) {
+            \core\task\manager::reschedule_or_queue_adhoc_task($task);
+        } else {
+            \core\task\manager::queue_adhoc_task($task);
```
keevan, line 61: Could we de-nest this else block?
```suggestion
            return;
        }

        \core\task\manager::queue_adhoc_task($task);
```

classes/task/process_dataflows.php
```diff
@@ -39,54 +41,54 @@ public function get_name() {
...
+        }
+    }
+
+    /**
+     * Executes the given dataflow.
+     *
+     * @param int $dataflowid
+     * @return void
+     */
+    public static function execute_dataflow(int $dataflowid): void {
```
keevan, line 71: Similar to above, `void` doesn't seem important here so should be removed.

keevan (commented):
Some more comments. Otherwise it looks like it's in a good shape. Once those are cleaned up I'll do another round of functional testing.

VARIABLES.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Caveats
+- Variables can only be written inside of a step's `execute` function. You can't 'set' variables before the step is running such as in the step definition.
+
+## Usage in a dataflow
+After exposing a variable inside of a step's `execute function`, it **will not** automatically display inside the step form UI or debugging output, as the data could vary at runtime.
+
+Because of the iterator setup, variables will only show up if requested further down the chain.
+
+For e.g, If I had a event reader step called `event_reader` that called `$variables->set('event', $data);`
```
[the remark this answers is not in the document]
matthewhilton: Ah I did change this from sql reader -> event reader but forgot to change it to trigger :+1:  will update

classes/local/event_processor.php
```diff
@@ -0,0 +1,185 @@
...
+     */
+    public static function get_events_for_dataflow(int $dataflowid): array {
+        global $DB;
+        return $DB->get_records(self::TABLE, ['dataflowid' => $dataflowid]);
+    }
+
+    /**
+     * Listens to Moodle events and processes the corresponding event triggers.
+     *
+     * @param \core\event\base $event
```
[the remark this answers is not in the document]
matthewhilton: If it was an exception that I would expect people to encounter yes, but I think here its something that likely only devs will encounter if they break something so probably not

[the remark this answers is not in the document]
matthewhilton: Actually I've changed my mind after looking more into other plugins it seems they specify coding exceptions are thrown. I guess its useful if you are a dev you can easily see whats possible without inspecting the function.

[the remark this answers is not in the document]
keevan: Yep :+1:, phpdocs are generally written with devs in mind, and if as part of typing out the function they see that it throws an exception, they might decide to handle it from the calling method if they want.

classes/local/step/reader_event.php
keevan: > Different events have different structures, so would we store a test event for every event type? What about installed plugin events?

By default it will be empty and that either passes through no data or fails. Users can define their own test data which would be based on what they expect it to look like.

> If I wanted to dry-run for e.g. a course_viewed event, it would probably have to load the test data and then extra form field for e.g. what user, what course id

Yes, and those things could come from expressions as well, or hard-coded based on what the tester wants.

db/upgrade.php
```diff
@@ -234,5 +234,28 @@ function xmldb_tool_dataflows_upgrade($oldversion) {
         upgrade_plugin_savepoint(true, 2022091600, 'tool', 'dataflows');
     }
 
+    if ($oldversion < 2022100601) {
```
keevan, line 237: This needs to match the one defined in version.php, and would be better if it was bumped by date instead of a micro bump. I had issues adding the table and noticed the zero was meant to be a 2. :')

keevan:
Functional testing seems OK. I encountered some issues.

Using the example workflow in this PR
- I had to change the type to `tool_dataflows\local\step\trigger_event` (from reader_event)
- I encountered this on the course page. I believe it to be outputting directly into the page which is not desired.
image
When using this setting 
image

db/upgrade.php
matthewhilton: ah I thought i fixed this... will fix it again :laughing:

matthewhilton:
I noticed something similar too, but I thought it was the result of the debugging writer. Maybe I might add output buffer catching and pass this to the debugging output instead when using this option.

matthewhilton:
Also some things discussed internally:

- [ ] Give warning when changing the event type that previous events are deleted
- [ ] Switch wording of policies:
```
Run immediately
Run asap in individual tasks in parallel
Run asap in batch tasks in series
```
- [ ] When a flow doesn't allow parallelism, method 2 cannot run, so it must use either 1 or 3 (probably 3, the queue method)

matthewhilton:
> I'm not super fussed on exactly what happens if the config changes as long as nonthing is dropped andit is clearly defined
> like I might guess that if you swap from single events in paralell to batch in series then its easier to let the old ones drain and just pick up the new ones
> so you might have 4 in parallel + another one in series running at the same time
> and then after they are all done we are back to just 1 in series

Events are processed via the execution policy set at the exact moment the event is recorded. 

- In the case of 'process immediately' this locks the session so it isn't possible to change the execution policy while these are running
- In the case of adhoc serial & adhoc queued, If previous events were recorded and ran in parallel, but then the config changes to serial queue, the previous

When adhoc tasks are created these are only recording the dataflow ID to run - no event data is included in the adhoc task itself, this is loaded from the DB when it is run based on the trigger's step id. And once it is loaded into a dataflow, the record is deleted. 

The trigger step itself does not really care at all how it is run, you could queue 500 events and 1000 adhoc tasks to process them and it wouldn't care, after all the events are processed the remaining tasks would just have empty iterators.

lib.php
```diff
@@ -76,6 +76,7 @@ function tool_dataflows_step_types() {
         new step\trigger_cron,
         new step\writer_debugging,
         new step\writer_stream,
+        new step\trigger_event
```
keevan, line 79: Could we please add a trailing comma here.

> Note the last item has a trailing comma left there which allows to extend the list of items later with a cleaner diff. For the same reason, it is better not to align the assignment operators. 
https://docs.moodle.org/dev/Coding_style#Wrapping_arrays

keevan (commented):
A few small comments. Going through functional testing now.

classes/local/event_processor.php
```diff
@@ -0,0 +1,217 @@
...
+        $timeout = 30;
+        $locktype = 'tool_dataflows_event_queue';
+        $resource = 'queue';
+        $lockfactory = \core\lock\lock_config::get_lock_factory($locktype);
+        if ($lock = $lockfactory->get_lock($resource, $timeout)) {
+            return $lock;
+        } else {
+            // We did not get access to the resource in time, give up.
+            throw new \moodle_exception('locktimeout');
+        }
```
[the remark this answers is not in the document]
matthewhilton: Agreed, this was just copied and pasted from the moodle example :stuck_out_tongue:

lib.php
matthewhilton: Ah interesting. I've had some weird behaviour with trailing commas before so I usually avoid them. Didn't realise they were in the moodle spec.

keevan (approved):
Looks good. Merging in for now and we'll tweak if we encounter any issues

PR #669, opened 2022-11-07
fix: authentication header not managed by filelib setHeader method

marcghaly:
Hi @keevan,

This PR solves the issue when passing an authentication token, it is currently not managed by the setHeader moodle lib.

Regards,

Marc


classes/local/step/connector_curl.php
```diff
@@ -288,6 +288,11 @@ public function execute($input = null) {
 
         // Provided a header is specified, add header to request.
         if (!empty($headers)) {
+            // Catch any authorization authentication passed through header and properly format it.
+            if (isset($headers['Authorization'])) {
+                array_unshift($headers, 'Authorization:' . $headers['Authorization']);
+                unset($headers['Authorization']);
+            }
```
keevan, line 295: Hi @marcghaly,

I've had a test of this locally. Whilst it works for Auth headers, it will certainly be broken for everything else.

The expected format of the headers, when passed into `$curl->setHeader`, seems to be `key: value`, which means for all headers, not just `Authorization`, should be converted into this format.

So the suggestion is to change this to behave similarly to this:
```
            foreach ($headers as $key => $value) {
                //  TODO: sanitise $key and $value here?
                $curlheaders[] = "$key: $value";
            }
            $curl->setHeader($curlheaders);
```
Going further, it might be a good idea to pull this out to another method in the step type (e.g. set_headers), and do this handling there.

keevan (changes requested):

classes/local/step/connector_curl.php
brendanheywood: I'm probably missing something, but the value in the field should be ready to go verbatim, and if it isn't then this should have been caught earlier at the form validation step.

For example I just opened up the dev tools testing this page and cut paste some headers out of it and I would expect these to Just Work ™

```
Accept-Ranges: none
Cache-Control: private, pre-check=0, post-check=0, max-age=0, no-transform
Connection: keep-alive
...
```

keevan: Yes, it should _just work_. 

Currently, the headers are being prepared like so:
```
// Extract headers.
$headers = helper::extract_http_headers($config->headers);
```
Which is causing it to return as an associative array (key value), which causes its own issues. (You can't declare multiple headers with the same key even if you wanted, and even if it was technically valid). 

Also, I'm not sure if there was ever a requirement to allow `#`-like comments in the original headers field, which might require some processing to remove it during actual usage.

If that's not the case, yes I believe $curl->setHeader accepts a string input and turns it into an array of header fields automatically, so that would simplify things further if used. (e.g. `$curl->setHeader($config->headers);`)

marcghaly:
Hi @keevan,

I have added a method to set proper format and added validation for the run as well.

Regards,

Marc

marcghaly:
ci err do not seem related

keevan (approved):
It seems mostly good now. 

I'll fix up the CI issues (phpdoc related) which are minor and get this merged.

Thanks for working on this @marcghaly

PR #670, opened 2022-11-09
Moodle 35 stable cronandcurl

marcghaly:
This PR solves an undefined error aborting cron run and an unfound curl class.

Regards,

Marc


keevan (approved):
Seems harmless enough :)

PR #672, opened 2022-11-14
Add sql flow step

### LINKED ISSUE #671: SQL flow step
As discussed internally, a step similar to `reader_sql` but does not have an iterator. It would return a single row and be used for data fetching from the DB. 

`reader_sql` cannot be used currently with other steps that have iterators since it requires the variables to be present at the time of iterator creation, instead of when `next()` is called.

**Example use case:**

1. Course viewed event triggered (userid, courseid)
2. SQL flow step then gets a customfield from a course e.g. `'SELECT * FROM {customfield_data} WHERE fieldid = 4 AND instanceid = ${{steps.moodle_event.vars.courseid}}'`
3. Everything else... 

To do:

- [ ] Move SQL functions into a trait that can shared by `flow_sql` and `reader_sql` (except the counter handling, since this only allows 1 row)
- [ ] Implement a `flow_sql` step
- [ ] Some handling around if 0 or > 1 rows returned.

### LINKED ISSUE #673: Dataflows can be aborted multiple times causing error with `stream_writer`
Scenario:

- A dataflow setup such as: `JSON Reader step -> Flow step -> Stream writer step`
- The flow step throws an exception, causing the flow to be aborted

This causes an error whereby the stream writer's stream is attempted to be closed twice, leading to an error on the second time (since its already closed)

**How it happens**

1. The exception is thrown inside a step, so the dataflow iterator catches this and aborts the engine
https://github.com/catalyst/moodle-tool_dataflows/blob/f7941e08d2c1871d438c8a9ea79939a1725f92bb/classes/local/execution/iterators/dataflow_iterator.php#L206-L210

2. This goes through each each enginestep and flowcap and calls their abort functions .
https://github.com/catalyst/moodle-tool_dataflows/blob/f7941e08d2c1871d438c8a9ea79939a1725f92bb/classes/local/execution/engine.php#L508-L530

3. As part of aborting the engine steps above, the stream writer iterator `on_stop` is called to close the stream
https://github.com/catalyst/moodle-tool_dataflows/blob/f7941e08d2c1871d438c8a9ea79939a1725f92bb/classes/local/step/writer_stream.php#L182-L190

4. Back in the iterator in step 1, the exception is re-thrown
https://github.com/catalyst/moodle-tool_dataflows/blob/f7941e08d2c1871d438c8a9ea79939a1725f92bb/classes/local/execution/iterators/dataflow_iterator.php#L206-L210

5. This is then caught by the engine flow cap, which then attempts to abort the engine again.
https://github.com/catalyst/moodle-tool_dataflows/blob/f7941e08d2c1871d438c8a9ea79939a1725f92bb/classes/local/execution/engine_flow_cap.php#L57-L60

6. This then goes back through each enginestep and flowcap iterators and aborts their execution, which throws an error with the stream since the stream is already closed.

**Possible fixes;**
1. Add check if stream is already closed / is close-able before closing it.
2. In when aborting the engine, check if the engine is already aborted. If it is, then we ignore.

I think option is good since the `on_close` IMO should be idempotent and not care how many times you call it.

matthewhilton:
## Summary:

1. New `flow_sql` step - this is for reading a single row using for e.g. variables. 
3. Refactored `reader_sql` with a new `sql_trait` - this is shared by the new `flow_sql` step. Note not all of the functions have been moved into the trait, for e.g. the row counting variable logic is redundant in the flow_sql step, so it only exists within the reader_sql step.
4. Fixed a tiny bug with the `stream_writer` that was causing the flow_sql unit tests to fail #673

## Functional testing
Create a small test course to generate users, then import the test dataflow  (rename to yml github doesn't allow straight yaml)
[Test_20221121_0211.txt](https://github.com/catalyst/moodle-tool_dataflows/files/10052034/Test_20221121_0211.txt)

This dataflow does the following:
1. Uses the sql reader to read enrolments and gets the $userid
2. Uses the new Flow SQL step to lookup a username from the $userid
3. Sends an email to test@test.com with the username gathered from the flow_sql call.
4. Check your Mailhog to confirm you get emails for each tool_generator user that was generated 

## Closes issues 

Closes #671 
Closes #673

### DIFF
--- a/classes/local/step/flow_sql.php
+++ b/classes/local/step/flow_sql.php
@@ -1,0 +1,107 @@
+<?php
+// This file is part of Moodle - https://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_dataflows\local\step;
+
+/**
+ * SQL flow step
+ *
+ * @package   tool_dataflows
+ * @author    Matthew Hilton <matthewhilton@catalyst-au.net>
+ * @copyright 2022, Catalyst IT
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class flow_sql extends flow_step {
+    use sql_trait;
+
+    /**
+     * Return the definition of the fields available in this form.
+     *
+     * @return array
+     */
+    public static function form_define_fields(): array {
+        return [
+            'sql' => ['type' => PARAM_TEXT, 'required' => true]
+        ];
+    }
+
+    /**
+     * Allows each step type to determine a list of optional/required form
+     * inputs for their configuration
+     *
+     * It's recommended you prefix the additional config related fields to avoid
+     * conflicts with any existing fields.
+     *
+     * @param \MoodleQuickForm $mform
+     */
+    public function form_add_custom_inputs(\MoodleQuickForm &$mform) {
+        // SQL example and inputs.
+        $sqlexample = "
+        SELECT id, username
+          FROM {user}
+      WHERE id > \${{steps.xyz.var.number}}
+      ORDER BY id ASC
+         LIMIT 10";
+        $sqlexamples = \html_writer::tag('pre', trim($sqlexample, " \t\r\0\x0B"));
+        $mform->addElement('textarea', 'config_sql', get_string('flow_sql:sql', 'tool_dataflows'),
+            ['cols' => 50, 'rows' => 7, 'style' => 'font: 87.5% monospace;']);
+        $mform->addElement('static', 'config_sql_help', '', get_string('flow_sql:sql_help', 'tool_dataflows', $sqlexamples));
+    }
+
+    /**
+     * Execute configured query
+     *
+     * @param   mixed $input
+     * @return  mixed
+     * @throws \dml_read_exception when the SQL is not valid.
+     */
+    public function execute($input = null) {
+        global $DB;
+
+        // Construct the query.
+        $variables = $this->get_variables();
+        $config = $variables->get('config');
+        $sql = $this->evaluate_expressions($config->sql);
+
+        // Execute the query using get_records instead of get_record.
+        // This is so we can expose the number of records returned which
+        // can then be used by the dataflow in for e.g. a switch statement.
+        $records = $DB->get_records_sql($sql);
+
+        $variables->set('count', count($records));
+
+        // Only return the query data if there is exactly 1 record.
+        // If multiple records are returned, it is undefined what should happen.
+        $invalidnum = ($records === false || count($records) !== 1);
+        $data = $invalidnum ? null : array_pop($records);
+        $variables->set('data', $data);
+    }
+
+    /**
+     * Validate the configuration settings.
+     *
+     * @param   object $config
+     * @return  true|\lang_string[] true if valid, an array of errors otherwise
+     */
+    public function validate_config($config) {
+        $errors = [];
+        if (empty($config->sql)) {
+            $errors['config_sql'] = get_string('config_field_missing', 'tool_dataflows', 'sql', true);
+        }
+
+        return empty($errors) ? true : $errors;
+    }
+}
--- a/tests/tool_dataflows_flow_sql_test.php
+++ b/tests/tool_dataflows_flow_sql_test.php
@@ -1,0 +1,277 @@
+<?php
+// This file is part of Moodle - https://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_dataflows;
+
+use Symfony\Component\Yaml\Yaml;
+use tool_dataflows\dataflow;
+use tool_dataflows\local\execution\engine;
+use tool_dataflows\step;
+use tool_dataflows\local\step\flow_sql;
+
+defined('MOODLE_INTERNAL') || die();
+
+// Include lib.php functions that aren't included automatically for Moodle 37- and below.
+require_once(dirname(__FILE__) . '/../lib.php');
+
+/**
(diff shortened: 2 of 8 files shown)


classes/local/step/flow_sql.php
```diff
@@ -0,0 +1,144 @@
...
+     */
+    public function form_add_custom_inputs(\MoodleQuickForm &$mform) {
+        // SQL example and inputs.
+        $sqlexample = "
+        SELECT id,
+               username
+          FROM {user}
+      WHERE id > \${{steps.xyz.var.number}}
+      ORDER BY id ASC
+         LIMIT 10";
```
keevan, line 69: Could we change this so that the indentation stays towards the RHS and has highlighting, this can be achieved by ensuring the start of the query is 
```
$sqlexample = "SELECT id,
...
```
instead of 
```
$sqlexample = "
SELECT id,
...
```
And see how that looks.

matthewhilton: I tried this but it makes the formatting when actually looking at the text in the static form element look weird:

Before: 
image

After:
image

I did tweak the `id, username` to be on a single line

classes/local/step/flow_sql.php
```diff
@@ -0,0 +1,143 @@
...
+        $variables = $this->get_variables();
+        $config = $variables->get('config');
+        $sql = $this->evaluate_expressions($config->sql);
+
+        // Execute the query using get_records instead of get_record.
+        // This ensures dataflow builders are not unaware when their query could
+        // potentially return multiple values for a given sql call.
+        $records = $DB->get_records_sql($sql);
+
+        if ($records === false || count($records) !== 1) {
```
keevan, line 105: A comment here I believe would be nicer than reading the conditions.

classes/local/step/flow_sql.php
```diff
@@ -0,0 +1,143 @@
...
+            $errormsg = get_string('flow_sql:error:invalidrowsreturned', 'tool_dataflows');
+
+            switch ($failureconfig) {
+                case self::ABORT_FLOW:
+                    throw new \moodle_exception($errormsg);
+                case self::SKIP_RECORD:
+                    $this->enginestep->log('Warn skipping record: ' . $errormsg);
+                    return false;
+                default:
+                    throw new \coding_exception("Invalid failure config in step.");
```
keevan, line 116: Could we update this to use single quotes here instead?

https://docs.moodle.org/dev/Coding_style#Single_quotes

tests/tool_dataflows_flow_sql_test.php
```diff
@@ -0,0 +1,333 @@
...
+            $engine = new engine($this->dataflow);
+            $engine->execute();
+        } catch (\Exception $e) {
+            $this->assertInstanceOf('moodle_exception', $e);
+        }
+        ob_get_clean();
+
+        $this->assertTrue($engine->is_aborted());
+    }
+
```
keevan, line 200: I'm not sure if the default behaviour for an empty SQL statement should abort the flow. 

It would potentially mean you wouldn't be able to do something like an `exists` check. Instead I think the behaviour should be the same as the SQL reader, in that if there are no results, an empty iterator is passed through (or false), which means it holds no record.

So for this flow step, instead of an object, it should hold `NULL` instead of aborting. This also means subsequent steps can handle this however they want.

If you want the abort behaviour, you can use the abort flow step and abort if the value set here is null.

tests/tool_dataflows_flow_sql_test.php
```diff
@@ -0,0 +1,333 @@
...
+        try {
+            $engine = new engine($this->dataflow);
+            $engine->execute();
+        } catch (\Exception $e) {
+            $this->assertInstanceOf('moodle_exception', $e);
+        }
+        ob_get_clean();
+
+        $this->assertTrue($engine->is_aborted());
+    }
```
keevan, line 225: Similar to the "abort if no results" test, I think it's better to only set the first record to the step, but output a warning in the logs to indicate more than one result was found but only the first was actually used. 

This would allow the user to tweak the query to `LIMIT 1`, and if they really want the flow to stop  if there are more than one result, we could add the count as another variable to the step. Something like `count`, or whatever you think describes it better here.

tests/tool_dataflows_flow_sql_test.php
keevan: I realised also that there are ABORT / SKIP options to be set in the flow_sql step. Is ABORT the default? It seems to be defaulted as part of the execute step, but I think this should be part of the validation (which means before you even run the dataflow, it should have stopped you if you didn't provide a value)

classes/local/step/flow_sql.php
keevan: Probably alluding to a similar point in other comments, I'm concerned that failing here would make this step too rigid.

You may want to check if something doesn't exist before creating it (which this won't let you do by default unless you change the failure type?), or making sure it exists and aborting it (in the next step).

To keep things simple, I think the only time it probably should fail is when there is a query execution error (bad syntax or connection issue), instead of one based on results returned.

keevan: Saying that, definitely still warn/log for potentially unintended resultsets, e.g. warn if more than one result, warn if no results, etc.

tests/tool_dataflows_flow_sql_test.php
matthewhilton: abort flow is the default, yes. I don't see how we would check this before running the dataflow? We could check this if the query doesn't use variables, but if it does, how would we input the variables to check if it returns != 1 rows ?

matthewhilton: If a user wanted to choose different paths based on if a record exists, I would think it would be better to have that as a seperate `sql_flow` where the query is if it exists.

tests/tool_dataflows_flow_sql_test.php
keevan: Yes, that seems like a reasonable use case. But is there anything preventing them adding that check in a subsequent step if it's that important to them? It feels weird to abort the whole dataflow, because of one potential query being different from the rest. I would in this case, push this into subsequent steps to deal with the different conditions.

Also, would it easier to use the `$DB->get_record` method and strictness flags instead or doing a custom implementation? (IGNORE_MISSING, IGNORE_MULTIPLE, MUST_EXIST)

matthewhilton:
Had some good discussion with @keevan who made a good point about in the future having error handling e.g. abort flow, skip record as a trait rather than each step handling itself.

With this in mind my thinking is the following:

1. Remove the config for the failure method (skip or abort)

2. Then in the following situations have the behavior be:
- 0 records returned - Pass through data as null, count = 0
- 1 record returned - Pass through data, count = 1
- greater than 1 record returned - Pass through data as null but log a warning that multiple records were returned. We don't know what record to use, so we should assume that no records are removed. count = num rows returned

3. Expose a count variable from the step as well for the rows returned. This makes it easy for a switch after it to simply check count = 1 = good, default: error. You could also the have the ability to for e.g.:

This means its easier to do queries that expect either 1 or 0 records (but not > 1) - for e.g. querying a user enrolment in a course
- count = 0  = user not enrolled in course
- count = 1 = user enrolled in course
- default: error - something isn't right

matthewhilton:
Also should note I decided against logging when 0 or > 1 records are returned, since with the setup now these might actually be things people are intending to happen.

keevan:
> Also should note I decided against logging a warning when 0 or > 1 records are returned, since with the setup now these might actually be things people are intending to happen.

I think it should still warn if there are multiple results. In most cases, as you've pointed out, this is probably not what the admin wants. If they wanted to confirm counts, they should probably use a count query instead. 

Thoughts?

matthewhilton:
I would think that if someone set up a dataflow that they wanted to check if returned > 1 row and it kept outputting a debugging message, it would be annoying (and not useful). Since we kind of allow it now, I think we shouldn't warn. Otherwise we would need to add a config setting for essentially 'I am not expecting multiple rows' which seems a bit silly.

keevan (approved):

PR #684, opened 2023-01-09
Issue 683: Fix clear_dependent() to correctly remove dependents.

### LINKED ISSUE #683: PHP8.0 - deprecated
Deprecated: array_udiff(): Returning bool from comparison function is deprecated, return an integer less than, equal to, or greater than zero in /var/www/html/admin/tool/dataflows/classes/local/variables/var_value.php on line 172

jaypha:
Resolves #683

### DIFF
--- a/classes/local/variables/var_value.php
+++ b/classes/local/variables/var_value.php
@@ -164,13 +164,10 @@ class var_value extends var_node {
      * @param var_value $var
      */
     protected function clear_dependent(var_value $var) {
-        $this->dependents = array_udiff(
-            $this->dependents,
-            [$var],
-            function ($a, $b) {
-                return $a === $b;
-            }
-        );
+        $idx = array_search($var, $this->dependents, true);
+        if ($idx !== false) {
+            unset($this->dependents[$idx]);
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


classes/local/variables/var_value.php
```diff
@@ -164,13 +164,12 @@ public function evaluate(?callable $errorhandler = null) {
...
-            [$var],
-            function ($a, $b) {
-                return $a === $b;
+        foreach ($this->dependents as $idx => $dep) {
+            if ($dep === $var) {
+                unset($this->dependents[$idx]);
+                return;
             }
-        );
+        }
```
keevan, line 172: This change does make it easier to understand what's happening, personally I was not really familiar with array_udiff. 

Instead of having a loop, would using `array_search` make more sense here? For example:
```
$idx = array_search($var, $this->dependents, true);
unset($this->dependents[$idx]);
```

keevan (commented):
One small comment, the rest looks okay to me.

classes/local/variables/var_value.php
jaypha: array_udiff does not appear to work if the array elements do not have a defined order. If I use 0 for same, 1 for not same, it will fail.

jaypha: Thanks for the tip. I've been working with PHP for almost 20 years now, and there is still so much I don't know.

jaypha: Done

keevan (approved):

Issue #690, opened 2023-03-09
lasttimestarted is not accessible from UI 


marcghaly:
This seems to not be evaluated 
![image](https://user-images.githubusercontent.com/76186988/224165698-2e4923b2-2512-465b-be6a-f2fd6c815cf1.png)
and even if it were, this would be each time I assume, user cannot use last run time in various steps.

PR #691, opened 2023-03-09
Issue #690: Accessible last started time from UI

marcghaly:
Hi @keevan,

This is a PR in relation to WR399328 I will add specific dataflow to test there, thank you.

image

Regards,

Marc

### DIFF
--- a/classes/dataflow.php
+++ b/classes/dataflow.php
@@ -55,6 +55,7 @@ class dataflow extends persistent {
             'name' => ['type' => PARAM_TEXT],
             'enabled' => ['type' => PARAM_BOOL, 'default' => false],
             'concurrencyenabled' => ['type' => PARAM_BOOL, 'default' => false],
+            'timelaststarted' => ['type' => PARAM_INT, 'default' => 0],
             'vars' => ['type' => PARAM_TEXT, 'default' => ''],
             'timecreated' => ['type' => PARAM_INT, 'default' => 0],
             'userid' => ['type' => PARAM_INT, 'default' => 0],
--- a/classes/local/variables/var_dataflow.php
+++ b/classes/local/variables/var_dataflow.php
@@ -34,6 +34,9 @@ class var_dataflow extends var_object_visible {
     /** @var dataflow The dataflow persistent object. */
     protected $dataflow;
 
+    /** @var timelaststarted Last time this dataflow has been started. */
+    protected $timelaststarted;
+
     /**
      * Construct the object.
      *
@@ -43,6 +46,7 @@ class var_dataflow extends var_object_visible {
     public function __construct(dataflow $dataflow, var_root $root) {
         parent::__construct('dataflow', $root, $root);
         $this->dataflow = $dataflow;
+        $this->get_timelaststarted();
     }
 
     /**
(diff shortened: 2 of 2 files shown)


classes/local/variables/var_dataflow.php
```diff
@@ -52,14 +56,26 @@ public function init() {
...
 
         foreach (engine::STATUS_LABELS as $state) {
             $this->set("states.$state", null);
         }
     }
+
+    /**
+     * Gets last time this dataflow started.
+     */
+    public function get_lasttimestarted() {
```
brendanheywood, line 73: An open ended question: pros and cons of starting this date directly in the flow record vs min / max in the runs table? So each time we kick off a run we just touch the timestamp and its just there already in memory without an extra sql

classes/dataflow.php
```diff
@@ -55,6 +55,7 @@ protected static function define_properties(): array {
             'name' => ['type' => PARAM_TEXT],
             'enabled' => ['type' => PARAM_BOOL, 'default' => false],
             'concurrencyenabled' => ['type' => PARAM_BOOL, 'default' => false],
+            'lasttimestarted' => ['type' => PARAM_INT, 'default' => 0],
```
brendanheywood, line 58: wondering if we should add lasttimefinished here as well, depending on the flow logic this might be a better date to use

keevan: - Could we rename this field to timelaststarted (keeping `time` at the start of the field for consistency with other time fields).

- In terms of adding a `timelastfinished` field in addition to this one. I think it can be added later if needed. The current requirements appear to work with a `record.date_created` value, which is likely used to prevent duplicate records from being processed. Comparing it to when the dataflow finished, might potentially cause records created between the time the dataflow starts, and when it finishes, to be skipped. Inversely, if there is a record created between the time it takes for dataflows to start and the records (json) to be fetched, then that might be processed twice, but that's much less likely to happen.

classes/local/variables/var_dataflow.php
keevan: By flow record, I understand that as being the dataflow configuration at that point in time.

Going with setting this value in the dataflow config/variables directly
**Pros:**
- allows you to see the value it checks before running it.
- avoids the extra (single) SQL lookup when a dataflow is constructed, since you are already reading the dataflow config at some point and it will already be in memory
- it is a good place to hold variables that might not always be used by all dataflows

**Cons:**
- Something will need to set it. If this happens automatically, it might not be intuitive. i.e. it might make more sense to have an additional step which sets this value to the configuration.
- dataflows which are self updating (updates its own configuration / state / variables) should generally be avoided. 

Regarding the query itself, max/min timestarted could also be `SELECT timestarted ... ORDER BY id desc LIMIT 1` so it could be slightly optimised that way. 

Going with making this value available magically via a dataflow field (e.g. `${{ dataflow.timelaststarted }} `).
**Pros**:
- Adding a field like this means similar usage in different dataflows can be simplified since it's a prebaked field.
- Because of the above, simpler dataflows, and behaviour/output is pretty clear in the current state.

**Cons**
- **Maintenance overhead** - Okay for one field, but what if a different field is required (Nth last, first, max, min, or one for a specific step). Adding fields this way can make maintenance a pain since we are now coding and maintaining logic for a field for a specific use case. As such, the setting a value directly as per Brendan's suggestion simplifies this by allowing the workflow author to specify the value they want and where they want it.
- It's not as robust, i.e. it might be more robust to add a way to reference a previous dataflow and access the values it held directly (e.g. `${{ dataflow.previous().timestarted }}`
- An extra SQL call when the dataflow is constructed, for every dataflow constructed.

Perhaps it might be ideal to have something like the following:
- capturing the request time of the step which fetches records.json,
- it would set this value to a field called `timerecordsfetched` in the request step
- depending on how much validation we want, we could choose to set and persist this value at the end of the dataflow, or in the same/adjacent step. This would indicate whether we don't are okay with skipping records.

The assumption here is the last processed record could not be stored for some reason, or perhaps it shouldn't be stored. And/Or if the request for records.json passes a timestamp parameter which is used to only fetch records after a certain timestamp (e.g. a remote delta)

The suggested solution is pretty generic and could be applied similarly for any other fields that might need some sort of persistence across runs. The concern is the fact it will update the current config state, but that is stored historically currently (just not queryable / available-via-expressions)

keevan:
Also, since the tracker which introduced CS changes has been reverted, I think it's safe to say that the CS fix comment should also be removed from this patch to keep it more focused on resolving the linked issue.

Ref: https://github.com/catalyst/moodle-tool_dataflows/issues/694#event-8750596804

keevan:
Some rough thoughts, I'm not sure how the latest changes would allow a user to set an arbitrary value (e.g. timestamp) which can be referenced across and within dataflow run(s). I'd lean away from defining anything specific which would be changed later.

marcghaly:
Hi @keevan, 

I am sorry I am not sure what the goal is, do you want this config to be added to global_vars input manually and not use this code ? If so, how should one proceed as it seems that those values are constants throughout the dataflow when we need something that is dynamically retrieved. Or do you want a step that solely does this eg: retrieving and setting the dataflow last run time ?

PR #695, opened 2023-03-13
Issue #693: Fix thirdpartylibs.xml to prevent vendor from being checked by CI.

### LINKED ISSUE #693: CI codesniffer sniffs third party files
The code sniffer tries to sniff files that are in the `vendor` directory. These are third party files and cannot be corrected.

jaypha:
Resolves #693

### DIFF
--- a/thirdpartylibs.xml
+++ b/thirdpartylibs.xml
@@ -77,4 +77,12 @@
     <license>MIT</license>
     <licenseversion></licenseversion>
   </library>
+  <library>
+    <location>vendor/composer</location>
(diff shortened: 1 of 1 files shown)


danmarsden:
this looks wrong to me.

The vendor folder is unlikely to always contain "all MIT" licensend content.

It's much better to explicitly define each library included in the vendor folder than defining a "catch-all" style rule like that.

brendanheywood:
@jaypha is this just the leftover bit specifically autoload.php ? You can add specific files to the xml not just a folder

PR #697, opened 2023-03-24
Add GPG encrypt/decrypt step

jaypha:


keevan:
Before I drill into the details. Could we make this a trait + flow and connector step. For the current client, we'd likely want it as a flow step (unless this supports multiple files at once?)

classes/local/step/connector_gpg.php
```diff
@@ -0,0 +1,166 @@
...
+namespace tool_dataflows\local\step;
+
+use tool_dataflows\helper;
+
+/**
+ * Encrypt/Decrypt with GPG.
+ *
+ * @package    tool_dataflows
+ * @author
+ * @copyright
```
keevan, line 26: Missing some data :)

classes/local/step/gpg_trait.php
```diff
@@ -0,0 +1,160 @@
...
+            return true;
+        }
+
+        $output = [];
+        $result = null;
+        $this->enginestep->log("Executing '$executable'");
+        exec($executable, $output, $result);
+        $this->enginestep->log($result === 0 ? 'Success' : 'Fail');
+
+        return ($result === 0);
```
keevan, line 109: I don't think this is used in a flow currently, but would it be an issue for the flow_gpg which I believe is expecting a php object to be returned (it might expect us to pass $input back, or false instead)

keevan (approved):
Looks mostly good to me. 

I noticed some double up of strings, and I think that's easy enough to change if needed later on.

One question but otherwise happy for it to be merged.

keevan (approved):

PR #698, opened 2023-04-04
Refactor copy file step to be able to append.

jaypha:


keevan:
I don't think it will be intuitive if the concat behaviour was added to the copy step. 

Would it make more sense to have its own discrete step? (e.g. concat / join / combine step)

jaypha:
It would be effectively the same thing with one line different.

PR #699, opened 2023-04-05
Add append step

jaypha:


keevan (commented):
Hi @jaypha. This is all looking good. Can we add some unit tests to confirm the expected behaviour.

keevan (commented):
Thanks Jason, appreciate the detailed unit tests. 

Could we add more comments to the setUp function as it's a bit hard to read and quickly understanding what's happening. 

In terms of the tests themselves, I wonder if we could simplify it to test only the step itself, and skip all the dataflow bootstrapping which should simplify tests going forward.

An example would be this: https://github.com/catalyst/moodle-tool_dataflows/blob/9f0b1ed/tests/tool_dataflows_connector_directory_file_count_test.php

This can be done later for all the current tests and classes as a bigger clean up exercise, but I wanted to hear your initial thoughts.

jaypha:
I'm not so sure. By creating actual dataflows, we test the class in the way it gets used. The simpler testing may work well for simple classes, but more complex classes could have subtle issues that the simpler testing wouldn't catch.
Setting up a dataflow isn't very complicated. It's mostly cut and paste.

keevan:
> I'm not so sure. By creating actual dataflows, we test the class in the way it gets used. The simpler testing may work well for simple classes, but more complex classes could have subtle issues that the simpler testing wouldn't catch. Setting up a dataflow isn't very complicated. It's mostly cut and paste.

Yes that is probably true. Ideally we'd have simple classes that are easy to reason about, do their main thing, and the tests should reflect that. Any issues with wiring of the step (for non logic steps like switches) that we find in an action step like copying a file or writing to a file should be isolated. 

That said I'm okay with these changes here. I think if it does get cleaned up, it will be cleaned up for all the steps in one go (the core handling would be mostly the same)

keevan (approved):

PR #702, opened 2023-04-17
Issue 142: Fix helper::path_is_relative() and helper::path_validate() to expand [dataroot] placeholder before testing.

### LINKED ISSUE #701: Helper methods fail to expand [dataroot] before performing checks.
Both helper::path_validate() and helper::path_is_relative() fail to expand [dataroot] before performing their checks.

jaypha:
Resolves #701

### DIFF
--- a/tests/tool_dataflows_helper_test.php
+++ b/tests/tool_dataflows_helper_test.php
@@ -35,6 +35,35 @@ class tool_dataflows_helper_test extends \advanced_testcase {
     }
 
     /**
+     * Tests path_is_relative().
+     *
+     * @dataProvider is_relative_provider
+     * @covers \tool_dataflows\helper::path_is_relative
+     * @param string $path
+     * @param bool $expected
+     */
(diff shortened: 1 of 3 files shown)


tests/tool_dataflows_helper_test.php
```diff
@@ -34,6 +34,35 @@ protected function setUp(): void {
...
+     */
+    public function is_relative_provider(): array {
+        global $CFG;
+        return [
+            ['one/path', true],
+            ['/one/path', false],
+            ['file://one/path', false],
+            ['sftp://one/path', false],
+            ['sftp/files/download.txt', true],
+            ['[dataroot]/stuff/one/path', strpos($CFG->dataroot, '/') !== 0],
```
brendanheywood, line 62: it is not ideal to have logic inside a data provider, shouldn't this one always return false as it will never be relative?

jaypha: fixed

PR #708, opened 2023-04-20
Add directory file list connector and reader steps.

jaypha:


classes/local/step/directory_file_list_trait.php
```diff
@@ -0,0 +1,130 @@
...
+                $sort = ' --sort=time';
+                break;
+            case 'time_reverse':
+                $sort = ' --sort=time -r';
+                break;
+        }
+
+        $path = $this->get_engine()->resolve_path($config->directory) . '/' . $pattern;
+
+        $command = 'ls -d1p --almost-all' . $sort . ' '. $path;
```
keevan, line 112: This section could be improved with some comments. In particular for options which aren't obvious to people who might not use the options extensively for `ls`

Some references which could help
- `man ls` (from the cli)
- or this site might help with some descriptions you can pull from https://explainshell.com/explain?cmd=ls%20-d1p%20--almost-all

classes/local/step/directory_file_list_trait.php
```diff
@@ -0,0 +1,130 @@
...
+        exec($command, $output, $result);
+
+        $variables->set('vars.success', $result === 0);
+
+        if (!isset($config->subdirectories)) {
+            $output = array_filter($output, function ($path) {
+                return substr($path, -1) !== '/';
+            });
+        }
+        $output = array_slice($output, $offset, $limit);
```
keevan, line 126: I also read through the offset variable usage (since I didn't know what it was doing). 
I like how you've put this together. Which makes it pretty flexible in how it can return the relevant values.

classes/local/step/directory_file_list_trait.php
```diff
@@ -0,0 +1,130 @@
...
+
+        $output = [];
+        $result = null;
+        $this->enginestep->log("Executing '$command'");
+        exec($command, $output, $result);
+
+        $variables->set('vars.success', $result === 0);
+
+        if (!isset($config->subdirectories)) {
+            $output = array_filter($output, function ($path) {
```
keevan, line 122: A comment above the array_filter to summarise what it's doing would be good here. 

Bonus points for a _before_ and _after_ example, but perhaps that's something that could be shown in the unit test instead?

classes/local/step/directory_file_list_trait.php
keevan, line 112: Security wise, the `$path` variable needs to be sanitised since that's the only place where user data can be added without restrictions.

jaypha:
- I want to add unit tests, but with time constraints, it may need to wait until after when we have some breathing space.

- As for 'ls'. I wanted to use it because it had sorting built in. But that sorting is primitive and limited. By using opendir or DirectoryIterator, we get more powerful options at the expense of having to provide the sorting ourselves.

classes/local/step/directory_file_list_trait.php
jaypha: I added some comments

classes/local/step/directory_file_list_trait.php
jaypha: I have sanitised the $path, but $pattern remains a problem. I was unable to find a way to do it. It looks like we will need to switch to code based listing.

jaypha: I added some comments

keevan (approved):
Looks good to me.

PR #713, opened 2023-04-24
Add filter step.

jaypha:


jaypha:
[Filter_demo_20230424_0535.yml.txt](https://github.com/catalyst/moodle-tool_dataflows/files/11307035/Filter_demo_20230424_0535.yml.txt)

lang/en/tool_dataflows.php
```diff
@@ -458,6 +458,10 @@
 $string['flow_logic_switch:casenotfound'] = 'The output position of #{$a} did not match any existing case';
 $string['flow_logic_switch:nomatchingcases'] = 'No matching cases, skipping record';
 
+// Flow transformer filter.
+$string['flow_transformer_filter:filter'] = 'Filter expression';
+$string['flow_transformer_filter:filter_help'] = "Put in a test expression. Since it is an expression, there is no need for '\${{' and '}}'.";
```
keevan, line 463: My lizard brain needs to do a lookup to remember that a filter will keep the things that evaluate as "true". Having the description mention this would be very helpful I think.

e.g.
```
Put in a test expression. Since ... there is no need for... . Truthy expressions will keep the previous records, and will filter everything else out.
```
You may word it better. Thoughts?

brendanheywood, line 463: It should be clear here that the expression should evaluate to a boolean true in order for a row to be kept

lang/en/tool_dataflows.php
```diff
@@ -458,6 +458,10 @@
 $string['flow_logic_switch:casenotfound'] = 'The output position of #{$a} did not match any existing case';
 $string['flow_logic_switch:nomatchingcases'] = 'No matching cases, skipping record';
 
+// Flow transformer filter.
+$string['flow_transformer_filter:filter'] = 'Filter expression';
+$string['flow_transformer_filter:filter_help'] = "This expression needs to evaluate to a boolean. Whenever the expression evaluaes to false, the current value for the flow will not be passed on to later steps in the flow. There is no need for '\${{' and '}}'.";
```
keevan, line 463:
```suggestion
$string['flow_transformer_filter:filter_help'] = "This expression must evaluate to a boolean. When the expression evaluates to false, the current value for the flow will not be passed on to later steps in the flow. There is no need for '\${{' and '}}'.";
```

keevan (approved):
Aside from the small lang changes, this one is good to go.

Issue #716, opened 2023-04-26
There is no way to alter data being passed through a flow step


jaypha:
Currently, a step should be able to alter the data being passed down a flow iteration by returning a new value.
This value, however, is not set anywhere and is ultimately thrown away.

keevan:
My understand of its current state is:
- flows will return the $input through (perhaps modified), and if the input should not flow through, it will pass a non-truthy value (e.g. false) instead
- connectors will pass true or false, but will only check on false explicitly to stop it from continuing.
- there are some steps which have been updated to have execute be the main entrypoint, but another "run" function is used to process the inputs and configuration provided as if it was called directly (like in more functional programming)

As such, I do believe the $input (or record) can be modified and should look modified as it is passed through to future steps. If this is not happening this is definitely a bug.

jaypha:
Fixed by #718

PR #718, opened 2023-04-26
Issue #716: Fix so that values can be passed through.

### LINKED ISSUE #715: Pull the boilerplate from the dataflow_iterator::next function into a separate function.
https://github.com/catalyst/moodle-tool_dataflows/blob/e0e5de88b91e41cbbadb7a8d5f22b706aae49d66/classes/local/execution/iterators/dataflow_iterator.php#L152-L180
Lines 152-180 can be extracted.

jaypha:
Resolves #715 #716

Remove boilerplate at beginning of dataflow_iterator::next(). Change iterator::next() signature to not return anything. Change dataflow_iterator::next() to set the current value to the return value of step::execute(). Fix steps to always return a value.
Fix flow_transformer_filter step
Add flow_transformer_filter unit test
Fix unit tests.

### DIFF
--- a/NEW_STEP.md
+++ b/NEW_STEP.md
@@ -23,6 +23,7 @@ Several convenience methods have been added to make this process a bit easier. Y
 - `$this->get_config()` which returns the configuration of the step, as an **object**, with all the expressions already evaluated.
 - `$this->is_dry_run()` which returns whether the dataflow was executed under dry run mode.
 - `$this->log()` which allows you to log a message during execution which if run directly would print the message to the user.
+-  $this->get_variables() which accesses the variables associated with this step.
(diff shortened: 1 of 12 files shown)


keevan (commented):
Hi Jason, thanks for making this change. This is great.

Would this correctly clone the php object being passed down, such that if we ever had a splitting flow, it would correctly retain a previous state of the object when it forks to a different step? It might not be immediately relevant now, but I recall this was discussed a bit a while back. 

There also seems to be some changes adding or removing the return $input value, adding `return false;` or only using `return;`. Once back, could you also update the NEW_STEP documentation to better explain the expected code flow going forward?

keevan (approved):
Looks good to me.

PR #721, opened 2023-04-27
Add the flow_transformer_alter step.

jaypha:


jaypha:
[Alter_demo_20230427_0122.yml.txt](https://github.com/catalyst/moodle-tool_dataflows/files/11338655/Alter_demo_20230427_0122.yml.txt)

classes/local/step/flow_transformer_alter.php
```diff
@@ -0,0 +1,93 @@
...
+     * Apply the filter based on configuration
+     *
+     * @param  mixed $input
+     * @return mixed The new value to be passed on the the next step.
+     */
+    public function execute($input = null) {
+        $variables = $this->get_variables();
+        $expressions = (array) $this->stepdef->config->expressions;
+        foreach ($expressions as $header => $expr) {
+            $input->$header = $variables->evaluate($expr);
```
keevan, line 89: Better if we use `$input->$field` here to match the form placeholder above.

lang/en/tool_dataflows.php
```diff
@@ -458,6 +459,10 @@
 $string['flow_logic_switch:casenotfound'] = 'The output position of #{$a} did not match any existing case';
 $string['flow_logic_switch:nomatchingcases'] = 'No matching cases, skipping record';
 
+// Flow transformer alteration.
+$string['flow_transformer_alter:expressions'] = 'Transform expressions';
+$string['flow_transformer_alter:expressions_help'] = 'Each line represents a transformation. The expression will be evaluated and the result assigned to the field identified by the label. The expressions will always use the original input values.';
```
keevan, line 464: It would be better to add an example, which could be the same as the placeholder? In particular when an admin has previously set a value already, and an admin wants to add another one and they have no real examples to draw from.

Potentially could be a follow up issue to improve on.

jaypha: That's why I think we need to create a github wiki. There is too much docs to keep putting it into the site.

tests/tool_dataflows_flow_transformer_alter_test.php
```diff
@@ -0,0 +1,177 @@
...
+
+    /** To be expected. */
+    const EXPECTED1 = [
+        [ 'a' => 'o', 'b' => '1', 'c' => '1 1', 'd' => 2 ],
+        [ 'a' => 'o', 'b' => '9', 'c' => '5 9', 'd' => 14 ],
+        [ 'a' => 'o', 'b' => '3', 'c' => '8 3', 'd' => 11 ],
+        [ 'a' => 'o', 'b' => '2', 'c' => '0 2', 'd' => 2 ],
+        [ 'a' => 'o', 'b' => '1', 'c' => '0 1', 'd' => 1 ],
+        [ 'a' => 'o', 'b' => '6', 'c' => '0 6', 'd' => 6 ],
+    ];
```
keevan, line 58: Should we move this into the test provider function? This seems to only be used once, and I don't think it would be used by anything outside of this class.

Also, was there a reason for the `1` suffix?

keevan (approved):
Thanks Jason, 

Everything looks good, and I have some very minor comments.

tests/tool_dataflows_flow_transformer_alter_test.php
jaypha: I don't want to move it inside the provider function as it would clutter it up too much. It has the 1 suffix to have the option to add more at a later date (which is also why I use a provider function).

Issue #377, opened 2022-07-29
In the step choose break connectors with side effects vs not into two groups


brendanheywood:
This also means we have 6 groups so it will stack nicer at different screen sizes

brendanheywood:
We also need to come up with a nice simple name for connectors that have actions, and unless noone has a better name I think we should go with 'Actions'

keevan:
Can you please elaborate on what you mean by connectors that have actions? Do you mean effects or is actions something else you have in mind?

brendanheywood:
I am proposing:

connector step + side effect === 'actions'

connector step - side effect === 'connectors'

keevan:
I feel like actions might be too generic. Other options I've considered:
- mutator
- change (step)
- effect 
- dirty (derived from the antonym being a pure step / function)

There's a big overlap with flow steps as well, in that both flow steps and connectors can have side effects, so it would be good if the convention chosen here works also for flow steps, though they don't _need_ to necessarily.

brendanheywood:
I'd also reconsider renaming readers and writers as they won't accurately describe some things that will end up in there, and I'm not sure they warrant the creation of yet another group. Things like the ws flow step, curl flow step and others. 

So how about we have some slightly more verbose but hopefully more obvious names:
* triggers
* connectors
* connector actions
* flow data sources (was readers, on the fence around this one)
* flow steps (replaces flow logic and will be anything which flows but has no side effects, like grep, filter, tee, head, tail etc)
* flow actions (ws, curl, email)

keevan:
Interesting. 

"Step" is a bit redundant since they are all steps technically (trigger step, connector step, etc), so perhaps:

- Triggers
- Connectors
- Connector Actions

- Sources / Resources / Data Source
- Flows
- Flow Actions

I'm also thinking of possibly overlapping connector actions and flow actions, e.g. we have an s3 copy connector, but we could have an s3 copy flow as well (e.g. reads DB record for file name and does the copy), vs a one off copy that happens in the connector step. 

The only difference (should they both exist) I see between the 2 is how they are configured. Logically they function more or less the same, e.g. run the step based on some configuration which could come from an expression. Connectors return true/false, flows pass the $input through, abort or stop the flow. Why shouldn't you be able to run a web service call as a once off connector step, potentially using data received from a webhook / event trigger?

It's late and I might have forgotten but how about this further suggestion:

- **F**lows (flow logic, filters, etc)
[...]

PR #724, opened 2023-04-28
Issue 707: Strip file list when appending files.

jaypha:
Will strip the first line when appending, but not if the file is empty.


keevan (approved):

keevan:
After some thought, this isn't as ideal as I'd hoped. The functionality added here should be part of the head/tail step loosely mentioned here.

https://github.com/catalyst/moodle-tool_dataflows/issues/377#issuecomment-1199244481

New issue created: https://github.com/catalyst/moodle-tool_dataflows/issues/725

PR #730, opened 2023-05-02
Refactor SQL execution to use prepared statements

Peterburnett:
SQL could be broken with values from a variable like `'`. Each variable is now processed as a parameter, and the whole SQL processed as a prepared statement

### DIFF
--- a/classes/local/step/flow_sql.php
+++ b/classes/local/step/flow_sql.php
@@ -73,13 +73,13 @@ class flow_sql extends flow_step {
 
         // Construct the query.
         $variables = $this->get_variables();
-        $config = $variables->get('config');
-        $sql = $this->evaluate_expressions($config->sql);
+        $config = $variables->get_raw('config');
+        [$sql, $params] = $this->evaluate_expressions($config->sql);
 
         // Execute the query using get_records instead of get_record.
         // This is so we can expose the number of records returned which
         // can then be used by the dataflow in for e.g. a switch statement.
-        $records = $DB->get_records_sql($sql);
+        $records = $DB->get_records_sql($sql, $params);
 
         $variables->set('count', count($records));
 
--- a/classes/local/step/reader_sql.php
+++ b/classes/local/step/reader_sql.php
@@ -76,19 +76,20 @@ class reader_sql extends reader_step {
      * @throws \moodle_exception
      */
     public function get_iterator(): iterator {
-        $query = $this->construct_query();
+        [$query, $params] = $this->construct_query();
 
-        return new class($this->enginestep, $query) extends dataflow_iterator {
+        return new class($this->enginestep, $query, $params) extends dataflow_iterator {
 
             /**
              * Create an instance of this class.
              *
              * @param  flow_engine_step $step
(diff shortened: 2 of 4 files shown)


classes/local/step/flow_sql.php
```diff
@@ -73,13 +73,13 @@ public function execute($input = null) {
 
         // Construct the query.
         $variables = $this->get_variables();
-        $config = $variables->get('config');
-        $sql = $this->evaluate_expressions($config->sql);
+        $config = $variables->get_raw('config');
+        list($sql, $params) = $this->evaluate_expressions($config->sql);
```
keevan, line 77:
```suggestion
        [$sql, $params] = $this->evaluate_expressions($config->sql);
```
An alternative which is shorter, and used more in modern languages (e.g. JS)

classes/local/step/reader_sql.php
```diff
@@ -76,19 +76,20 @@ public function is_concurrency_supported() {
      * @throws \moodle_exception
      */
     public function get_iterator(): iterator {
-        $query = $this->construct_query();
+        list($query, $params) = $this->construct_query();
```
keevan, line 79: Same as above
```suggestion
        [$query, $params] = $this->construct_query();
```

classes/local/step/flow_sql.php
Peterburnett: Ah yeah I forgot the deconstruction got cleaner a while back. neat

classes/local/step/reader_sql.php
```diff
@@ -183,9 +184,9 @@ function ($message, $e = null) use ($rawsql, &$errormsg) {
         }
 
         // Evalulate any remaining expressions as per normal.
-        $finalsql = $this->evaluate_expressions($finalsql);
+        list($finalsql, $params) = $this->evaluate_expressions($finalsql);
 
-        return $finalsql;
+        return [$finalsql, $params];
```
keevan, line 189: This doesn't look necessary to me. I might be missing something, are you able to clarify?

Peterburnett: its very not necessary. Didnt spot it at the time, as I was just following blind refactoring the touch points at the time

keevan (approved):
Thanks Peter. Looks good.

classes/local/step/flow_sql.php
```diff
@@ -73,13 +73,13 @@ public function execute($input = null) {
 
         // Construct the query.
         $variables = $this->get_variables();
-        $config = $variables->get('config');
-        $sql = $this->evaluate_expressions($config->sql);
+        $config = $variables->get_raw('config');
+        [$sql, $params] = $this->evaluate_expressions($config->sql);
```
brendanheywood, line 77: Should this pr touch lang packs / docs?

keevan: Potentially here: 
https://github.com/catalyst/moodle-tool_dataflows/blob/MOODLE_35_STABLE/VARIABLES.md

Or in the SQL steps docs

PR #731, opened 2023-05-03
Fix CSV encode to put an EOL at end of file.

jaypha:
Also add headers to beginning of file.

### DIFF
--- a/classes/local/formats/encoders/csv.php
+++ b/classes/local/formats/encoders/csv.php
@@ -27,13 +27,8 @@ use tool_dataflows\local\formats\encoder_base;
  * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 class csv extends encoder_base {
-    /** @var int escape with slashes. */
-    const ESC_SLASHES = 1;
-    /** @var int escape with double quotes. */
-    const ESC_DOUBLEQUOTES = 2;
-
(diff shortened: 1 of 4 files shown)


classes/local/formats/encoders/csv.php
```diff
@@ -42,57 +37,38 @@ class csv extends encoder_base {
...
+     * @throws \moodle_exception
      */
-    protected function escape_field_slashes(string $value): string {
-        if ((mb_strpos($value, '"') !== false) || (mb_strpos($value, ',') !== false) ||
-            (mb_strpos($value, "\n") !== false) || (mb_strpos($value, "\r") !== false)) {
-            return '"' . str_replace('"', '\\"', str_replace('\\', '\\\\', $value)) . '"';
-        } else {
-            return $value;
+    public static function str_putcsv(array $fields, string $delimiter = ',', string $enclosure = '"'): string {
+        $fp = fopen('php://memory', 'r+b');
```
keevan, line 62: Ah this is a very interesting approach. First time I've seen it done this way :)

keevan (approved):
Looks good to me.

classes/local/formats/encoders/csv.php
jaypha: https://gist.github.com/johanmeiring/2894568

PR #732, opened 2023-05-03
More helpful debugging

### LINKED ISSUE #723: SQL flow step doesn't expose the exception message for DML exceptions
The message contains the DB message, eg relation does not exist, so it should be thrown back out

Peterburnett:
image
Closes #723

### DIFF
--- a/classes/local/execution/engine.php
+++ b/classes/local/execution/engine.php
@@ -508,6 +508,7 @@ class engine {
     public function abort(?\Throwable $reason = null) {
         if (isset($reason)) {
             $message = $reason->getMessage();
+            $message .= PHP_EOL . $reason->debuginfo;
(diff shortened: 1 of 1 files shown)


brendanheywood (approved):

classes/local/execution/engine.php
```diff
@@ -508,6 +508,7 @@ public function finalise() {
     public function abort(?\Throwable $reason = null) {
         if (isset($reason)) {
             $message = $reason->getMessage();
+            $message .= PHP_EOL . $reason->debuginfo;
```
keevan, line 511: Would this be better?
```suggestion
            $message = $reason->getMessage() . PHP_EOL . $reason->debuginfo;
```

PR #733, opened 2023-05-03
Allow SQL to route via correct pathways for write queries

Peterburnett:


classes/local/step/flow_sql.php
```diff
@@ -76,18 +76,26 @@ public function execute($input = null) {
         $config = $variables->get_raw('config');
         [$sql, $params] = $this->evaluate_expressions($config->sql);
 
-        // Execute the query using get_records instead of get_record.
-        // This is so we can expose the number of records returned which
-        // can then be used by the dataflow in for e.g. a switch statement.
-        $records = $DB->get_records_sql($sql, $params);
+        // Now that we have the query, we want to get the first token to figure out what operation to run.
+        // This is not used for security, just to route the request via the correct pathway for readonly databases.
+        $token = strtok($sql, ' ');
```
keevan, line 81: Would this work?
```
(SELECT 1)
```
My test returned `(SELECT`

classes/local/step/flow_sql.php
```diff
@@ -76,18 +76,26 @@ public function execute($input = null) {
...
+        switch($token) {
+            case 'SELECT':
+                // Execute the query using get_records instead of get_record.
+                // This is so we can expose the number of records returned which
+                // can then be used by the dataflow in for e.g. a switch statement.
+                $records = $DB->get_records_sql($sql, $params);
+                $variables->set('count', count($records));
+                $invalidnum = ($records === false || count($records) !== 1);
+                $data = $invalidnum ? null : array_pop($records);
+                $variables->set('data', $data);
```
brendanheywood, line 92: is it worth setting these in the update case but setting them to unknown or null? that way the docs are consistent and the var tree is accruate

classes/local/step/flow_sql.php
brendanheywood, line 81: as dicsused regex this to be less brittle

classes/local/step/flow_sql.php
```diff
@@ -76,18 +76,36 @@ public function execute($input = null) {
         $config = $variables->get_raw('config');
         [$sql, $params] = $this->evaluate_expressions($config->sql);
 
-        // Execute the query using get_records instead of get_record.
-        // This is so we can expose the number of records returned which
-        // can then be used by the dataflow in for e.g. a switch statement.
-        $records = $DB->get_records_sql($sql, $params);
+        // Now that we have the query, we want to get the first token to figure out what operation to run.
+        // This is not used for security, just to route the request via the correct pathway for readonly databases.
+        $pattern = '/(SELECT|UPDATE|INSERT|DELETE)/m';
```
brendanheywood, line 81: this needs to be /i to be case insensitive

also the comment above it not strictly accurate as its not always the first token

classes/local/step/flow_sql.php
```diff
@@ -76,18 +76,36 @@ public function execute($input = null) {
...
-        $variables->set('data', $data);
+        switch($token) {
+            case 'SELECT':
+                // Execute the query using get_records instead of get_record.
+                // This is so we can expose the number of records returned which
+                // can then be used by the dataflow in for e.g. a switch statement.
+                $records = $DB->get_records_sql($sql, $params);
+                $variables->set('count', count($records));
+                $invalidnum = ($records === false || count($records) !== 1);
+                $data = $invalidnum ? null : array_pop($records);
```
brendanheywood, line 96: also wouldn't this be better to always be an array which is sometimes a zero length array rather than a null?

classes/local/execution/engine.php
```diff
@@ -508,7 +508,9 @@ public function finalise() {
     public function abort(?\Throwable $reason = null) {
         if (isset($reason)) {
             $message = $reason->getMessage();
-            $message .= PHP_EOL . $reason->debuginfo;
+            if (isset($reason->debuginfo)) {
+                $message .= PHP_EOL . $reason->debuginfo;
+            }
         } else {
             $message = '';
         }
```
keevan, line 516: We can probably clean this up a bit as well.
```
        $message = '';
        if (isset($reason)) {
            $message = $reason->getMessage();
            if (isset($reason->debuginfo)) {
                $message .= PHP_EOL . $reason->debuginfo;
            }
        }
```

classes/local/step/flow_sql.php
```diff
@@ -76,18 +76,36 @@ public function execute($input = null) {
...
+                $variables->set('data', $data);
+                break;
+            default:
+                // Default to execute.
+                $success = $DB->execute($sql, $params);
+                if ($success) {
+                    // We can't really do anything with the response except check for success.
+                    $variables->set('count', 1);
+                    $variables->set('data', null);
```
keevan, line 105: Can we set the data to $success? That might make it more robust. 

I also thought the count (of changes) was something that could be returned from the query execution, but I might be wrong here and it might be DB engine specific.

Peterburnett: execute only returns a bool in the DML. Ive updated the set to int coerce it directly instead

classes/local/step/flow_sql.php
```diff
@@ -76,18 +76,34 @@ public function execute($input = null) {
...
 
-        // Execute the query using get_records instead of get_record.
-        // This is so we can expose the number of records returned which
-        // can then be used by the dataflow in for e.g. a switch statement.
-        $records = $DB->get_records_sql($sql, $params);
+        // Now that we have the query, we want to get info on SQL keywords to figure out where to route the request.
+        // This is not used for security, just to route the request via the correct pathway for readonly databases.
+        $pattern = '/(SELECT|UPDATE|INSERT|DELETE)/im';
+        $matches = [];
+        preg_match($pattern, $sql, $matches);
```
brendanheywood, line 83: this is fine, but it might be neater to just use the boolean returned from preg_match instead of $token

brendanheywood: and its an if instead of a switch

brendanheywood (approved):

PR #734, opened 2023-05-04
Issue #696: Add validation to dataflow config form.

### LINKED ISSUE #696: Add validation for dataflow config
One can add an incorrect variable in the following field:
Screenshot from 2023-03-14 10-09-54

Then whenever calling method `get_vars` in admin/tool/dataflows/classes/dataflow.php, error will be thrown.

For instance, when selecting a step :

Screenshot from 2023-03-14 10-10-59

jaypha:
Resolves #696

### DIFF
--- a/classes/form/dataflow_form.php
+++ b/classes/form/dataflow_form.php
@@ -16,8 +16,8 @@
 
 namespace tool_dataflows\form;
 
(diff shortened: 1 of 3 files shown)


keevan (approved):
Looks good to me

PR #735, opened 2023-05-04
Add flow sftp step

jaypha:


classes/local/step/sftp_trait.php
```diff
@@ -0,0 +1,362 @@
...
+                return true;
+            }
+
+            // Upload to remote.
+            $this->upload($sftp, $sourcepath, $targetpath);
+        } finally {
+            $sftp->disconnect();
+        }
+
+        return true;
```
keevan, line 262: I believe this needs to return `$input|false` as it should work for both a connector and a flow.

keevan (commented):
There also needs to be a version bump. 

The rest looks good. Thanks Jason.

jaypha:
Bumped and rebased

keevan (approved):
Looks good to me

PR #738, opened 2023-05-08
Issue #729: Fix set variable step to not persist during dry runs.

### LINKED ISSUE #729: Set variable step should not save if it is a dry run.

jaypha:
Resolves #729 

Includes a trivial fix in GPG step.


keevan (approved):
Makes sense.

PR #740, opened 2023-05-08
Moodle 35 stable 690

marcghaly:
Hi @keevan,

I have found an alternate way to do what tickets 399328 demands, however this needs small changes included in this PR,

Regards,

Marc

### DIFF
--- a/classes/local/step/sql_trait.php
+++ b/classes/local/step/sql_trait.php
@@ -75,8 +75,8 @@ trait sql_trait {
                 [$hasexpression] = $parser->has_expression($el);
                 $max--;
             }
-            if (!is_string($el)) {
-                throw new \moodle_exception('reader_sql:finalsql_not_string', 'tool_dataflows');
+            if (!in_array(gettype($el), ['string', 'int'])) {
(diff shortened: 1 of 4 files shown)


keevan:
The change here seems OK to me. 

@Peterburnett can you confirm that this change is good as well? Since you've recently touched this space.

classes/local/step/sql_trait.php
```diff
@@ -75,7 +75,7 @@ private function evaluate_expressions(string $sql) {
                 [$hasexpression] = $parser->has_expression($el);
                 $max--;
             }
-            if (!is_string($el)) {
```
keevan, line 78: This change seems OK.

Should we update the error message since it now covers string and integers? It currently doesn't make sense either.

> SQL must evaluate to a string.

A possibility might be something like

> The SQL parameter must be a valid type (string, int)

I also wonder if the expression might read more easily, if we used `gettype` and the check could be something like 
```
if (!in_array(gettype($el), ['string', 'int']) {
    ...
}
```

marcghaly:
Hi @keevan, I moved trait-related strings to a specific section in lang file, used your suggestion and bumped the version,

Thank you !

keevan (approved):

Issue #741, opened 2023-05-10
copy_file_trait not allowing input data use


marcghaly:

jaypha:
There is a step for writing data to a file.
I suggest that, instead of modifying copy_file, create a read-from-file step.

PR #742, opened 2023-05-10
Issue #741: Allow input data to be used in copy file trait

marcghaly:
Hi @keevan,

This PR improves the copy_file_trait to allow input data to be used as a to / from parameter,

Regards,

Marc


jaypha:
There is already a step for putting content to a file. https://github.com/catalyst/moodle-tool_dataflows/blob/MOODLE_35_STABLE/classes/local/step/file_put_content_trait.php
It would be better to create a step to read from a file (e.g. '*_file_get_content').
Then this PR will not be needed.

PR #749, opened 2023-05-15
Moodle 35 stable 743 litteralsupport

marcghaly:


marcghaly:
Hi @keevan,

I have checked your comment and think the implementation you propose is better, I have made the changes

Thanks a lot !

Marc

classes/local/step/reader_json.php
```diff
@@ -84,6 +84,10 @@ protected function parse_json() {
             ['data' => $decodedjson]
         );
 
+        if (!is_array($returnarray)) {
+            $returnarray = (object) $returnarray;
+        }
```
keevan, line 89: Can you give me a real example where this is required?

Issue #751, opened 2023-05-18
Create new regex transformation step


andrewmadden:
Provided a string, the step should use regex to return the first substring that matches the pattern.

PR #752, opened 2023-05-18
issue #751: Add new regex transformation flow step

andrewmadden:


classes/local/step/flow_transformer_regex.php
```diff
@@ -0,0 +1,104 @@
...
+    public function execute($input = null) {
+        $variables = $this->get_variables();
+        $pattern = (string) $this->stepdef->config->pattern;
+        $field = (string) $this->stepdef->config->field;
+        $haystack = $variables->evaluate($field);
+        $matches = [];
+        preg_match($pattern, $haystack, $matches);
+        // Capture the first matched string as a variable.
+        $uniquekey = $variables->get('alias');
+        $input->$uniquekey = $matches[0] ?? null;
```
Peterburnett, line 101: I don't think ?? null actually does anything, as it evaluates to null to hit the second block. I think we either allow the null, or set it to an empty string.

classes/local/step/flow_transformer_regex.php
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * Apply the filter based on configuration
+     *
+     * @param  mixed $input
+     * @return mixed The new value to be passed on to the next step.
+     */
+    public function execute($input = null) {
+        $variables = $this->get_variables();
+        $pattern = (string) $this->stepdef->config->pattern;
```
Peterburnett, line 94: Are these string casts actually required? I would expect it gets coerced correctly during use

classes/step.php
```diff
@@ -69,7 +69,7 @@ protected static function define_properties(): array {
             'description' => ['type' => PARAM_TEXT, 'default' => ''],
             'type' => ['type' => PARAM_TEXT],
             'name' => ['type' => PARAM_TEXT],
-            'config' => ['type' => PARAM_TEXT, 'default' => ''],
+            'config' => ['type' => PARAM_RAW, 'default' => ''],
```
Peterburnett, line 72: This might be worth an explainer comment as to why.

tests/tool_dataflows_flow_transformer_regex_test.php
```diff
@@ -0,0 +1,175 @@
...
+namespace tool_dataflows;
+
+use Symfony\Component\Yaml\Yaml;
+use tool_dataflows\local\execution\engine;
+
+/**
+ * Unit test for append file step.
+ *
+ * @package   tool_dataflows
+ * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
```
Peterburnett, line 26: Update author here

classes/local/step/flow_transformer_regex.php
andrewmadden: Maybbeee not. I can remove them until it breaks again. Just found the symfony type coercion a bit aggressive/unpredictable. i.e. `1,2,3` coerced to `1.23`

classes/local/step/flow_transformer_regex.php
andrewmadden: It's a trick to avoid a throwable when the array is empty, and avoiding an if/else to check if there are any matches.

It's basically "If it's null, don't panic, set it to null"

https://onlinephp.io?s=s7EvyCjg5VLJzCsoLVGwVdDIT8pKTS7RVIiOtebl0tdXUMlNLEnOSC0GykWrZ6Tm5OSrg2QgGnTtwLJAOZiyaINYBXt7hbzSnBygqrLEoviU0twCDYhyTWsA&v=8.2.6

vs

https://onlinephp.io?s=s7EvyCjg5VLJzCsoLVGwVdDIT8pKTS7RVIiOtebl0tdXUMlNLEnOSC0GykWrZ6Tm5OSrg2QgGnTtwLJAOZiyaAOQbFliUXxKaW6BBkSZpjUA&v=8.2.6

classes/local/step/flow_transformer_regex.php
Peterburnett: Understood. In that case, leave it.

classes/local/step/flow_transformer_regex.php
Peterburnett: Ah right it suppresses the warning. That flew over my head, never seen that one before. I typically just use

`array_key_exists() ? $array['key'] : null;`

Thats a neat trick, ill keep it in mind if I ever need it

Peterburnett: There is a null safe check operator now though in PHP 8 which can likely make this better when its exclusive,

https://stitcher.io/blog/php-8-nullsafe-operator

tests/tool_dataflows_flow_transformer_regex_test.php
andrewmadden: updated

classes/local/step/flow_transformer_regex.php
andrewmadden: I removed them and added a unit test. I think it should be fine. Only the `field` gets evaluated by symfony expressions and it should be a variable.

classes/local/step/flow_transformer_regex.php
andrewmadden: Yeah the nullsafe operator is cool. It's a pretty key part of swift. Will definitely start using it more as the support moves up.

The null coalescing operator is very similar logic https://www.decodingweb.dev/php-double-question-marks

Peterburnett (approved):

PR #753, opened 2023-05-18
Add remove file steps.

jaypha:


tests/tool_dataflows_remove_file_test.php
```diff
@@ -0,0 +1,99 @@
...
+        $namespace = '\\tool_dataflows\\local\\step\\';
+
+        $dataflow = new dataflow();
+        $dataflow->enabled = true;
+        $dataflow->name = 'dataflow';
+        $dataflow->save();
+
+        $reader = new step();
+        $reader->name = 'hoover';
+        $reader->type = $namespace . 'connector_remove_file';
```
keevan, line 91: I prefer the `class::classname` approach since IDEs can more easily pick up this info, and it's a bit clearer if it doesn't exist.

Could we include `use tool_dataflows\local\step\connector_remove_file;` and use `connector_remove_file::classname` here instead?

jaypha: done

keevan (approved):
Looks good.

PR #755, opened 2023-05-18
Moodle 35 stable 754 jsonflow

marcghaly:
Hi @keevan,

This is a PR that allows the JSON read to be used in subsequent step as a flow,

Regards,

Marc


keevan:
Thanks Marc, can we please add some unit tests to test for any extra functionality we are adding here.

PR #765, opened 2023-05-29
feat(append): add append connector step

### LINKED ISSUE #764: Append connector step doesn't exist
It has a flow and trait step but not a connector step. Adding this should be trivial.

keevan:
Closes #764


lang/en/tool_dataflows.php
```diff
@@ -119,6 +119,7 @@
 $string['step_name_connector_directory_file_count'] = 'Directory file count';
 $string['step_name_connector_directory_file_list'] = 'Directory file list';
 $string['step_name_connector_abort'] = 'Abort connector';
+$string['step_name_connector_append_file'] = 'Append file';
```
Peterburnett, line 122: Would this be more clear as 'Append file contents' ?

Peterburnett (approved):

lang/en/tool_dataflows.php
keevan: Perhaps. I used a similar naming as the copy file step, and if we change this as well, we should change probably change the step name itself, and the lang string for both flow and connector steps to match.

PR #779, opened 2023-06-06
[#778] bugfix: register schedule when importing cron steps

### LINKED ISSUE #778: Importing dataflow with cron trigger does not create schedule record, and so is never run
Importing dataflow with cron trigger does not create `mdl_tool_dataflows_schedule` record, and so therefore never get run.

**Recreate:**
- Totara 13
- Import the dataflow attached to this issue (rename to .yml)
- Confirm there is no record in `mdl_tool_dataflows_schedule`
- Confirm when running dataflows scheduled task or adhoc task, the dataflow is never run

[Test_flow_2_21313123_20230606_0157.txt](https://github.com/catalyst/moodle-tool_dataflows/files/11659181/Test_flow_2_21313123_20230606_0157.txt)

matthewhilton:
Closes #778 

-  <strike>Since we cannot override the persistant's `save` function, I made a new function `save_step` which calls both `save()` and any `on_save()` callbacks</strike>
-  <strike>Hopefully this means in the future we don't run into this issue for anything else that uses this callback, as long as we are using `save_step`</strike>

- Have just added hook callbacks for after updating and after creating to trigger the steptype and dataflow save callbacks.
- Added unit tests for the cron trigger to unit test this as well

**Manual testing:**

- Upload the attached yml file [Test_flow_2_21313123_20230606_0157.txt](https://github.com/catalyst/moodle-tool_dataflows/files/11659409/Test_flow_2_21313123_20230606_0157.txt)
- Check that a record was made in `mdl_tool_dataflows_schedule`

### DIFF
--- a/classes/step.php
+++ b/classes/step.php
@@ -647,6 +647,30 @@ class step extends persistent {
     }
 
     /**
+     * Called after updating.
+     *
+     * Triggers steptype on_save callback and dataflow on_steps_save callback.
+     *
+     * @param bool $result True if the delete was successful.
+     */
+    protected function after_update($result) {
+        if ($result) {
+            $this->get_steptype()->on_save();
+            $this->get_dataflow()->on_steps_save();
+        }
+    }
+
+    /**
+     * Called after creating.
+     *
+     * Triggers steptype on_save callback and dataflow on_steps_save callback.
+     */
+    protected function after_create() {
+        $this->get_steptype()->on_save();
+        $this->get_dataflow()->on_steps_save();
+    }
+
+    /**
      * Sets the validated flag to false, such that validation can take place on
      * the persistent again.
      */
--- a/tests/fixtures/sample-cron.yml
+++ b/tests/fixtures/sample-cron.yml
@@ -1,0 +1,24 @@
+name: Example dataflow with Cron trigger
+
+steps:
+  cron:
+    name: cron
+    type: tool_dataflows\local\step\trigger_cron
+    config:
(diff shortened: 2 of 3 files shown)


classes/dataflow.php
```diff
@@ -605,7 +605,7 @@ public function import(array $yaml) {
 
                     $step->import($stepdata);
                     // Save persistent and base fields.
-                    $step->save();
```
brendanheywood, line 608: should save() never be called at all? If so can we protect against this somehow?

matthewhilton: Ideally yes but not sure how we can stop it since its a final function as defined by the persistent class

matthewhilton: have switched to using the persistent callbacks.

brendanheywood (approved):

classes/step.php
```diff
@@ -646,6 +646,30 @@ protected function after_delete($result) {
...
+     *
+     * Triggers steptype on_save callback and dataflow on_steps_save callback.
+     *
+     * @param bool $result True if the delete was successful.
+     */
+    protected function after_update($result) {
+        if ($result) {
+            $this->get_steptype()->on_save();
+            $this->get_dataflow()->on_steps_save();
+        }
```
keevan, line 660:
```suggestion
        if (!$result) {
            return;
        }
        $this->get_steptype()->on_save();
        $this->get_dataflow()->on_steps_save();
```

PR #780, opened 2023-06-06
768 fix concurrency

### LINKED ISSUE #768: Important bug with concurrency of adhoc tasks
There is some edge cases happening where a cron triggered dataflow is running more than once a minute. This might be only in totara which is effectively 3.3

* [ ] rename the method to something more meaningful: execute_from_record -> queue_adhoctask

* [ ] deduplicate tasks in a way when falling back to 3.3 / and concurrency is not supported for a specific flow

queue_adhoc_task $checkforexisting = true

https://github.com/catalyst/moodle-tool_dataflows/blob/23bcdcb3973991df07ea7ef03f62fb921c2c589a/classes/task/process_dataflow_ad_hoc.php#LL63C8-L63C53

* [ ] double check the data fed into the adhoc tasks to deduplicate is minimal data so they do deduplicate correctly, ie it is just a dataflow id and nothing else

* [ ] remove the special case where we run one data flow right away, make them all adhoc tasks. Not strictly needed but might make the code more readable. If we need to run 5 tasks and we keep this code then it should NOT spawn a 5th adhoc task if it is going to run it directly itself

matthewhilton:
Closes #768 

## Changes
- Now when scheduled task is run, it simply makes an adhoc task for each dataflow. It does not try to run one in the same thread - this is much cleaner and simpler.
- The adhoc tasks now only take in a dataflow id - no other custom data is allowed. This ensures they can be de-duplicated
- Will de-duplicate adhoc tasks based on concurrency settings

## Testing
-  Covered by unit tests.

### DIFF
--- a/classes/task/process_dataflow_ad_hoc.php
+++ b/classes/task/process_dataflow_ad_hoc.php
@@ -37,29 +37,26 @@ class process_dataflow_ad_hoc extends \core\task\adhoc_task {
     }
 
     /**
-     * Create an ad-hoc task for the given dataflow record
+     * Create an ad-hoc task for the given dataflow
      * and schedules for it to be run.
      *
-     * @param object $dataflowrecord
+     * @param dataflow $dataflow
      * @return void
      */
-    public static function execute_from_record($dataflowrecord): void {
-        $dataflow = new dataflow($dataflowrecord->dataflowid);
+    public static function queue_adhoctask(dataflow $dataflow): void {
         $task = new process_dataflow_ad_hoc();
-        $task->set_custom_data($dataflowrecord);
 
-        // For concurrent tasks, queue them up as an independant adhoc task.
-        if ($dataflow->is_concurrency_enabled()) {
-            \core\task\manager::queue_adhoc_task($task);
-            return;
-        }
+        // No extra data can go in here, otherwise the de-duplication will not work.
+        $customdata = [
+            'dataflowid' => $dataflow->get('id'),
+        ];
+        $task->set_custom_data($customdata);
 
-        // Only available 3.7 onwards.
-        if (function_exists('\core\task\manager::reschedule_or_queue_adhoc_task')) {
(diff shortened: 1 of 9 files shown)


classes/task/process_dataflow_ad_hoc.php
```diff
@@ -37,29 +37,27 @@ public function execute() {
...
+     * Create an ad-hoc task for the given dataflow
      * and schedules for it to be run.
      *
-     * @param object $dataflowrecord
+     * @param int $dataflowid
      * @return void
      */
-    public static function execute_from_record($dataflowrecord): void {
-        $dataflow = new dataflow($dataflowrecord->dataflowid);
+    public static function queue_adhoctask(int $dataflowid): void {
```
brendanheywood, line 46: The custom data inside the adhoc task should be just the bare minimum, but I think if we pass in the data flow record or object here is would save additional db lookups

classes/task/process_dataflows.php
```diff
@@ -45,20 +45,17 @@ public function get_name() {
      * Run the dataflows.
      */
     public function execute() {
-        $scheduledrecords = scheduler::get_due_dataflows();
+        // When unit testing, set the reftime to the future, so that tasks are always waiting to run.
+        $reftime = PHPUNIT_TEST ? time() + 100 : null;
```
brendanheywood, line 49: Not sure I grok this

brendanheywood: I think the cleaner way to do this is in the unit test you create the flows which creates the schedule, and then you can call scheduler::set_scheduled_times to set its next run to right now

classes/task/process_dataflow_ad_hoc.php
brendanheywood: but this is optional bonus points, not critical if its a hassle

matthewhilton: Yep nice idea - have updated it

brendanheywood (approved):

PR #782, opened 2023-06-15
Issue #781: Add config to set webservice step params as lowercase

### LINKED ISSUE #781: Webservice step sets all params to lowercase
The webservice step sets all params to lowercase https://github.com/catalyst/moodle-tool_dataflows/blob/MOODLE_35_STABLE/classes/local/step/flow_web_service.php#L216 This is not always required so should be set as a config setting that can be enabled when required.

petersistrom:
Resolves #781

### DIFF
--- a/classes/local/step/flow_web_service.php
+++ b/classes/local/step/flow_web_service.php
@@ -211,10 +211,6 @@ EOF;
         // Ensure structures are arrays for webservice.
         $replacement = json_decode(json_encode($replacement), true);
         $params = array_replace_recursive($params, $replacement);
-        // If password is need will throw exception.
-        array_walk_recursive($params, function (&$item) {
-            $item = strtolower($item);
-        });
         $failure = !isset($config->failure) ? 'abortflow' : $config->failure;
 
         if (external_api::external_function_info($config->webservice)->type === 'read') {
--- a/classes/local/provider/expression_provider.php
+++ b/classes/local/provider/expression_provider.php
@@ -66,6 +66,7 @@ class expression_provider implements ExpressionFunctionProviderInterface {
             ExpressionFunction::fromPhp('date', 'date'),
             ExpressionFunction::fromPhp('get_object_vars', 'get_object_vars'),
             ExpressionFunction::fromPhp('strtotime', 'strtotime'),
+            ExpressionFunction::fromPhp('strtolower', 'strtolower'),
         ];
     }
 }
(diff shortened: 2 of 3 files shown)


classes/local/step/flow_web_service.php
```diff
@@ -57,6 +57,7 @@ public static function form_define_fields(): array {
             'parameters' => ['type' => PARAM_RAW],
             'failure' => ['type' => PARAM_TEXT, 'required' => true],
             'path' => ['type' => PARAM_PATH],
+            'paramslowercase' => ['type' => PARAM_RAW],
```
matthewhilton, line 60: Shouldn't this be `PARAM_BOOL` ?

petersistrom: Yes, yes it should! Updated

brendanheywood:
My preference would be for this not to land. I don't think the original strlower should not have landed and my preference would be for this to be handled as a new expression helper here:

https://github.com/catalyst/moodle-tool_dataflows/blob/9c3ce2e55ab78b03b00306de260a23e474893720/classes/local/provider/expression_provider.php#L68

This means you could selectively lowercase (or uppercase) specific vars as you need to.

@marcghaly I appreciate this may affect you, we could write an upgrade script to try and fix the steps that do this but if there is only a small number of cases in the wild using this it might be better to manually fix these and pretend it was never there.

Happy to hear arguments on the other side of the fence :)

brendanheywood (changes requested):
See comment:

https://github.com/catalyst/moodle-tool_dataflows/pull/782#issuecomment-1592456653

petersistrom:
I've updated the patch to use the expression provider. @marcghaly Could you please let us know your thoughts and how it will affect existing flows?

classes/local/step/flow_web_service.php
```diff
@@ -212,9 +212,6 @@ public function execute($input = null) {
         $replacement = json_decode(json_encode($replacement), true);
         $params = array_replace_recursive($params, $replacement);
         // If password is need will throw exception.
```
keevan, line 214: Hi, could we also clean up this comment as I'm not sure how this would happen. Is this still relevant? Does this comment belong here?

Thanks!

petersistrom: I've removed the comment, not sure what it is referring to. I agree doesn't look like it belongs there.

petersistrom:
Hi @brendanheywood Just a bump on this one, hoping to get it merged in soon. Do you think there is anything else required to get it merged?

brendanheywood (approved):

Issue #783, opened 2023-06-19
Directory file list improvements to listing


keevan:
Currently, only the base name for the files matched are returned. Whilst this is good for files in the same directory, or files anywhere that need to be enumerated in a flat structure, this is not idea for when you need to find and do something with the file, since the path information is lost.

I'm proposing we move the basename() handling to another step, and have the base functionality either return:
- a list of relative paths from the directory specified (default)
- a list of absolute paths

This means subsequent steps won't need to use `${{some_directory}}${{record.filename}}`,  and can potentially use the `${{record.filename}}` value directly.

When relative pathing becomes the new default, if the specified files were in the same directory specified, this would not be a breaking change. The only time this would be a breaking change is if the usage was purely for base filename enumeration.

keevan:
I ended up leaving the basename functionality as the default since I didn't want to introduce a breaking change without a second opinion :)

PR #784, opened 2023-06-19
#783 improve directory file list step

keevan:
- feat(directory-file-list): add different return value formats: absolutepath, basename, relativepath
- feat(file-directory-list): add glob braces for multi pattern support

Relative path output with glob brace support, pattern of `{*,**/*}`
image

Relative path output with pattern `**/*`
image

Absoulte path output, pattern `**/*`
image

Basename (which is the default), pattern `**/*`
image

### DIFF
--- a/classes/local/step/directory_file_list_trait.php
+++ b/classes/local/step/directory_file_list_trait.php
@@ -137,11 +137,11 @@ trait directory_file_list_trait {
             $this->get_engine()->abort($error);
         }
 
-        $path = $path . DIRECTORY_SEPARATOR . $pattern;
-        $filelist = glob($path);
+        $pathpattern = $path . DIRECTORY_SEPARATOR . $pattern;
+        $filelist = glob($pathpattern, GLOB_BRACE);
 
         // Apply filter for excluding directories/folders.
-        // This is not related to recursive lookups in any way
+        // This is not related to recursive lookups in any way.
         $filelist = array_filter($filelist, function ($pathname) use ($includedir) {
             $filename = basename($pathname);
             if ($filename === '.' || $filename === '..') {
--- a/classes/local/step/directory_file_list_trait.php
+++ b/classes/local/step/directory_file_list_trait.php
@@ -30,10 +30,15 @@ trait directory_file_list_trait {
 
     /** @var string The default pattern for matching files. */
     static protected $patterndefault = '*';
+
     /** @var int The default offset. */
     static protected $offsetdefault = 0;
+
     /** @var int The default limit (effectively infinity). */
     static protected $lmiitdefault = 0;
+
+    /** @var string The format of the return value */
+    static protected $returnvaluedefault = 'basename';
 
     /**
      * Return the definition of the fields available in this form.
@@ -44,6 +49,7 @@ trait directory_file_list_trait {
         return [
             'directory'   => ['type' => PARAM_TEXT],
             'pattern' => ['type' => PARAM_TEXT],
+            'returnvalue' => ['type' => PARAM_TEXT],
             'sort' => ['type' => PARAM_TEXT],
             'subdirectories'   => ['type' => PARAM_BOOL],
             'offset'     => ['type' => PARAM_INT],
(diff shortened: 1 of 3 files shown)


classes/local/step/directory_file_list_trait.php
```diff
@@ -105,6 +118,7 @@ public function run() {
 
         $pattern  = $config->pattern ?: self::$patterndefault;
         $offset = $config->offset ?: self::$offsetdefault;
+        $returnvalue = $config->returnvalue ?? self::$returnvaluedefault;
```
Peterburnett, line 121: Worth using a ternary as above/below? Just seems a tiny bit out of place

classes/local/step/directory_file_list_trait.php
```diff
@@ -123,10 +137,11 @@ public function run() {
...
         }
 
-        $path = $path . DIRECTORY_SEPARATOR . $pattern;
-        $filelist = glob($path);
+        $pathpattern = $path . DIRECTORY_SEPARATOR . $pattern;
+        $filelist = glob($pathpattern, GLOB_BRACE);
 
-        // Apply filter for excluding subdirectories.
+        // Apply filter for excluding directories/folders.
+        // This is not related to recursive lookups in any way
```
Peterburnett, line 144: fulltstop at the end, or CI will be mad

classes/local/step/directory_file_list_trait.php
```diff
@@ -138,10 +153,21 @@ public function run() {
             return true;
         });
 
-        // Strip out the path.
-        $filelist = array_map(function ($filename) {
-            return basename($filename);
-        }, $filelist);
+        // Convert the original to return the base file name.
+        // For example, "/my/path/to/file" would return "file".
+        if ($returnvalue === 'basename') {
```
Peterburnett, line 158: worth pulling these out to a set of constants in the class?

classes/local/step/directory_file_list_trait.php
```diff
@@ -138,10 +153,21 @@ public function run() {
...
+            $filelist = array_map(function ($filename) {
+                return basename($filename);
+            }, $filelist);
+        }
+
+        // Convert the original to return the relative file path.
+        // For example, "/my/path/to/file" would return "to/file".
+        if ($returnvalue === 'relativepath') {
```
Peterburnett, line 166: Also this almost looks like it would be cleaner as a switch. Dealers choice

lang/en/tool_dataflows.php
```diff
@@ -530,18 +530,23 @@
...
+$string['directory_file_list:alpha'] = 'Alphabetical (A->Z)';
+$string['directory_file_list:alpha_reverse'] = 'Reverse alphabetical (Z->A)';
+$string['directory_file_list:basename'] = 'Base Name';
+$string['directory_file_list:filenames'] = 'List of filenames.';
+$string['directory_file_list:limit'] = 'Limit';
+$string['directory_file_list:offset'] = 'Offset';
 $string['directory_file_list:pattern'] = 'File pattern';
-$string['directory_file_list:sort'] = 'Sort';
+$string['directory_file_list:relativepath'] = 'Relative Path';
+$string['directory_file_list:returnvalue'] = 'Return Value';
```
Peterburnett, line 542: Dupe string

classes/local/step/directory_file_list_trait.php
keevan: Nice catch. Thanks, I need to properly set up those lints on my new editor.

classes/local/step/directory_file_list_trait.php
keevan: Maybe. Maybe not. This feels like a personal preference, and I would avoid the switch for 3 small reasons.

- Switch is good for small simple conditions (which this is) and small, straightforward cases. This would has an array map which adds to the complexity a bit. It might be easier to reason about as an independent conditional block, at least that's what I think.
- Conditions in switches are checked loosely, not strict type checks. 
- There's more jumping around visually for switch statements, by their nature. The condition is the combination of the value in the opening switch, and the case. They are both located separately, compared with a traditional `if`, where the condition is laid out on the same line and reads well. Switches might be better when the list of cases grows, but for 2 I think there isn't much benefit `switch`-ing.

Here's an example of both changes side by side (vertically):
image

classes/local/step/directory_file_list_trait.php
keevan: I had a look at the other ones, and it seems like the defaults are better as-is, due to the other fields being text input (as opposed to a combobox). 

So I'm thinking no action is required here.

Peterburnett (approved):

PR #789, opened 2023-06-23
786 add readonly flag to lock dataflow settings

keevan:
- style: phpcs fixes
- feat: add readonly flag to lock dataflow settings
- feat: add readonly checks to all the forms

### DIFF
--- a/classes/admin/admin_setting_configcheckbox.php
+++ b/classes/admin/admin_setting_configcheckbox.php
@@ -16,31 +16,17 @@
 
 namespace tool_dataflows\admin;
 
-use Symfony\Component\Yaml\Yaml;
+use tool_dataflows\helper;
+use tool_dataflows\admin\readonly_trait;
 
 /**
- * A setting textarea for YAML format values.
+ * A custom setting for the permitted directories that a dataflow can interact with.
  *
  * @package   tool_dataflows
- * @author    Jason den Dulk <jasondendulk@catalyst-au.net>
- * @copyright 2022, Catalyst IT
+ * @author    Kevin Pham <kevinpham@catalyst-au.net>
+ * @copyright Catalyst IT, 2023
  * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
-class admin_setting_yaml extends \admin_setting_configtextarea {
+class admin_setting_configcheckbox extends \admin_setting_configcheckbox {
     use readonly_trait;
-
-    /**
-     * Validate the setting data as a YAML config block.
-     *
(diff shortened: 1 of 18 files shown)


classes/admin/admin_setting_configcheckbox.php
```diff
@@ -0,0 +1,32 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_dataflows\admin;
+
+use tool_dataflows\helper;
```
Peterburnett, line 19: Is the helper use required?

classes/form/dataflow_form.php
```diff
@@ -16,8 +16,10 @@
 
 namespace tool_dataflows\form;
 
+use moodle_exception;
```
Peterburnett, line 19: can we prepend the \ and remove the use? Not sure if that is still the recommended coding style

classes/form/import_form.php
```diff
@@ -14,7 +14,10 @@
 // You should have received a copy of the GNU General Public License
 // along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
 
-namespace tool_dataflows;
+namespace tool_dataflows\form;
+
+use moodle_exception;
```
Peterburnett, line 19: same here

classes/admin/admin_setting_configcheckbox.php
keevan: Good pickup. Will remove.

classes/form/dataflow_form.php
keevan: Per the coding style example, a similar call is made to moodle_url where it had been included via `use XXX`. Since that was categorised as good usage, I'm leaning towards this staying.

https://moodledev.io/general/development/policies/codingstyle#further-notes

We've agreed in internal chat, and I'll resolve this point. Thanks for pointing it out.

classes/form/import_form.php
keevan: Similar to the previous point:

Ref: https://moodledev.io/general/development/policies/codingstyle#further-notes

keevan:
Added to view, edit, step (form), manage pages. Please let me know if I missed anything.

Thanks!

Peterburnett (approved):

PR #791, opened 2023-06-27
feat: add webservice trigger

### LINKED ISSUE #790: Webservice trigger step
Similar to https://github.com/catalyst/moodle-tool_dataflows/issues/92, but this will expect and confirm to Moodle's service api.

keevan:
Currently adds the dataflow to the adhoc task queue, and does not process it in place. This is typically the recommended default for api calls.

Closes #790

### DIFF
--- a/classes/external/trigger_dataflow.php
+++ b/classes/external/trigger_dataflow.php
@@ -20,6 +20,7 @@ use external_function_parameters;
 use external_single_structure;
 use external_value;
 use tool_dataflows\dataflow;
+use tool_dataflows\local\step\trigger_webservice;
 use tool_dataflows\task\process_dataflow_ad_hoc;
 
 /**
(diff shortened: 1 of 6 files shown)


db/services.php
```diff
@@ -0,0 +1,46 @@
...
+        'description' => 'Triggers a dataflow.',
+
+        // Options include read, and write.
+        'type'        => 'write',
+
+        // Whether the service is available for use in AJAX calls from the web.
+        'ajax'        => false,
+
+        // An optional list of services where the function will be included.
+        'services' => []
```
Peterburnett, line 44: lets make a default service here that only includes the trigger function, that way we can generate a token specifically for this just a bit easier.

classes/external/trigger_dataflow.php
```diff
@@ -0,0 +1,77 @@
...
+     *
+     * Currently it only supports asynchronous execution.
+     *
+     * @param array $dataflow array of dataflow run information.
+     * @return array of newly created groups
+     */
+    public static function execute($dataflow) {
+        $params = self::validate_parameters(self::execute_parameters(), ['dataflow' => $dataflow]);
+
+        $dataflow = new dataflow($params['dataflow']['id']);
```
Peterburnett, line 68: my gut feel here is that while we dont need a trigger specifically for it, I still want to have one I think. I like the idea that you can't execute an arbitrary dataflow, ONLY ones that opt into it via a webservice trigger. So basically, if step 0 isnt a webservice trigger, just throw a warning and do nothing.

keevan:
Example webservice call against 2 dataflows
id: 2 has a webservice trigger step, id: 1 does not and returns the error as shown below
image

When a WS to trigger a dataflow is called, but it is invalid, it will throw an exception (and return the relevant response), instead of the suggested warn and do nothing approach.

I also added a services array based on search elastic, and filled in the optional service array as per the docs here: https://moodledev.io/docs/apis/subsystems/external/writing-a-service#declare-the-web-service-function

Peterburnett (approved):

PR #795, opened 2023-06-30
fix: set default delimeter when delimeter value is not truthy

### LINKED ISSUE #794: CSV reader - delimiter must be a character
`Warning:  fgetcsv(): delimiter must be a character in /var/www/client/admin/tool/dataflows/classes/local/step/reader_csv.php on line 85`

For some reason, it's not defaulting to a comma `,` though the placeholder suggests that's the value if nothing is set. If it doesn't default, it should be a validation error, if it does default, then it should not show this warning.

keevan:
This fixes the warning when no delimeter is set, and it's unlikely that the delimeter is intentionally an empty value.

Closes #794

### DIFF
--- a/classes/local/step/reader_csv.php
+++ b/classes/local/step/reader_csv.php
@@ -62,6 +62,7 @@ class reader_csv extends reader_step {
         $maxlinelength = 1000;
         $config = $this->get_variables()->get('config');
         $strheaders = $config->headers;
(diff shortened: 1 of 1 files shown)


keevan:
Just fixed up CI issue with `self` method call

Peterburnett (approved):

PR #798, opened 2023-07-03
[#797] fix: sql flow step to switch cases in case insensitive manner

### LINKED ISSUE #797: SQL flow step match not working for different cases
https://github.com/catalyst/moodle-tool_dataflows/blob/09b7d2f8c0c2b307febf97251fa5b96e5b8e4933/classes/local/step/flow_sql.php#L81-L88

Here the check is case insensitive but the switch-case statement assumes a particular case (upper), this is causing dropped results from SQL queries which do not have the proper casing.

keevan:
Closes #797

### DIFF
--- a/classes/local/step/flow_sql.php
+++ b/classes/local/step/flow_sql.php
@@ -85,7 +85,7 @@ class flow_sql extends flow_step {
         // Matches[0] contains the match. Fallthrough to default on no match.
         $token = $matches[0] ?? '';
 
-        switch($token) {
+        switch(strtoupper($token)) {
             case 'SELECT':
                 // Execute the query using get_records instead of get_record.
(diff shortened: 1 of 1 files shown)


classes/local/step/flow_sql.php
```diff
@@ -85,7 +85,7 @@ public function execute($input = null) {
         // Matches[0] contains the match. Fallthrough to default on no match.
         $token = $matches[0] ?? '';
 
-        switch($token) {
+        switch(strtoupper($token)) {
```
keevan, line 88: Potentially could transform the case to upper in a prior line. Might be cleaner to read. Let me know anyways :)

Peterburnett (approved):

PR #803, opened 2023-07-05
796 fix moodle 42 fatal with core external lib changes

keevan:
- no longer using externallib.php
- fix: add backwards compatibility layer for external_lib usage under Moodle 4.2


version.php
```diff
@@ -25,8 +25,8 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-$plugin->version = 2023063000;
-$plugin->release = 2023063000;
+$plugin->version = 2023070500;
+$plugin->release = 2023070500;
 $plugin->requires = 2017051500;    // Our lowest supported Moodle (3.3.0).
 $plugin->supported = [35, 401];    // Available as of Moodle 3.9.0 or later.
```
brendanheywood, line 31: bump this to 402

Peterburnett:
image
 
 Just a small CI blocker then this should be good

PR #806, opened 2023-07-14
feat: SFTP directory file list steps 

keevan:
- feat: sftp directory list connector step
- feat: reader step for SFTP directory listing

Preview of what the settings look like:

image

### DIFF
--- a/classes/local/step/directory_file_list_trait.php
+++ b/classes/local/step/directory_file_list_trait.php
@@ -16,7 +16,6 @@
 
 namespace tool_dataflows\local\step;
 
-use MoodleQuickForm;
 use tool_dataflows\helper;
 
 /**
@@ -171,19 +170,21 @@ trait directory_file_list_trait {
     /**
      * Apply various constraints/transforms to the list of files
      *
-     * @param  array<string> $filelist
-     * @param  mixed         $returnvalue
-     * @param  mixed         $offset
-     * @param  mixed         $limit
+     * @param  array         $filelist
+     * @param  string        $returnvalue
+     * @param  string        $sort
+     * @param  int           $offset
+     * @param  int           $limit
      * @param  bool          $includedir
-     * @param  string        $path
+     * @param  string        $basepath
+     * @param  string        $pattern
      * @return array Will return true or an array of errors.
      */
     public function apply_list_constraints(
         array $filelist,
-        $returnvalue,
+        string $returnvalue,
         string $sort,
-        $offset,
+        int $offset,
         ?int $limit,
         bool $includedir,
         string $basepath,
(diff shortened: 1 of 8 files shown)


classes/local/step/directory_file_list_trait.php
```diff
@@ -195,4 +166,86 @@ public function validate_for_run() {
...
+     * @param  string        $returnvalue
+     * @param  string        $sort
+     * @param  int           $offset
+     * @param  int           $limit
+     * @param  bool          $includedir
+     * @param  string        $basepath
+     * @param  string        $pattern
+     * @return array Will return true or an array of errors.
+     */
+    public function apply_list_constraints(
```
Peterburnett, line 183: This feels like it should probably be 3 functions (filter, path transform, sort), but this is still a good step to cleaning up the complexity of the execute function

keevan: I agree. I think it should probably be pulled out when the steps that use this trait want to apply different logic, e.g. to how it sorts, or how it transforms the path.

I'm a bit cautious of making this into an issue if it doesn't apply anywhere else just yet. Do you have an example of where this might prove to be useful? I'm not sure if it's worth adding 3x function calls to every step that uses this trait at the moment.

Peterburnett: No this is fine. Just something that popped into my mind, but this is perfectly sufficient.

Peterburnett (approved):

classes/local/step/reader_sftp_directory_file_list.php
```diff
@@ -0,0 +1,54 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_dataflows\local\step;
+
+use tool_dataflows\local\execution\iterators\iterator;
+use tool_dataflows\local\execution\iterators\dataflow_iterator;
```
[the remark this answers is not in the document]
keevan: https://moodledev.io/general/development/policies/codingstyle#further-notes

> Each use statement must be on its own line.

:-)

classes/local/step/sftp_directory_file_list_trait.php
```diff
@@ -0,0 +1,128 @@
...
+ *
+ * @package   tool_dataflows
+ * @author    Kevin Pham <kevinpham@catalyst-au.net>
+ * @copyright Catalyst IT, 2023
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+trait sftp_directory_file_list_trait {
+    // Utilises the functionality in both traits for a combined step.
+    use sftp_trait {
+        sftp_trait::form_define_fields as sftp_form_define_fields;
```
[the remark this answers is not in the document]
keevan: haha, this would be a nightmare if the aliases were split out classic style.

PR #808, opened 2023-07-14
feat: config to overwrite headers for csv reader

### LINKED ISSUE #807: CSV headers needs the ability to overwrite the headers
The current behaviour is to add a headers row before the main data, which works if the file is set up that way. 

But if the columns are fixed, we can reduce a potential mapping step and replace the would-be fields to a fixed list of headers. In particular this is to resolve the limitation of referencing fields with spaces and symbols which normally couldn't be referenced through the expression language syntax, such as `field.with.dot`, `Field Name With Spaces`, etc.

Potentially other odd edge cases as well, which this would resolve.

keevan:
Resolves #807

image
This is what it looks like.

Added a test to ensure it works as expected.

### DIFF
--- a/classes/local/step/reader_csv.php
+++ b/classes/local/step/reader_csv.php
@@ -42,6 +42,7 @@ class reader_csv extends reader_step {
         return [
             'path' => ['type' => PARAM_TEXT, 'required' => true],
             'headers' => ['type' => PARAM_TEXT],
+            'overwriteheaders' => ['type' => PARAM_BOOL],
             'delimiter' => ['type' => PARAM_TEXT],
         ];
     }
@@ -62,6 +63,7 @@ class reader_csv extends reader_step {
         $maxlinelength = 1000;
         $config = $this->get_variables()->get('config');
         $strheaders = $config->headers;
+        $overwriteheaders = !empty($config->overwriteheaders);
         $delimiter = $config->delimiter ?: self::DEFAULT_DELIMETER;
         $path = $this->enginestep->engine->resolve_path($config->path);
 
@@ -71,7 +73,7 @@ class reader_csv extends reader_step {
 
         try {
             // Prepare and resolve headers.
-            if (empty($strheaders)) {
+            if (empty($config->headers)) {
                 $strheaders = fgets($handle);
             }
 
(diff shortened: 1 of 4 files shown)


classes/local/step/reader_csv.php
```diff
@@ -81,6 +83,12 @@ public function csv_contents_generator() {
                 return;
             }
 
+            // Used the configured headers in place of the first row if overwriteheaders is true.
```
Peterburnett, line 86: Will this lose data if there arent headers configured in the file?

tests/tool_dataflows_reader_csv_test.php
```diff
@@ -159,12 +162,60 @@ public function test_bad_input_stream() {
...
+        $reader->config = Yaml::dump([
+            'path' => $this->inputpath,
+            'headers' => implode(',', $headers),
+            'overwriteheaders' => '1',
+            'delimiter' => ',',
+        ]);
+
+        // Create test input file.
+        $data = [
+            ['apple', 'berry', 'cucumber'],
```
Peterburnett, line 182: can we also just add a test for a file with no headers, but override headers set

Peterburnett: ignore, CSV is untyped, this isnt easy to do

classes/local/step/reader_csv.php
Peterburnett: Probably, but this is fine. ignore comment

classes/local/step/reader_csv.php
```diff
@@ -42,6 +42,7 @@ public static function form_define_fields(): array {
         return [
             'path' => ['type' => PARAM_TEXT, 'required' => true],
             'headers' => ['type' => PARAM_TEXT],
+            'overwriteheaders' => ['type' => PARAM_TEXT],
```
Peterburnett, line 45: This is a checkbox, should it be PARAM_BOOL?

classes/local/step/reader_csv.php
```diff
@@ -62,6 +63,7 @@ public function csv_contents_generator() {
         $maxlinelength = 1000;
         $config = $this->get_variables()->get('config');
         $strheaders = $config->headers;
+        $overwriteheaders = !empty($config->overwriteheaders);
```
Peterburnett, line 66: If swapped to PARAM bool, can we just treat that as a bool? rather than empty()

keevan: The field might not exist yet. This just supports both and every other scenario.

Peterburnett (approved):

PR #810, opened 2023-07-18
[#809] fix: SQL input field width to improve UX for larger queries

### LINKED ISSUE #809: SQL textareas are too small
The width should fill the whole area (width: 100%) since nothing else is used in that space, and the height should be bigger so larger queries can breathe more easily. 

For smaller queries, I believe the extra space is a small price to pay for the enhanced experience of managing larger queries.

Perhaps a nice balance might be 20 lines for reader_sql and 10 lines for flow_sql. Ideally they'd be the same size for consistency.

Here's a preview of what it might look like
image

An example with a slightly longer query (courtesy of Bard)
image

You can see once we start getting into more realistic / advanced queries, it can quickly get much longer.

keevan:
Resolves #809

Before:
image

After:

image

### DIFF
--- a/classes/local/step/base_step.php
+++ b/classes/local/step/base_step.php
@@ -299,6 +299,18 @@ abstract class base_step {
     }
 
     /**
+     * Allow steps to setup the form depending on current values.
+     *
+     * This method is called after definition(), data submission and set_data().
+     * All form setup that is dependent on form values should go in here.
+     *
+     * @param \MoodleQuickForm $mform
+     * @param \stdClass $data
+     */
+    public function form_definition_after_data(\MoodleQuickForm &$mform, \stdClass $data) {
+    }
+
+    /**
      * Sets the default types for each input field defined
      *
      * This is mostly defined upfront and unlikely to change, and will show the following if not set:
--- a/classes/form/step_form.php
+++ b/classes/form/step_form.php
@@ -140,6 +140,24 @@ class step_form extends \core\form\persistent {
     }
 
     /**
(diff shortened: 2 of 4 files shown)


keevan:
I tried using a class (w-100) instead of adding to the styles, but I don't think I'm applying it properly.

e.g. these didn't work:
```
        $mform->addElement('textarea', 'config_sql', get_string('reader_sql:sql', 'tool_dataflows'),
            ['class' => 'w-100', 'cols' => 50, 'rows' => 7, 'style' => 'font: 87.5% monospace;']);
```

```
        $mform->addElement('textarea', 'config_sql', get_string('reader_sql:sql', 'tool_dataflows'),
            ['class' => 'w-100', 'rows' => 7, 'style' => 'font: 87.5% monospace;']);
```

classes/local/step/base_step.php
```diff
@@ -298,6 +298,18 @@ public function form_setup(\MoodleQuickForm &$mform) {
...
+    /**
+     * Allow steps to setup the form depending on current values.
+     *
+     * This method is called after definition(), data submission and set_data().
+     * All form setup that is dependent on form values should go in here.
+     *
+     * @param \MoodleQuickForm $mform
+     * @param \stdClass $data
+     */
+    public function form_definition_after_data(\MoodleQuickForm &$mform, \stdClass $data) {
```
Peterburnett, line 310: This makes me wonder whether this should be an abstract class instead. This is more an idle thought, no changes required.

Peterburnett (approved):

PR #812, opened 2023-07-21
add psr3 logging

### LINKED ISSUE #681: Aborted flows display the wrong "finish time"
If a flow gets aborted the finish time displays in the UI as `1 January 1970, 10:00:00 AM` probably because it is unix time 0.

Ideally, this should show the time it was aborted, or nothing at all.

keevan:
- feat: add log to file in dataflow specific run format
- build: add symfony/monolog-bundle
- build: composer update to update packages
- feat: implement basic psr3 logging, replacing the current implementation
- add 3rd party libs to the XML
- chore: remove unused logging_context class
- refactor: calls from log(message) to log->info(message)
- fix: issue when run isn't created due to it bring a dry run
- fix: add engine step context (step name) to the log
- fix: logging structure to be more compact
- improve logging messages inside step exec
- fix: update logging for switch and variable steps
- fix: update logging such that the individual step context is added automatically
- fix: calling abort multiple times won't cause an exception
- feat: add admin level settings to configure the log handlers enabled for all flows
- feat: add dataflow level settings to configure log handlers

Resolves most of https://github.com/catalyst/moodle-tool_dataflows/issues/381

Only missing the ability to adjust log level view from UI (aside from in browser console / dev tools)

Resolves #681


classes/local/execution/engine.php
```diff
@@ -506,21 +521,30 @@ public function finalise() {
...
         $message = '';
         if (isset($reason)) {
             $message .= $reason->getMessage();
             if (isset($reason->debuginfo)) {
                 $message .= PHP_EOL . $reason->debuginfo;
             }
         }
-        $this->log('Aborted: ' . $message);
+
+        $this->set_current_step(null);
```
Peterburnett, line 537: This appears to be handled by the caller as well. Worth making it a single responsibility point?

classes/local/execution/engine.php
```diff
@@ -676,4 +706,99 @@ public function resolve_path(string $pathname) {
...
+
+        // Each channel represents a specific way of writing log information.
+        $log = new Logger($channel);
+
+        // Add a custom formatter.
+        $log->pushProcessor(new PsrLogMessageProcessor(null, true));
+
+        // Ensure step names are used if supplied.
+        $log->pushProcessor(function ($record) {
+            if (isset($this->currentstep)) {
```
Peterburnett, line 743: I find the logic of this callback a little hard to follow, I think it can be slightly simplified
```
           if (isset($this->currentstep)) {
               $record['context']['step'] = $this->currentstep->name;
               $record['message'] = '{step}: ' . $record['message'];
            }
            if (isset($record['context']['record'])) {
                $record['message'] .= ' {record}';
                $record['context']['record'] = json_encode($record['context']['record']);
            }
```

Peterburnett: Also is the '{step}' string processing as a literal intended? (If im reading this right)

classes/local/execution/engine.php
```diff
@@ -676,4 +706,99 @@ public function resolve_path(string $pathname) {
...
+        }
+
+        // Default Moodle handler. Always on.
+        $mtracehandler = new mtrace_handler(Logger::DEBUG);
+        $mtracehandler->setFormatter($lineformatter);
+        $log->pushHandler($mtracehandler);
+
+        // Log to the browser's dev console for a manual run.
+        if (isset($loghandlers[log_handler::BROWSER_CONSOLE])) {
+            $log->pushHandler(new BrowserConsoleHandler(Logger::DEBUG));
```
Peterburnett, line 774: The BrowserconsoleHandler doesnt get the same treatment as the other 2 handlers (microtime formatting). Any reason it cant be consistent between all handers?

classes/local/execution/iterators/dataflow_iterator.php
```diff
@@ -222,9 +223,12 @@ public function next($caller) {
             // Expose the number of times this step has been iterated over.
             $this->stepvars->set('iterations', $this->iterationcount);
 
-            $this->step->log('Iteration ' . $this->iterationcount . ': ' . json_encode($newvalue));
+            // Log the iteration for real steps.
+            if (!$this->steptype instanceof flow_cap) {
```
Peterburnett, line 227: I might be mixing up my contexts but isnt the flow cap check handled in the log method on the engine step? I think this check might be redundant

Peterburnett (commented):
Looking largely great. Its really hard to follow the different levels of 'log()' in general, as its somewhat nested in internals, but what I read *does* make sense, and I think that im largely missing some wider context when working with it.

classes/local/execution/engine.php
keevan: Do you have any particular suggestions to simplifying it? The record one might no longer be needed, as the logs have since changed to print out the context as well. 

> Also is the '{step}' string processing as a literal intended? (If im reading this right)

Keen observation. It's levaraging the PSR log class, where the "step" value replaces any instances of "{step}" in the message. It's pretty neat. And it only applies when the  "step" value is supplied

classes/local/execution/engine.php
keevan: I believe because to get it working, it needs to work with it's internal message formatter. The other ones direct the logs to a stream, or a rotating file or to mtrace. To keep things consistent there, the log message format is utilised.

I can definitely try and apply the formatter, but I feel like you'd see the console.log time in addition to the message, in addition to the context, without it being a nice JS object you can pull out and play with.

classes/local/execution/iterators/dataflow_iterator.php
keevan: That's fair. It was re-added because I converted all the checks to `->log->info()` calls and you're correct. If I changed this back, it would work without the additional check.

classes/local/execution/engine.php
keevan: image

Doesn't seem to be (when I remove this), so there might be a place where this isn't applied cleanly. There might be a nicer way to go about this in general, but calling set_current_step(null) from within the context of a step feels weird.

keevan: Hint: `Abort if failed: ` is the name of a step, and `Engine: ` should be the start of the message.

classes/local/execution/engine.php
Peterburnett: Gotcha, second layer of string interpolation. Neat

classes/local/execution/engine.php
Peterburnett: No thats fine, just something I spotted that was different to the other 2. Happy to leave as is

keevan: I tested this in the browser. I think it's OK leaving it with just Iteration 1, iteration 2, etc.

You can easily filter down to specific content as required, which would do more or less the same thing as visual inspection.

I think the debugging experience is OK, and is probably better to have less noise by default until you drill down to a specific example, there is probably some degree of browser performance benefit there by not including the context data by default (in this case, the record)

Peterburnett (approved):
LGTM

PR #815, opened 2023-07-25
fix: directory file list step to properly handle glob returns

keevan:


keevan:
Small update to PR, the filter were applied but the filtered list wasn't being returned :zany_face: .. which is a bug

Also some surrounding doc updates

Peterburnett (approved):

PR #817, opened 2023-07-25
feat: named capture group support for regex transformer step

### LINKED ISSUE #816: Add support for named capture groups on regex transform step
Currently it sets the first capture on `record.<stepalias>`, which is okay but limiting. Potentially, you don't want a step alias to match a flowing record field, or you may want to reuse it later on in a different flow step.

I propose we add support for named capture groups, which gives us the flexibility to declare all the matches we expect, and we can use the names to set the relevant keys. Win-win?

keevan:
Resolves #816


Peterburnett (approved):
This is really nice, makes it much easier to capture slippery data.

PR #827, opened 2023-08-11
close #826 - fix: broken reader_sftp_directory_file_list

### LINKED ISSUE #826: Missing SFTP directory file list reader
The code is there, but doesn't get picked up due to incorrect file/class name.

dmitriim:


keevan:
Changes look good. Could you also update the commit message?

dmitriim:
@keevan Any particular suggestion for commit message? I was trying to follow the pattern of the existing commits.

keevan:
> @keevan Any particular suggestion for commit message? I was trying to follow the pattern of the existing commits.

You can change `fix: fix broken reader_sftp_directory_file_list` to `fix: broken reader_sftp_directory_file_list` if you'd like. Mainly getting rid of the extra/repeated word.

dmitriim:
Oh. I thought the pattern here is like: <type of commit>: <commit message>.

keevan (approved):
LGTM

keevan:
> Oh. I thought the pattern here is like: [type of commit]: [commit message]. Would be good to document it somewhere all edge cases of this to save people some time.

The convention I'm using is https://www.conventionalcommits.org/en/v1.0.0-beta.4/

Regardless of convention, I think its odd if it reads "fix fix" and sounds a bit redundant.

dmitriim:
> I think its odd if it reads "fix fix"

That's probably just you. 

I think the best suggestion would be to replace the second fix by another similar verb. But anyway...

PR #830, opened 2023-08-24
fix: loghandlers now optional and saving an existing flow should work

### LINKED ISSUE #829: Persistent errors when saving an existing or importing a dataflow without log handlers
Log handlers should be optional, yet they are causing an error to be shown when a value is not provided (e.g. null)

keevan:
Resolves #829

Before:

image

After:

image


Peterburnett (approved):
This is just updating the persistent def to match the schema. LGTM

PR #831, opened 2023-08-25
[#628] Add compression connector

### LINKED ISSUE #628: Compression/decompression steps
Steps to compress and decompress.
- Select algorithm from a list (zip, gz, bz2, 7z etc).
- Read/write either files or variables. (Perhaps only variables and use in conjunction with #627)
- Connector and flow variants.

matthewhilton:
- Adds flow/connector steps for compression/decompression
- Currently only `gzip` is implemented, but it is structured in a way you can in the future easily implement other executables or functions to zip/unzip
- Covered by unit tests

image

Closes #628


classes/local/step/compression_trait.php
```diff
@@ -0,0 +1,267 @@
...
+     *
+     * @return  array of outputs
+     */
+    public function define_outputs(): array {
+        return ['success' => get_string('compression:output_success', 'tool_dataflows')];
+    }
+
+    /**
+     * Returns the method info that has been selected in the configuration.
+     * @param string $method method name
```
keevan, line 166: It would appear more consistent to space the doc summary and parameters.

```suggestion
     * Returns the method info that has been selected in the configuration.
     *
     * @param string $method method name
```

This would apply to other similar places in this PR too.

matthewhilton: Good catch - have updated these to be consistent :+1:

classes/local/step/compression_trait.php
```diff
@@ -0,0 +1,270 @@
...
+        $movefilename = $config->command == 'compress' ? $config->from . '.gz' : rtrim($config->from, '.gz');
+        $movefilename = escapeshellarg($movefilename);
+
+        // See https://www.gnu.org/software/gzip/manual/html_node/Invoking-gzip.html.
+        // -f: force override destination file if it exists
+        // -v: verbose
+        // -k: keep input file
+        // 2>&1: pipe stderror to stdout.
+        $gzipcommand = "{$gzip} -f -v -k {$compressionmode} {$from} 2>&1 && mv {$movefilename} {$to}";
+        $this->enginestep->log("Command: " . $gzipcommand);
```
keevan, line 250: I believe logging should now be in this convention, which allows you set specify the logging level per message. 

`$this->log->info(message);`

keevan: https://www.php-fig.org/psr/psr-3/

keevan: For example, logging at a NOTICE level.
https://github.com/catalyst/moodle-tool_dataflows/blob/dce86202b4abe4c3973b93f7a4889ecbf780d6c6/classes/local/step/directory_file_list_trait.php#L148

matthewhilton: Nice! Didn't know dataflows had different logging levels

I have switched the logging levels in 7a006d41dcd4ba6752851f00c96e62534696875e

I also removed the log in the execute_gzip, since it is outputted anyway in `execute` so no need to double up.

image

keevan (approved):
Looks great. Thanks @matthewhilton

Issue #832, opened 2023-08-31
SFTP step private pub key auth not working


keevan:
https://github.com/catalyst/moodle-tool_dataflows/blob/4c83e2e1c6ec8152f1a73e855be7cc045b5512fb/classes/local/step/sftp_trait.php#L300

There might have been some changes made since. $config here hasn't been provided anywhere, so it will never enter the block and thus private/pub key auth won't work.

PR #833, opened 2023-08-31
[#832] fix the issue when private key was never used

dmitriim:


classes/local/step/sftp_trait.php
```diff
@@ -293,13 +293,22 @@ public function execute($input = null) {
...
      */
     private function load_key() {
-        // Because variables use redacted values by default, we evaluate the secret explicitly.
-        $password = $this->get_variables()->evaluate($this->stepdef->config->password);
+        $password = false;
+        $config = $this->stepdef->get('config');
+
+        if (empty($config)) {
+            return $password;
+        }
```
keevan, line 301: Is config itself ever empty for this step? I think it's not possible given it's validation.

classes/local/step/sftp_trait.php
```diff
@@ -293,13 +293,22 @@ public function execute($input = null) {
...
+
+        if (!empty($config->password)) {
+            // Because variables use redacted values by default, we evaluate the secret explicitly.
+            $password = $this->get_variables()->evaluate($config->password);
+        }
```
keevan, line 306: I believe this would miss cases where the password is an empty string or zero(es). It might be too edgy of a case to care though, and it's probably something that shouldn't be set so I'm okay with this.

classes/local/step/sftp_trait.php
```diff
@@ -293,13 +293,22 @@ public function execute($input = null) {
...
 
         // Use key authorisation if privkeyfile is set.
         if (!empty($config->privkeyfile)) {
             $privkeycontents = file_get_contents($this->resolve_path($config->privkeyfile));
-            $privkeypassphrase = $config->password ?: false;
+            $privkeypassphrase = $password ?: false;
```
keevan, line 311: By this stage, password would have already been set. It's safe to use the password variable as-is and you do not have to assign it to a new variable (or default it back to false)

classes/local/step/sftp_trait.php
dmitriim: well, it's a trait so we can't be 100% how it could be used... for me it makes sense to add this check.

classes/local/step/sftp_trait.php
```diff
@@ -293,13 +293,22 @@ public function execute($input = null) {
      * @return AsymmetricKey|string
      */
     private function load_key() {
-        // Because variables use redacted values by default, we evaluate the secret explicitly.
-        $password = $this->get_variables()->evaluate($this->stepdef->config->password);
+        $password = false;
```
keevan, line 296: I'm also concerned whether or not the current user:pass and user:nopassword will work, if false is returned. If it's OK, that's great. We should update the phpdoc to support `AsymmetricKey|string|false`

classes/local/step/sftp_trait.php
dmitriim: good point. Changing to isset

classes/local/step/sftp_trait.php
dmitriim: This is annoying, but it's not very clear what will be returned by `$this->get_variables()->evaluate($config->password); `Ideally we should add more validation here to make sure it's a string.

dmitriim:
All those good points that needs to be sorted. For now we moving on with a minimum changes https://github.com/catalyst/moodle-tool_dataflows/pull/834/files

PR #834, opened 2023-08-31
[#832] fix the issue when private key was never used

dmitriim:


keevan (commented):
This seems OK. Can you also update the `$privkeypassphrase = $config->password ?: false` line to use `$password` instead of `$config->password` since that seems obviously wrong as well. 

If used as-is with a private key and a passphrase, it would load the redacted **** as the passphrase and it would not work, which is the reason the evaluate call was required.

Otherwise it looks good as-is.

dmitriim:
Done and force pushed.

keevan (approved):

PR #838, opened 2023-10-04
add log step type

### LINKED ISSUE #836: Add a generic "log" step which logs a custom templated message
There are some cases where we want human readable log messages when certain things happen.

If possible, this should be output as part of the relevant step, but sometimes, it needs to be a summary or a contextual message for the flow as opposed to one specific to the step. 

For example, if a certain type of validation failed, there is no way to see what failed, what it was expected to be, and how the user can action it. At most, the dataflow would be able to show where it failed validation (e.g. a switch case was entered), but some additional steps need to be taken to figure out what needs to be done about it.

Another way to think about it, would that it serves as a means to "log a message" instead of emailing the relevant users. This might work in cases where accumulation of errors/warnings are done via the logging mechanism, and a notification (e.g. email) is sent afterwards with a count/summary and further action can be done after viewing the logs).

Bonus reason for this would be to piggyback on the logging structure, and if we wanted to output the logs to some remote store, or email them after each run, that can be added as an additional log handler instead of creating another step type for it.

keevan:
- feat: add log step type
- fix: abort flow
- fix: current vars will now only log if it contains values
- feat: ensure iterations is always available for flow steps
- fix: log levels for some messages (sftp, engine abort, iterations)
- feat: add more styling to output messages for manual runs, based on log levels

Resolves #836


classes/local/execution/engine.php
```diff
@@ -535,14 +535,21 @@ public function abort(?\Throwable $reason = null) {
...
 
         $this->set_current_step(null);
-        $this->log('Engine: aborting steps');
+        $this->log('Engine: aborting steps', $message ? ['reason' => $message] : [], Logger::NOTICE);
         $this->exception = $reason;
         foreach ($this->enginesteps as $enginestep) {
-            $this->set_current_step($enginestep);
-            $enginestep->abort();
+            $status = $enginestep->status;
+            if ($status !== self::STATUS_FINISHED && !in_array($status, self::STATUS_TERMINATORS)) {
```
Peterburnett, line 542: This check is used a fair bit, might be worth extracting to an is_terminated static method. Might not be worth, but might be a bit more readable/writeable

classes/local/step/log_trait.php
```diff
@@ -0,0 +1,106 @@
...
+    ];
+
+    /**
+     * Returns whether the step has a side effect.
+     *
+     * @return     bool whether or not this step has a side effect
+     * @link https://en.wikipedia.org/wiki/Side_effect_(computer_science)
+     */
+    public function has_side_effect(): bool {
+        return false;
```
Peterburnett, line 55: This should likely be true, my understanding is a  file WILL be generated during a dry run which may not be intended

keevan: Logs will always be created somewhere, somehow (even without this step type). Using that definition of a side effect means every step has a side effect since they all log something, and makes this function redundant. 

Perhaps the definition of a side effect in terms of dataflow needs to be clarified. From what I recall, it is described as "any state that changes behaviour", but there was some murky cases where creating a local / temp file in request_dir was considered not being a side effect, until it moves into a main file location or if it is created in an external system. A post request would typically create a resource, and therefore has a side effect, a GET request is typically idempotent and would not normally have a side effect (though external systems may still "log" the request taking place and thus create some data somewhere).

Should this move to a new issue, to clarify in the README somewhere as a single reference point?

classes/local/execution/engine.php
keevan: I had thought about that also. It's currently used in 2 places, within the same main function. The definition of a terminator seems to be different than a step which is in a finished state, which is why the finished was added as a separate check, as opposed to adding finished to the list of status terminators. One example I can think of is that steps in a finished state, will not "release" the resource until they transition to finalised (e.g. on_finalise)

In terms of keeping things pretty DRY, I'm all for it, though I prefer approaching it with more pragmatism and avoid abstracting too early. I personally follow the rule of 3.
https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

keevan: Perhaps an alternative would be to replace the check with something more readable such as this:
```suggestion
            $isdone = status !== self::STATUS_FINISHED && !in_array($status, self::STATUS_TERMINATORS);
            if ($isdone) {
```

Peterburnett: Yeah this is a nice middle ground.

classes/local/step/log_trait.php
Peterburnett: This is a great justification, and explains your position well. Logging is certainly the lowest risk operation that could be classed as a side effect, andl I am fine with leaving it as is in this case. I think probably a new issue is all we need to do here, just to clarify for future us that its not truly side effect free to do a dry run, as logs might still be generated

Peterburnett (approved):

PR #846, opened 2023-10-25
Moodle 401 stable 839 fix

Peterburnett:


classes/local/execution/logging/mtrace_handler.php
```diff
@@ -26,17 +27,29 @@
  * @copyright Catalyst IT, 2023
  * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
-class mtrace_handler extends AbstractHandler {
+class mtrace_handler extends AbstractProcessingHandler {
 
     /**
      * Default handler for Moodle.
      *
      * @param array $record the log record
```
Peterburnett, line 35: The comment should be updated to contain the return

gbarat87: Return added

PR #847, opened 2023-11-09
add user step and find step

keevan:
- feat: add find step - returns a match from a list
- feat: add update user step


classes/local/step/update_user_trait.php
```diff
@@ -0,0 +1,89 @@
...
+            'textarea',
+            'config_fields',
+            get_string('update_user:fields', 'tool_dataflows'),
+            ['cols' => 60, 'rows' => 5],
+        );
+        $mform->addElement('static', 'config_fields_help', '', get_string('update_user:fields_help', 'tool_dataflows'));
+    }
+
+    /**
+     * Executes the step and copies what is in $from, to the $to path
```
Peterburnett, line 70: Comment outdated

classes/local/step/update_user_trait.php
```diff
@@ -0,0 +1,89 @@
...
+     *
+     * @param  mixed|null $input
+     * @return mixed
+     */
+    public function execute($input = null) {
+        $variables = $this->get_variables();
+        $config = $variables->get('config');
+        $userobject = (object) array_merge(
```
Peterburnett, line 78: Can we prevent a couple fields from being changed here? Specifically auth and password (should already be covered by the false param to update but still)

Peterburnett (approved):

PR #849, opened 2023-11-28
Update file glob copy to work on single files

Peterburnett:
After:

image

### DIFF
--- a/classes/local/step/copy_file_trait.php
+++ b/classes/local/step/copy_file_trait.php
@@ -76,22 +76,26 @@ trait copy_file_trait {
         }
 
         // Attempt to copy the file to the destination.
-        // If $to is not a directory, then it should not glob anything and copy as-is.
-        if (!is_dir($to)) {
+        // If $to is not a directory and $from exists as a file, then it should not glob anything and copy as-is.
+        if (!is_dir($to) && file_exists($from)) {
             $this->copy($from, $to);
             return $input;
         }
 
-        // Otherwise, it is probably multiple files, and should be globbed.
+        // Otherwise, it is probably multiple files or a file to be renamed, and should be globbed.
         $files = glob($from);
         if (empty($files)) {
             return $input;
         }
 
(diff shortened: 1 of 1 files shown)


classes/local/step/copy_file_trait.php
```diff
@@ -76,23 +76,33 @@ public function execute($input = null) {
...
+                    $dest = realpath($to . DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR . basename($file);
+                    $this->copy($file, $dest);
+                }
+            }
+        } else {
+            // Here we are trying to copy a glob to a specified target filename.
+            // Only move the first glob.
+            $file = reset($files);
+            if (!empty($file)) {
+                $this->copy($file, $to);
```
keevan, line 105: I think it's simpler to keep all the handling the same. (e.g. the original handling does the same thing)

If there was a specific need for a glob to only process the first N results, that should be structured differently in the flow, and we shouldn't assume it here.

What do you think?

keevan: Ah, one other thing I did notice. And I think we should fix this.

If we set a destination as `something/`, it should treat it like a directory, and use the name of the original file. Otherwise, if it's not already a directory, it should use it as the target filename (which matches the linux cp/mv commands).

Peterburnett: I think I disagree. We already swap behaviour based on [dataroot]/dir/ VS [dataroot]/dir/target.file, and I think we should continue with that pattern. The problem with the foreach loop here is it constructs the final path based on the assumption that it IS a dir target. if its not, we should always use the absolute path.

As for the looping change, this is just to prevent confusion. EG if there were 2 files that matched, I think the most correct behaviour is just to copy the first match, rather than looping through, which would effectively result in only copying the LAST match, which seems counter intuitive

keevan (approved):
Looks okay to me, thanks Peter.

PR #852, opened 2023-11-30
[#851] Update composer libs

### LINKED ISSUE #851: Test timeouts on PHP 8.1
I've got a strong suspicion that test timeouts we've been seeing on our CI server is a regression of https://github.com/catalyst/moodle-tool_dataflows/pull/846

You can see even the Github CI actions for any php 8.1 simply dies

matthewhilton:
Closes #851 

The issue seemed to be the old version of symfony was spamming a ton of debug output messages, and then eventually got stuck in what seemed an infinite loop.

Changes:

- Updated symfony/yaml from 3.4 -> 4.0.
    - 3.4 is the last 3.* version according to https://packagist.org/packages/symfony/yaml#v4.0.0
    - So I had to go to the next version, with is 4.0
    - 4.0 requires php 7.1 or above, so we should be good to make this a dataflows-wide change assuming tests pass seeing as we target php 7.4 and above


keevan (commented):
Hi Matt, it looks mostly good. I'm concerned about all the extra base composer file changes, and would like to keep it somewhat consistent. 

Could you please check the changes here https://github.com/catalyst/moodle-tool_dataflows/compare/851-update-dependencies...851-update-dependencies-extra-changes and see if you can cherry pick them into this PR? 

All I need was update composer (composer self-update) and run the full install command in **readme_moodle.txt**.

matthewhilton:
Good point - looks like my composer is on an older version as well. I just updated my composer to `2.6.6` and ran the `composer update` command which updated a few packages. Will wait and see if the CI passes now or if it still gets stuck

keevan (approved):
Looks good, thanks Matt

PR #855, opened 2023-12-21
additional goodies csv sftp perf sql and more

keevan:
- feat: add connector SQL step
- fix: allow values to be corrected in bad states
- feat: add set multiple variables step type
- fix: warning logs now display and are more legible
- fix: add copy file error handling for directory creation
- feat: run name and id are now available as expressable vars
- fix: SQL trait handling of integer values and improve error message
- fix: update email flow step to support it as a final step
- fix: reader csv error to be more correct for header/fields count alignment
- fix: include row number in csv reader error
- perf: cache sftp object per identity to speed up subsequent actions
- feat: add an option to continue processing the csv file on known parse issues

For the SFTP one in particular, it was doing about 200ms initial connection time, 600-700ms initial upload time, but otherwise breezing for subsequent SFTP actions, unlike previously.

### DIFF
--- a/classes/form/step_form.php
+++ b/classes/form/step_form.php
@@ -107,11 +107,22 @@ class step_form extends \core\form\persistent {
         $select->setMultiple(true);
 
         $dataflow = new dataflow($dataflowid);
-        $varroot = $dataflow->get_variables_root();
 
-        // List all the available fields available for configuration, in dot syntax.
-        $mform->addElement('static', 'fields', get_string('available_fields', 'tool_dataflows'),
-            $this->prepare_available_fields($varroot->get()));
+        try {
+            $varroot = $dataflow->get_variables_root();
+
+            // List all the available fields available for configuration, in dot syntax.
+            $mform->addElement(
+                'static',
+                'fields',
+                get_string('available_fields', 'tool_dataflows'),
+                $this->prepare_available_fields($varroot->get())
+            );
+        } catch (\Throwable $e) {
+            global $OUTPUT;
+            $errtext  = $OUTPUT->notification($e->getMessage());
+            $mform->addElement('static', 'vars_error', get_string('available_fields', 'tool_dataflows'), $errtext);
+        }
 
         // Check and set custom form inputs if required. Defaulting to a
         // textarea config input for those not yet configured.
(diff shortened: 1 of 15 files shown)


classes/form/step_form.php
```diff
@@ -107,11 +107,22 @@ public function definition() {
...
+            // List all the available fields available for configuration, in dot syntax.
+            $mform->addElement(
+                'static',
+                'fields',
+                get_string('available_fields', 'tool_dataflows'),
+                $this->prepare_available_fields($varroot->get())
+            );
+        } catch (\Throwable $e) {
+            global $OUTPUT;
+            $errtext  = $OUTPUT->notification($e->getMessage());
```
Peterburnett, line 123: Could this ever leak anything sensitive?

classes/local/step/sftp_trait.php
```diff
@@ -287,6 +290,17 @@ public function execute($input = null) {
         return $input;
     }
 
+    /**
+     * Hook function that gets called when an engine step has been finalised.
+     */
+    public function on_finalise() {
```
Peterburnett, line 296: is that also caught when aborting ?

Peterburnett: Also, in the case a fatal is thrown from further down? Its probably fine, but just in case

classes/form/step_form.php
keevan: Was there an example of anything you had in mind? 

This change is added so that the error shown to the user, doesn't block out the whole page, so the user can fix the issue on the form below. This typically happens from bad data due to faulty expressions.

Of the exceptions I saw myself, nothing of a sensitive nature was ever part of the error message.

Peterburnett: Nah thats likely fine. Im just wary of directly outputting system stuff, especially as we have a bunch of secrets and vars etc in dataflows. But its admin restricted, and I think this really is fine.

classes/local/step/sftp_trait.php
keevan: Good question. I'll check and come back to you. It might also need to be set in an on_abort() method as well.

I wonder if there's a better more standard way to use and free up resources.

keevan: For now I'll add it to the on_abort callback too.

Peterburnett (approved):
Approved to merge once CI green

PR #857, opened 2023-12-22
improvements from 35

keevan:
- feat: add connector SQL step
- fix: allow values to be corrected in bad states
- feat: add set multiple variables step type
- fix: warning logs now display and are more legible
- fix: add copy file error handling for directory creation
- feat: run name and id are now available as expressable vars
- fix: SQL trait handling of integer values and improve error message
- fix: update email flow step to support it as a final step
- fix: reader csv error to be more correct for header/fields count alignment
- fix: include row number in csv reader error
- perf: cache sftp object per identity to speed up subsequent actions
- feat: add an option to continue processing the csv file on known parse issues
- build: bump version


Peterburnett (approved):
Approved to merge once CI green

PR #868, opened 2024-02-27
fix: dataflow table view

### LINKED ISSUE #867: Exception abs() arg 1 must be int or float but was given a string
```
Exception - abs(): Argument #1 ($num) must be of type int|float, string given

More information about this error
Debug info:
Error code: generalexceptionmessage
Stack trace:

    line 2231 of /lib/moodlelib.php: TypeError thrown
    line 2231 of /lib/moodlelib.php: call to abs()
    line 206 of /admin/tool/dataflows/classes/dataflows_table.php: call to format_time()
    line 842 of /lib/tablelib.php: call to tool_dataflows\dataflows_table->col_lastrunduration()
    line 2059 of /lib/tablelib.php: call to flexible_table->format_row()
    line 63 of /admin/tool/dataflows/classes/sql_table.php: call to table_sql->build_table()
    line 194 of /admin/tool/dataflows/classes/visualiser.php: call to tool_dataflows\sql_table->build()
    line 68 of /admin/tool/dataflows/index.php: call to tool_dataflows\visualiser::display_dataflows_table()
```

keevan:
Resolves #867

### DIFF
--- a/classes/dataflows_table.php
+++ b/classes/dataflows_table.php
@@ -203,9 +203,9 @@ class dataflows_table extends sql_table {
                     . get_string('runningfor', 'tool_dataflows', format_time($delta, 0)) . '</span>';
             }
             if ($delta > 10) {
-                return format_time(number_format($delta, 0));
+                return format_time($delta);
             }
-            return format_time(number_format($delta, 1));
+            return format_time((float) number_format($delta, 1));
         }
         return '';
     }


classes/dataflows_table.php
```diff
@@ -203,9 +203,9 @@ public function col_lastrunduration(\stdClass $record): string {
                     . get_string('runningfor', 'tool_dataflows', format_time($delta, 0)) . '</span>';
             }
             if ($delta > 10) {
-                return format_time(number_format($delta, 0));
+                return format_time((float) number_format($delta, 0));
```
matthewhilton, line 206: This seems a bit weird - number_format is turning $delta (an integer) into a string with commas and decimal points and stuff
But then it is converted back to a float, to run format_time?

Why not just pass the $delta straight to format_time?

It might be because we want number_format to strip the decimal points off it. But in that case, you could just use abs() instead.

keevan: Your analysis is correct.

matthewhilton (approved):

PR #875, opened 2024-03-02
feat: Add notification pathway for aborted flows

Peterburnett:


keevan (approved):
Code looks good. 

Happy for this to be merged in. I checked the email functionality and it is working as expected. 

One small observation which may be useful. Should the email subject contain the name of the dataflow (in addition to the abort message)? 

This may help address cases where many dataflows are active, and would make identifying the problematic one quicker to e.g. to filter on, instead of filtering on the email body.

An example of this would be changing this:
`Dataflow run aborted`

to something like this:
`Dataflow run aborted: SMS import and sync`

Peterburnett:
This is a good idea, I'll add this

Peterburnett:
Updated. Ill merge once green

PR #876, opened 2024-03-03
feat: Smart retries with delay + retry count for cron trigger

Peterburnett:


keevan (commented):
Thanks Peter. Code looks good. Some small comments. Thanks also for adding the unit tests :heart:, and the config helper is pretty neat. I'm sure it will be used more when things change.

keevan (approved):

PR #882, opened 2024-05-31
feat: allow multiple emails in Notify on abort to fix #881

### LINKED ISSUE #881: Allow multiple emails in Notify on abort
Would be nice to be able to provide few emails separated by comma.

dmitriim:


matthewhilton:
Closes #882

matthewhilton (approved):
Looks good!

Would be good to see unit tests but since there weren't any for the single email function before I won't let that be a blocker.

Gave it a quick test on my local and worked as expected.

PR #886, opened 2024-07-05
Allow auth and password mutation in Update user step

golenkovm:
There is a need to change auth method to SAML2 and clear the original password


golenkovm:
It looks like this could be implemented via sql flow (simply call an update sql). Closing the PR

PR #888, opened 2024-07-25
make the eventlist compatible with Moodle 4.3

danielneis:
The way the events list is composed has changed in 4.3 (or maybe in 4.2 but I have not tested).
This change makes the events list work again.


danielneis:
I've rebased on top of current MOODLE_401_STABLE

dmitriim:
@danielneis thank you! This has been applied to MOODLE_405_STABLE branch of the plugin

jaypha:
The 401 branch needs to support both 4.1 and 4.3 functionality.
Issue #954 has been created for this.

PR #894, opened 2024-09-17
Add config to publish to plugin directory

djarran:
**Description:** This PR adds a secret to the GitHub Actions workflow and bumps the plugin version to enable publishing to the Moodle Plugin Directory.

The plugin is passing all tests as seen in previous commits.


danmarsden:
We need to get someone to create an entry in the moodle.org plugins database before this is merged otherwise it will just fail because it doesn't exist in the plugins db yet.

PR #904, opened 2024-10-10
Enhance logging details and verbosity

keevan:
### Verbosity / min log level setting (per dataflow)

image

### Log comparison:

Before:
`[2024-10-10 11:08:57.219433] dataflow/12.WARNING: Log: Example log entry 1 [] [] `
After:
`[2024-10-10 11:08:57.219433] dataflow/12.WARNING: Log: Example log entry 1 {"first":"1","second":"2","third":"3"} [] `

Example test dataflow:
[log_testing_20241010_0254.yml.txt](https://github.com/user-attachments/files/17321158/log_testing_20241010_0254.yml.txt)

### DIFF
--- a/classes/local/execution/engine.php
+++ b/classes/local/execution/engine.php
@@ -679,8 +679,8 @@ class engine {
         $this->get_variables()->set("states.$statusstring", microtime(true));
 
         $context = [
-                'isdryrun' => $this->isdryrun,
-                'status' => get_string('engine_status:'.self::STATUS_LABELS[$this->status], 'tool_dataflows'),
+            'isdryrun' => $this->isdryrun,
+            'status' => get_string('engine_status:'.self::STATUS_LABELS[$this->status], 'tool_dataflows'),
         ];
 
         $level = Logger::INFO;
@@ -865,7 +865,7 @@ class engine {
     /**
      * Send a notification email for abort if required.
      *
-     * @param \Throwable $reason A throwable representing the reason for abort.
+     * @param ?\Throwable $reason A throwable representing the reason for abort.
      */
     public function notify_on_abort(?\Throwable $reason) {
         // If configured to send email, attempt to notify of the abort reason.
--- a/classes/local/execution/engine.php
+++ b/classes/local/execution/engine.php
@@ -815,14 +815,17 @@ class engine {
             $loghandlers = $dataflowloghandlers;
         }
 
+        // Minimum logging levels - will display this level and above.
+        $minloglevel = $this->dataflow->get('minloglevel');
+
         // Default Moodle handler. Always on.
-        $mtracehandler = new mtrace_handler(Logger::DEBUG);
+        $mtracehandler = new mtrace_handler($minloglevel);
         $mtracehandler->setFormatter($lineformatter);
         $log->pushHandler($mtracehandler);
 
         // Log to the browser's dev console for a manual run.
         if (isset($loghandlers[log_handler::BROWSER_CONSOLE])) {
-            $log->pushHandler(new BrowserConsoleHandler(Logger::DEBUG));
+            $log->pushHandler(new BrowserConsoleHandler($minloglevel));
         }
 
         // Dataflow run logger.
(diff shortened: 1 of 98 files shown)


classes/local/execution/engine.php
```diff
@@ -223,6 +224,7 @@ public function __construct(dataflow $dataflow, bool $isdryrun = false, $automat
     /**
      * Destructor function.
      * Releases any locks that are still held.
+     * @return void
```
Peterburnett, line 227: Void returns are apparently just for returns with an explicit return statement.

https://moodledev.io/general/development/policies/codingstyle#phpdoc

classes/local/execution/engine.php
```diff
@@ -862,7 +880,8 @@ private function setup_logging() {
     /**
      * Send a notification email for abort if required.
      *
-     * @param \Throwable $reason A throwable representing the reason for abort.
+     * @param ?\Throwable $reason A throwable representing the reason for abort.
```
Peterburnett, line 883: You have \Throwable in the type, but also import it. Can probably just remove the use

db/install.xml
```diff
@@ -19,6 +19,7 @@
         <FIELD NAME="usermodified" TYPE="int" LENGTH="10" NOTNULL="false" SEQUENCE="false" COMMENT="Who last modified this record?"/>
         <FIELD NAME="confighash" TYPE="char" LENGTH="40" NOTNULL="false" SEQUENCE="false" COMMENT="The config hash of the most recent run"/>
         <FIELD NAME="notifyonabort" TYPE="char" LENGTH="255" NOTNULL="false" SEQUENCE="false" COMMENT="Email address to notify aborted runs to."/>
+        <FIELD NAME="minloglevel" TYPE="int" LENGTH="1" NOTNULL="false" DEFAULT="100" SEQUENCE="false"/>
```
Peterburnett, line 22: Length 1 Default 100 seems wrong right? length was the digit count I though. Unless its bytes?

tests/local/execution/tool_dataflows_engine_test.php
```diff
@@ -144,7 +144,7 @@ function ($s) {
      *
      * @return array
      */
-    public function dataflow_provider(): array {
+    public static function dataflow_provider(): array {
```
Peterburnett, line 147: Is this being called statically? Not sure where this comes from, just checking

tests/tool_dataflows_append_file_step_test.php
```diff
@@ -126,7 +126,7 @@ public function test_has_sideeffect(string $file, bool $expected) {
      *
      * @return array[]
      */
-    public function has_sideeffect_provider(): array {
+    public static function has_sideeffect_provider(): array {
```
Peterburnett, line 129: Is this a new requirement of PHPUnit to make data providers static?

Peterburnett:
There is a test failure as well about a non matching list of params.

Peterburnett (changes requested):
Test fail needs to be fixed

classes/local/execution/engine.php
keevan: Yeah, the strange thing is that those were flagged in the codechecker ci locally and was an autofix..

I'll check it out again and see if it doesn't need to be there. I normally don't include them.

db/install.xml
keevan: Good catch, it should be 2 since it can go up to 900. (It's bytes)

tests/local/execution/tool_dataflows_engine_test.php
keevan: Coming from local codechecker ci.

Don't have the output anymore but non-static providers are deprecated since they should be static.

keevan: It's a php test function dataprovider, so it's defined in the @dataProvider attribute of the php docs for the test.

tests/tool_dataflows_append_file_step_test.php
keevan: Yes, it's being deprecated, just staying ahead of the curve.

tests/local/execution/tool_dataflows_engine_test.php
keevan: Similar deprecation warning emitted for https://github.com/symfony/symfony/issues/48281

Unable to reproduce now but it was there at one point.

Peterburnett (approved):

PR #912, opened 2024-12-09
[#911] fix sql cleaning

### LINKED ISSUE #911: Cannot use `<=` in sql statements
It seems if you use a single `<=` in the sql reader's sql, it deletes the rest of the string. I think this is some weird interplay with HTML filtering.

E.g. if I input the sql into the sql reader step:
```
SELECT * FROM {user}
WHERE timecreated <= 5
```
and press save, it will actually save:
```
SELECT * FROM {user}
  WHERE timecreated 
```

matthewhilton:
Closes #911 

**Summary**

It was `clean_param` being called when `PARAM_TEXT` that mangles the `<=` in the sql. I checked a similar plugin report_customsql, and it used PARAM_RAW

**Testing**

- Manually tested by putting in the SQL from the issue linked above into step definitions, saved, and confirmed no longer removes the text after the `<=`
- (side note, I attempted to add a unit test for this which appears possible https://moodledev.io/docs/4.5/apis/subsystems/form#unit-testing but I could not get it to work properly and it does not seem worth it to spend any longer)

** Cherry picks**
- https://github.com/catalyst/moodle-tool_dataflows/pull/913

### DIFF
--- a/classes/local/step/reader_sql.php
+++ b/classes/local/step/reader_sql.php
@@ -48,7 +48,8 @@ class reader_sql extends reader_step {
      */
     public static function form_define_fields(): array {
         return [
-            'sql' => ['type' => PARAM_TEXT, 'required' => true],
+            // Must be PARAM_RAW to avoid clean_param mangling the sql.
+            'sql' => ['type' => PARAM_RAW, 'required' => true],
             'counterfield' => ['type' => PARAM_TEXT],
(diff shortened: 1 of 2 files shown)


Peterburnett (approved):
LGTM. We need to allow carets, all other poweful syntax for SQL is already allowed in PARAM_TEXT eg ; '  "

PR #917, opened 2024-12-13
issue #915: allow floor php function

### LINKED ISSUE #915: Allow floor php function
In my case I'd like to floor time to 5 minutes like date("dmYhi", floor(time() / 300) * 300); to be able to match the file name.

dmitriim:
Fixes #915


danmarsden:
Can we please update the commit message to use 'allowlist' instead of 'whitelist''?

Issue #927, opened 2025-01-31
HTTP headers are concatenated if running in Windows environment


dmitriim:
For a CURL step if you have multiple headers and your Moodle runs on Windows infrastructure, then HTTP headers are concatenated when sent to API end point.

dmitriim:
FIxed

PR #929, opened 2025-01-31
issue #927: fix an issue with PHP_EOL on Windows

dmitriim:


classes/helper.php
```diff
@@ -303,7 +303,7 @@ public static function extract_http_headers(string $content) {
         if ($content === '') {
             return [];
         }
-        $lines = explode(PHP_EOL, $content);
+        $lines = explode("\n", str_replace("\r\n", "\n", $content));
```
brendanheywood, line 306: I think add comment to the rfc for why it should be this way, regardless of which platform you are on

HTTP/1.1 defines the sequence CR LF as the end-of-line marker for all
   protocol elements except the entity-body (see appendix 19.3 for
   tolerant applications). 

https://datatracker.ietf.org/doc/html/rfc2616#section-2.2

brendanheywood: Also I'm assuming we can't unit test this unless we add windows into the matrix?

dmitriim: True. Also tests would need to be rewritten

dmitriim: done

Issue #931, opened 2025-02-03
tool_dataflows still dependent on local_aws in MOODLE_405_STABLE Branch


nbozhkov-ucl:
In MOODLE_405_STABLE branch tool_dataflows is still dependent (although not declared as dependency in the plugins check) on local_aws plugin for AWS SDK which is part of core Moodle since 4.4.

https://github.com/catalyst/moodle-tool_dataflows/blob/MOODLE_405_STABLE/classes/local/step/s3_trait.php#L103

This is causing all AWS related steps to fail if you don't have local_aws installed.

PR #934, opened 2025-02-05
issue #931: get rid of local_aws

dmitriim:
We use marked for deprecation method, but HQ may change their mind about that. See some comms in https://tracker.moodle.org/browse/MDL-82459


glepod:
The changes look ok to me.

PR #952, opened 2025-03-31
New step remove sftp file

gbarat87:
image

This step is to remove a file from the sftp server.
This is needed as there is no possibility to move a file from a folder to another in the sftp server.
The step "sftp copy file" following by this new one can allows one to move a file around.


classes/local/step/connector_sftp_delete_file.php
```diff
@@ -0,0 +1,129 @@
...
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class connector_sftp_delete_file extends connector_sftp {
+    use sftp_trait{
+        sftp_trait::form_define_fields as  sftp_form_define_fields;
+        sftp_trait::form_add_custom_inputs as sftp_form_add_custom_inputs;
+        sftp_trait::validate_for_run as sftp_validate_for_run;
+        sftp_trait::validate_config as sftp_validate_config;
+
+
```
matthewhilton, line 38: empty space

classes/local/step/connector_sftp_delete_file.php
```diff
@@ -0,0 +1,129 @@
...
+            $sftp = $this->init_sftp($config);
+
+            $sourceisremote = helper::path_is_scheme($config->source, self::$sftpprefix);
+            $sourcepath = $this->resolve_path($config->source);
+
+            // Deleting from remote.
+            if ($sourceisremote) {
+                $this->delete_from_remote($sftp, $sourcepath);
+                return $input;
+            }
```
matthewhilton, line 90: If the source isn't a remote, it should throw an exception here. Because it's validated at form validation time, if this is not a remote here it means something is very wrong.

lang/en/tool_dataflows.php
```diff
@@ -158,6 +158,7 @@
 $string['step_name_connector_set_variable'] = 'Set variable';
 $string['step_name_connector_set_multiple_variables'] = 'Set multiple variables';
 $string['step_name_connector_sftp'] = 'SFTP file copy';
+$string['step_name_connector_sftp_delete_file'] = 'SFTP file to delete';
```
matthewhilton, line 161: Could this be `SFTP delete file` instead ?

classes/local/step/sftp_trait.php
```diff
@@ -190,19 +205,42 @@ public function validate_config($config, $behaviour = 'copy') {
...
+            }
+        } else {
+            if (!empty($config->source) && !empty($config->target)) {
+                // Check if the source or target is an expression, and evaluate it if required.
+                $sourceremote = helper::path_has_scheme($config->source, self::$sftpprefix);
+                $targetremote = helper::path_has_scheme($config->target, self::$sftpprefix);
+                $hasremote = $sourceremote || $targetremote;
+            }
+            if (!$hasremote) {
+                $errormsg = get_string('connector_sftp:missing_remote', 'tool_dataflows', null, true);
```
matthewhilton, line 240: This lang string outputs as `At least one of source/target must be remote. ` but there is only one field for the delete step - it might need another string for just `Source must be remote. `

classes/local/step/sftp_trait.php
```diff
@@ -131,6 +136,16 @@ public function form_add_custom_inputs(\MoodleQuickForm &$mform, $behaviour = 'c
                 )
             );
         }
+        if ($behaviour === 'delete') {
+            $mform->addElement('text', 'config_source', get_string('connector_sftp:source', 'tool_dataflows'));
+            $mform->addElement('static', 'config_source_desc', '',  get_string('connector_sftp:source_desc', 'tool_dataflows').
+                    \html_writer::nonempty_tag(
+                            'pre',
+                            get_string('connector_sftp:path_example', 'tool_dataflows').
```
matthewhilton, line 144: Needs a different lang string, this one is currently:

```

Path to the source file. This can be a local file or remote path e.g.

    sftp://file/path  # Any remote file url;
    out.csv           # A path relative to the flows temp working dir;
    /var/data.json    # An absolute path;
    file:///my/in.txt # Any valid php stream url;
```

but in the case of this step, only `sftp://file/path` is valid

matthewhilton: Also, `Source` is probably not the most ideal word,  `File to delete` might fit better ?

classes/local/step/sftp_trait.php
```diff
@@ -190,19 +205,42 @@ public function validate_config($config, $behaviour = 'copy') {
...
+                // Check if the source or target is an expression, and evaluate it if required.
+                $sourceremote = helper::path_has_scheme($config->source, self::$sftpprefix);
+                $hasremote = $sourceremote;
+            }
+            if (!$hasremote) {
+                $errormsg = get_string('connector_sftp:missing_remote', 'tool_dataflows', null, true);
+                $errors['config_source'] = $errors['config_source'] ?? $errormsg;
+            }
+        } else {
+            if (!empty($config->source) && !empty($config->target)) {
```
matthewhilton, line 233: Should this care about `$config->target` at all? It's not shown as a mform element

matthewhilton (commented):
Could this also have unit tests added? e.g. see https://github.com/catalyst/moodle-tool_dataflows/blob/MOODLE_405_STABLE/tests/tool_dataflows_sftp_test.php
Since there's no actual SFTP to test against, it might just end up being a test for 1. sanity check that step loads properly, and 2. testing that source is always a remote, and errors otherwise

matthewhilton:
Another thought - could this also be added as a flow step just for consistency.

classes/local/step/sftp_trait.php
gbarat87: The text has been updated for delete and rename

classes/local/step/sftp_trait.php
gbarat87: this part is in the behavior "copy" so we do need the target here

classes/local/step/connector_sftp_delete_file.php
gbarat87: This has been removed

lang/en/tool_dataflows.php
gbarat87: This has been rewritten

classes/local/step/connector_sftp_delete_file.php
gbarat87: An exception has been raised here

classes/local/step/sftp_trait.php
gbarat87: This part has been updated

gbarat87:
@matthewhilton Thank you for the review.
After testing, I ave decided to simply add the step "rename" to allow to rename a file in the sftp server. That is making the move easier.
I decided to keep the delete step as it can be useful as well.
All the methods have been moved to a trait to reuse the same code.

classes/local/step/sftp_trait.php
```diff
@@ -131,6 +136,16 @@ public function form_add_custom_inputs(\MoodleQuickForm &$mform, $behaviour = 'c
                 )
```
matthewhilton, line 136: The strings here for `config_target_desc` and `config_source_desc` don't make sense for the rename step,

for rename the source and target must be remotes, but the help text says it can be a local file

matthewhilton (commented):
Gave this a test locally and works as expected, including field validation (for reference, I used https://hub.docker.com/r/atmoz/sftp)

Just 1 tiny change regarding some lang strings.

Also because we changing the lang strings the version.php needs a bump

matthewhilton:
Thanks for updating those lang strings @gbarat87 - the CI is failing now because of the php copy paste detector. Honestly i'd be open to disabling it to get the ci green as it isn't super valuable.

matthewhilton (approved):

Issue #954, opened 2025-04-01
Event trigger step stopped working at 4.3


jaypha:
The event trigger step makes use of eventlist::get_plugin_list().
This was changed in 4.3, breaking the event trigger step code.

jaypha:
A fix was made for MOODLE_405_STABLE. But one needs to be made for MOODLE_403_STABLE as well.

PR #965, opened 2025-04-07
Issue #954: Fix get_events_list() to work with 4.3 changes.

jaypha:
Also
- Fix deprecation notices in unit test.
- Fix get_steptype() bug in step.
- Fix deprecation notices in step.

### DIFF
--- a/classes/local/step/trigger_event.php
+++ b/classes/local/step/trigger_event.php
@@ -130,14 +130,31 @@ class trigger_event extends trigger_step {
      * @return array
      */
     private function get_events_list() {
+        global $CFG;
+
         $eventlist = \tool_monitor\eventlist::get_all_eventlist();
         $pluginlist = \tool_monitor\eventlist::get_plugin_list($eventlist);
         $plugineventlist = [];
-        foreach ($pluginlist as $plugin => $pluginname) {
-            foreach ($eventlist[$plugin] as $event => $eventname) {
-                // Filter out events which cannot be triggered for some reason.
-                if (!$event::is_deprecated()) {
-                    $plugineventlist[$event] = "{$pluginname}: {$eventname}";
+
+        // Check for the version where this structure was changed.
+        if ($CFG->version >= 2023062300) {
+            foreach ($pluginlist as $plugintype => $plugins) {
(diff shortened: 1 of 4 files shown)


classes/local/step/trigger_event.php
```diff
@@ -130,14 +130,31 @@ public static function form_define_fields(): array {
...
         $pluginlist = \tool_monitor\eventlist::get_plugin_list($eventlist);
         $plugineventlist = [];
-        foreach ($pluginlist as $plugin => $pluginname) {
-            foreach ($eventlist[$plugin] as $event => $eventname) {
-                // Filter out events which cannot be triggered for some reason.
-                if (!$event::is_deprecated()) {
-                    $plugineventlist[$event] = "{$pluginname}: {$eventname}";
+
+        // Check for the version where this structure was changed.
+        if ($CFG->version >= 2023062300) {
```
Peterburnett, line 140: I think there is a major flag which outputs something like 43, but its probably not necessary, just a bit more readable

version.php
```diff
@@ -25,8 +25,8 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-$plugin->version = 2024101001;
-$plugin->release = 2024101001;
+$plugin->version = 2024101002;
+$plugin->release = 2024101002;
 $plugin->requires = 2022112800;    // Our lowest supported Moodle (3.3.0).
 $plugin->supported = [400, 402];
```
Peterburnett, line 31: Should we change the supported, given that this change is to support at least 4.3 if not higher?

jaypha: It should work with both, that's why I had the check. It if is an older version, it will run the older code.

classes/local/step/trigger_event.php
jaypha: There is $CFG->branch, which is '403', etc. However, when I looked around, comparisons were mostly done with 'version'.

version.php
jaypha: I have updated this. Curious that it didn't cause problems when I was working with a 4.4 setup.

Peterburnett (approved):

PR #990, opened 2025-06-20
downgraded monolog/logger method headers 

### LINKED ISSUE #989: PHPUnit Fatal error - Monolog logger
Moodle MOODLE_405_STABLE
dataflows MOODLE_405_STABLE
PHP 8.2

```
Fatal error: Declaration of Monolog\Logger::emergency(Stringable|string $message, array $context = []): void must be compatible with Psr\Log\LoggerInterface::emergency($message, array $context = []) in /var/www/html/admin/tool/dataflows/vendor/monolog/monolog/src/Monolog/Logger.php on line 683
  stdout_lines: <omitted>
```

opitz:
to provide compatibility with LoggerInterface.

I assume it is rooted in an older Google abstract Logger class here: https://github.com/moodle/moodle/blob/7d9030acd6e874fd1bb0bd221cb6373e60958aa2/lib/google/src/Google/Logger/Abstract.php#L25

This will fix #989.

### DIFF
--- a/vendor/monolog/monolog/src/Monolog/Logger.php
+++ b/vendor/monolog/monolog/src/Monolog/Logger.php
@@ -558,13 +558,13 @@ class Logger implements LoggerInterface, ResettableInterface
      *
      * This method allows for compatibility with common interfaces.
      *
-     * @param mixed             $level   The log level (a Monolog, PSR-3 or RFC 5424 level)
-     * @param string|Stringable $message The log message
-     * @param mixed[]           $context The log context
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       codechecker_max_warnings: 0 # CI should now fail on phpcs / code checker warnings.
       disable_behat: true
       disable_grunt: true
-      min_php: 8.1
+      min_php: 8.2
```
dmitriim, line 16: Moodle 4.5 can run on PHP 8.1 why are you bumping version here?

